PR #58, opened 2026-07-06
feat: raising hand &  play sicker sound 

phuocnguyenncc:


crates/mezon-store/src/voice.rs
```diff
@@ -342,6 +377,238 @@ impl VoiceStore {
...
+        self.active_sounds.insert(
+            user_id.clone(),
+            ActiveSound {
+                _player: None,
+                _remove_timer: None,
+            },
+        );
+        cx.notify();
+        cx.spawn(async move |this, cx| {
+            let bytes = match mezon_client::transport_runtime::fetch_bytes(&sound_url).await {
```
huylexuan44, line 615: Sound reaction refetches and redecodes from network on every event; cache decoded PCM by URL (raise sound already caches).

crates/mezon-store/src/voice.rs
```diff
@@ -342,6 +377,238 @@ impl VoiceStore {
...
+            .ok();
+        })
+        .detach();
+    }
+
+    pub fn is_sound_active(&self, user_id: &str) -> bool {
+        self.active_sounds.contains_key(user_id)
+    }
+
+    fn handle_sound_reaction(
```
huylexuan44, line 578: No receive-side rate limit; a peer can spam sound reactions, each triggering a fetch+decode+notify. Throttle by sender_id.

crates/mezon-store/src/voice.rs
```diff
@@ -342,6 +377,238 @@ impl VoiceStore {
...
+            self.raised_hands.push(user_id.clone());
+        }
+        let key = user_id.clone();
+        let timer = cx.spawn(async move |this, cx| {
+            cx.background_executor().timer(RAISE_HAND_TTL).await;
+            this.update(cx, |this, cx| this.remove_raised_hand(&key, cx))
+                .ok();
+        });
+        self.raised_hand_timers.insert(user_id, timer);
+        cx.notify();
```
huylexuan44: notify() fires even on re-raise when raised_hands is unchanged; only notify when the list actually changes.

crates/mezon-store/src/voice.rs
```diff
@@ -342,6 +377,238 @@ impl VoiceStore {
...
+
+    pub fn is_local_hand_raised(&self) -> bool {
+        self.local_user_id()
+            .is_some_and(|id| self.raised_hands.contains(&id))
+    }
+
+    fn register_realtime(cx: &mut Context<Self>) {
+        let entity = cx.entity();
+        RealtimeDispatch::global(cx).update(cx, |dispatch, _| {
+            dispatch.on(RealtimeKind::VoiceReaction, &entity, |this, event, cx| {
```
huylexuan44, line 432: Confirm server scopes VoiceReaction to room participants; otherwise every idle client wakes the foreground thread per event.

crates/mezon-store/src/voice.rs
```diff
@@ -342,6 +377,238 @@ impl VoiceStore {
...
+    ) {
+        self.active_sounds.insert(
+            user_id.clone(),
+            ActiveSound {
+                _player: None,
+                _remove_timer: None,
+            },
+        );
+        cx.notify();
+        cx.spawn(async move |this, cx| {
```
huylexuan44: Detached fetch task is not cancelled on reset(); it runs to completion after leaving the room. Store the handle to cancel.

PR #68, opened 2026-07-09
Feat/ctrl k global search

giachuvan:


crates/mezon-store/src/direct.rs
```diff
@@ -141,11 +143,39 @@ impl DirectChannelList {
             .iter()
             .map(|c| (c.id, c.unread_count))
             .collect();
+        let pending: Vec<DirectChannel> = self
```
huylexuan44: replace() is no longer authoritative: local DMs absent from the server response are retained forever. Scope the carry-over to the just-created DM instead of unioning the whole list.

crates/mezon-store/src/direct.rs
```diff
@@ -141,11 +143,39 @@ impl DirectChannelList {
...
+            .channels
+            .iter()
+            .filter(|local| !channels.iter().any(|remote| remote.id == local.id))
+            .cloned()
+            .collect();
         merge_dm_unread_counts(&mut channels, badge_map, &local_counts);
+        channels.extend(pending);
```
huylexuan44: pending is appended after merge_dm_unread_counts, so these entries skip badge reconciliation and keep stale unread_count. Also O(local x remote).

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -0,0 +1,598 @@
...
+    let normalized_query = normalize_string(query);
+    if normalized_query.is_empty() {
+        return plain(text);
+    }
+
+    let Some(index) = normalized_text.find(normalized_query.as_str()) else {
+        return plain(text);
+    };
+
+    if normalized_text.len() != text.len() {
```
huylexuan44: Panic: the byte-length guard does not imply matching byte offsets. normalize_string("Maß")="MASS" (4 bytes, same as text), query "s" finds index 2, then rest.split_at(1) splits inside ß. Match on the normalized string and map back via char_indices.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -0,0 +1,598 @@
...
+        return plain(text);
+    }
+
+    let (before, rest) = text.split_at(index);
+    let match_len = query.len().min(rest.len());
```
huylexuan44: match_len uses the raw query byte length, not normalized_query.len(), so diacritic queries highlight the wrong span.

crates/mezon-ui/src/command_palette/filter.rs
```diff
@@ -0,0 +1,272 @@
...
+    indices.sort_by(|&left, &right| compare_items(&items[left], &items[right], raw_query, true));
+}
+
+fn compare_items(
+    left: &PaletteItem,
+    right: &PaletteItem,
+    search_text: &str,
+    use_name: bool,
+) -> Ordering {
+    let search = normalize_search_string(search_text);
```
huylexuan44: Hot path: normalize_search_string runs inside the sort comparator, so an NFD pass plus three String allocations happen O(n log n) times per keystroke. Hoist it into sort_indices and pass &str.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -0,0 +1,692 @@
...
+        ) {
+            self.selected_visible = visible_ix;
+            cx.notify();
+        }
+    }
+}
+
+impl Render for CommandPaletteModal {
+    fn render(&mut self, window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        self.refresh_items_if_needed(cx);
```
huylexuan44: build_palette_items + filter + sort run inside render(), triggered by observers on five global stores. Any incoming message rebuilds the whole palette next frame.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -0,0 +1,692 @@
...
+        };
+        self.display_rows = Rc::new(build_display_rows(
+            self.items.as_ref(),
+            self.filtered.as_ref(),
+            &self.debounced_query,
+            &previous,
+            browse_context,
+            &section_labels(&self.locale),
+        ));
+        self.selected_visible = first_selectable_row(self.display_rows.as_ref());
```
huylexuan44: recompute_filtered resets selected_visible and scrolls to top, so a message arriving mid arrow-key navigation snaps the highlight back to row one. Preserve the selection by item id. Also avoid mutating the scroll handle from render().

crates/mezon-store/src/channel.rs
```diff
@@ -1134,6 +1312,55 @@ impl ChannelList {
...
+        cx: &mut Context<Self>,
+    ) {
+        let entry = self.previous_channels.entry(clan_id).or_default();
+        entry.retain(|id| *id != channel_id);
+        entry.insert(0, channel_id);
+        entry.truncate(5);
+        self.persist_previous_channels(cx);
+    }
+
+    fn persist_previous_channels(&self, cx: &mut Context<Self>) {
```
huylexuan44: persist_previous_channels clones the map and writes JSON on every navigation; a palette selection writes twice (select_visible, then sync_channel_route). Debounce it — collapse_state only persists on toggle.

crates/mezon-ui/src/command_palette/groups.rs
```diff
@@ -0,0 +1,323 @@
+use gpui::{AnyElement, FontWeight, SharedString, div, prelude::*, px};
+
+use mezon_store::{ChannelId, ChannelType, ClanId, DirectKind};
+
+use crate::theme::Theme;
+
+use super::items::{PaletteItem, PaletteItemKind, ROW_PX};
+
+const GROUP_ITEM_LIMIT: usize = 5;
```
huylexuan44, line 9: React caps neither Mentions nor Unread (ListGroupSeacrhModal renders all, capped only visually at max-h-250px). Confirm the 5-item limit is intentional and note it in the description.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -0,0 +1,598 @@
...
+        return String::new();
+    }
+    value
+        .nfd()
+        .filter(|ch| !('\u{0300}'..='\u{036f}').contains(ch))
+        .collect::<String>()
+        .to_uppercase()
+}
+
+pub(crate) fn normalize_search_string(value: &str) -> String {
```
huylexuan44, line 28: clippy -D warnings (Justfile:86) fails here: consecutive str::replace. Use a single replace with a char slice pattern.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -0,0 +1,598 @@
...
+    users_store: Option<&Entity<UsersByUserStore>>,
+    cx: &App,
+) -> SharedString {
+    if dm.kind != DirectKind::Dm {
+        return SharedString::default();
+    }
+    if !dm.peer_username.is_empty() {
+        return SharedString::from(dm.peer_username.clone());
+    }
+    if let Some(user_id) = dm.peer_user_id {
```
huylexuan44: clippy -D warnings fails here: collapsible if. Use let-chains.

crates/mezon-store/src/direct.rs
```diff
@@ -281,12 +311,60 @@ impl DirectMessageStore {
...
         cx: &mut Context<Self>,
     ) -> Task<anyhow::Result<(ChannelId, i32)>> {
         let api = self.api.clone();
-        cx.spawn(async move |_this, _cx| {
+        cx.spawn(async move |this, cx| {
             let desc = api.create_direct_channel(&[user_id.0]).await?;
-            Ok((ChannelId(desc.channel_id), desc.channel_type as i32))
+            let channel_id = ChannelId(desc.channel_id);
+            let channel_type = desc.channel_type as i32;
+            let _ = this.update(cx, |this, cx| {
```
huylexuan44: clippy -D warnings fails here: let_unit_value.

crates/mezon-store/src/channel.rs
```diff
@@ -1751,6 +1988,33 @@ fn save_collapse_state(pairs: Vec<(String, String)>) {
     }
 }
 
+fn load_previous_channels() -> HashMap<ClanId, Vec<ChannelId>> {
+    let path = previous_channels_path();
+    let data = match std::fs::read_to_string(&path) {
+        Ok(d) => d,
+        Err(_) => return HashMap::new(),
+    };
+    match serde_json::from_str(&data) {
```
huylexuan44: clippy -D warnings fails here: match can be simplified to unwrap_or_default().

crates/mezon-ui/Cargo.toml
```diff
@@ -26,6 +26,7 @@ chrono.workspace = true
 flume.workspace = true
 rust-embed.workspace = true
 unicode-segmentation.workspace = true
+unicode-normalization  = "0.1"
```
huylexuan44: Should be unicode-normalization.workspace = true — the PR already adds it to [workspace.dependencies], and every other dep here uses workspace.

crates/mezon-ui/src/lib.rs
```diff
@@ -70,8 +78,16 @@ pub fn init(cx: &mut gpui::App) {
     )]);
     #[cfg(debug_assertions)]
     cx.bind_keys([gpui::KeyBinding::new("cmd-alt-i", ToggleInspector, None)]);
+    cx.bind_keys([
+        gpui::KeyBinding::new("secondary-k", OpenCommandPalette, None),
+        gpui::KeyBinding::new("ctrl-k", OpenCommandPalette, None),
```
huylexuan44: secondary-k already maps to Cmd+K on macOS and Ctrl+K elsewhere. This extra binding adds a macOS-only Ctrl+K, which conventionally means kill-to-end-of-line in text fields.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -341,9 +341,9 @@ impl ChatLayout {
             } => self.sync_channel_route(clan_id, channel_id, cx),
             Route::Thread {
                 clan_id,
-                channel_id,
+                thread_id,
                 ..
-            } => self.sync_channel_route(clan_id, channel_id, cx),
+            } => self.sync_channel_route(clan_id, thread_id, cx),
```
huylexuan44: Unrelated change: Route::Thread now syncs thread_id, but channel_sidebar/mod.rs:180 still highlights by channel_id. Split it out or note it in the PR description.

crates/mezon-ui/src/command_palette/filter.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    if raw_query.is_empty() {
+        let mut indices: Vec<usize> = (0..items.len()).collect();
+        sort_indices(items, &mut indices, raw_query);
+        return indices;
+    }
+
+    if raw_query.starts_with('@') {
+        let search = normalize_search_string(raw_query.get(1..).unwrap_or_default());
+        let mut indices: Vec<usize> = items
```
huylexuan44, line 18: cargo fmt --check fails on this PR: all four command_palette files plus channel.rs (297, 2038, 2048), direct.rs (322, 352), layout.rs (340, 354). Run just fmt.

crates/mezon-ui/src/image_cache.rs
```diff
@@ -1,6 +1,6 @@
 use std::future::Future;
 use std::sync::Arc;
-use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicU64, Ordering};
```
huylexuan44, line 3: AtomicBool is still used at line 105 under cfg(target_os = "macos") — this breaks the macOS build. Restore the import.

crates/mezon-store/src/gallery.rs
```diff
@@ -178,15 +178,14 @@ pub fn resolve_attachment_uploader(
...
-        && !profile.display_name.is_empty()
-    {
-        let (avatar, avatar_raw) = uploader_urls(&profile.avatar_url, cfg);
-        return UploaderInfo {
-            name: profile.display_name,
-            avatar,
-            avatar_raw,
-        };
-    }
+        && !profile.display_name.is_empty() {
```
huylexuan44, line 181: Reverted to pre-let-chain rustfmt style. CI runs cargo fmt -p mezon-store --check and fails here.

crates/mezon-store/src/channel.rs
```diff
@@ -720,16 +756,142 @@ impl ChannelList {
                 }
                 visible_changed =
                     visible_changed || was_unread != ch.is_unread() || was_badge != ch.badge_count;
+                updated_channel = Some(ch.clone());
```
huylexuan44, line 759: Hot path: clones the whole Channel (4 Strings + Vec) per socket message just to copy 5 Copy scalars. Use a snapshot struct.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -0,0 +1,734 @@
...
+                _users_observe: Subscription::new(|| ()),
+                _members_observe: Subscription::new(|| ()),
+                _router_observe: Subscription::new(|| ()),
+            }
+        });
+
+        let view_for_observe = view.clone();
+        cx.defer(move |cx| {
+            view_for_observe.update(cx, |this, cx| {
+                this._channel_observe = cx.observe(&ChannelList::global(cx), |this, _, cx| {
```
huylexuan44, line 151: Five store observers each trigger a full build_palette_items rebuild — one per incoming message while the palette is open. Debounce it.

crates/mezon-store/src/direct.rs
```diff
@@ -281,12 +318,54 @@ impl DirectMessageStore {
...
+                    member_label
+                } else if !peer_username.is_empty() {
+                    peer_username.clone()
+                } else {
+                    desc.channel_label.clone()
+                };
+                let channel =
+                    direct_from_created(&desc, user_id, &label, &peer_avatar, &peer_username);
+                this.channels.upsert_created(channel);
+                this.freshness.mark_fetched();
```
huylexuan44, line 364: mark_fetched() without a real fetch blocks ensure_loaded for CACHE_TTL. pending_created already carries the new DM; drop this.

crates/mezon-store/src/channel.rs
```diff
@@ -886,6 +1057,11 @@ impl ChannelList {
             record_topic_parent_badge(&mut self.topic_parent_badges, clan_id, parent_id, topic_id);
             self.notify_channel_list(clan_id, cx);
         }
+        if let Some(parent) = self.channel(clan_id, parent_id).cloned() {
```
huylexuan44, line 1060: Clan cached but parent_id missing: no sidebar or pending badge, yet user_channels still bumps. Gate the else on the None arm.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -422,6 +429,7 @@ impl ChatLayout {
             self.pending_channel_id = None;
             if !already_active {
                 self.channel_list.update(cx, |channel_list, cx| {
+                    channel_list.record_previous_channel(clan_id, channel_id, cx);
```
huylexuan44, line 432: Records the channel being navigated to, so Previous Channels preselects the current channel — Cmd+K then Enter is a no-op.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -0,0 +1,734 @@
...
+    member_label: String,
+    member_avatar: String,
+    member_username: String,
+    locale: SharedString,
+    cx: &mut App,
+) {
+    let Some(store) = DirectMessageStore::try_global(cx) else {
+        return;
+    };
+    let error_message: SharedString = mezon_i18n::t(&locale, "common.searchModal.noResults")
```
huylexuan44, line 659: noResults reads "Can't seem to find what you're looking for?" — wrong text for a create-DM failure toast.

crates/mezon-store/src/direct.rs
```diff
@@ -73,6 +75,7 @@ const DM_PAGE_SIZE: i32 = 500;
 struct DirectChannelList {
     channels: Vec<DirectChannel>,
     by_id: HashMap<ChannelId, usize>,
+    pending_created: Option<ChannelId>,
```
huylexuan44, line 78: Only one pending DM is tracked; creating two before the next replace() drops the first. Use a HashSet.

PR #75, opened 2026-07-10
Feat/ctrl f message search

giachuvan:


crates/mezon-store/src/message_search.rs
```diff
@@ -0,0 +1,1131 @@
...
+    pub local_date: Option<NaiveDate>,
+    pub image_attachment: Option<SearchHitImage>,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MessageSearchEvent {
+    SearchFailed,
+}
+
+pub struct MessageSearchStore {
```
huylexuan44, line 71: MessageSearchStore has no reset() and is missing from reset_all_user_stores - cached hits bleed across account switch on logout.

crates/mezon-ui/src/chat/message_search.rs
```diff
@@ -0,0 +1,1991 @@
...
+                            )
+                        })
+                        .when(has_query, |el| {
+                            el.cursor_pointer()
+                                .on_mouse_down(MouseButton::Left, move |_, window, cx| {
+                                    if let Some(search_input_for_close) =
+                                        search_input_for_close.clone()
+                                    {
+                                        search_input_for_close.update(cx, |input, cx| {
+                                            input.set_value("", window, cx);
```
huylexuan44, line 1337: Clear (X) handler lacks cx.stop_propagation(); the click bubbles to the bar handler and re-expands/refocuses instead of dismissing.

crates/mezon-ui/src/chat/message_search.rs
```diff
@@ -0,0 +1,1991 @@
...
+    div()
+        .child(StyledText::new(text).with_highlights(highlights))
+        .into_any_element()
+}
+
+fn search_highlight_ranges(text: &str, terms: &[String]) -> Vec<Range<usize>> {
+    if terms.is_empty() || text.is_empty() {
+        return Vec::new();
+    }
+    let lower = text.to_lowercase();
```
huylexuan44, line 1188: Match offsets computed on text.to_lowercase() but applied to the original text; non-ASCII case-folding shifts bytes -> misaligned highlight or panic.

crates/mezon-store/src/message_search.rs
```diff
@@ -0,0 +1,1131 @@
...
+                                .results
+                                .iter()
+                                .filter(|hit| hit.ogp.is_none() && search_hit_may_have_ogp(hit))
+                                .map(|hit| (hit.clan_id, hit.channel_id, hit.message_id))
+                                .collect::<Vec<_>>()
+                        }
+                        Err(err) => {
+                            tracing::warn!("search_message failed: {err}");
+                            state.has_error = true;
+                            state.total = 0;
```
huylexuan44, line 246: Error branch wipes total/results, wedging the pager on a transient page error. Keep prior results/total; set only has_error + is_searching=false.

crates/mezon-client/src/transport.rs
```diff
@@ -921,6 +1082,51 @@ mod vec_i32_flex {
...
+    pub fn deserialize<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
+    where
+        D: Deserializer<'de>,
+        T: DeserializeOwned,
+    {
+        Ok(Option::<Vec<T>>::deserialize(deserializer)?.unwrap_or_default())
+    }
+}
+
+mod vec_content_token_lenient {
```
huylexuan44, line 1098: Double-parses (Value then from_value) and silently drops tokens for every message on the shared decode path, not just search. Measure the hot-path cost.

crates/mezon-ui/src/chat/message_search.rs
```diff
@@ -0,0 +1,1991 @@
...
+                .hover(|s| s.bg(theme.tokens.bg_item_theme_hover))
+                .on_mouse_down(MouseButton::Left, on_click)
+        })
+        .text_size(px(14.))
+        .text_color(theme.tokens.text_theme_primary)
+        .child(page.to_string())
+        .into_any_element()
+}
+
+fn build_search_list_rows(hits: &[SearchHit]) -> Vec<SearchListRow> {
```
huylexuan44, line 619: Grouped rows are rebuilt and each SearchHit cloned twice on every render. Build the grouped rows once when results change, not in render.

crates/mezon-ui/src/chat/message_search.rs
```diff
@@ -0,0 +1,1991 @@
...
+impl MessageSearchPanel {
+    pub fn new(
+        layout: WeakEntity<ChatLayout>,
+        channel_id: ChannelId,
+        clan_id: ClanId,
+        is_direct: bool,
+        locale: SharedString,
+        cx: &mut Context<Self>,
+    ) -> Self {
+        cx.observe(&MessageSearchStore::global(cx), |this, _, cx| {
```
huylexuan44, line 188: .detach() on a transient panel leaks one observer per recreation. Store the Subscription in a panel field so it drops with the panel.

crates/mezon-client/src/transport.rs
```diff
@@ -565,6 +556,172 @@ fn parse_message_attachments(bytes: &[u8]) -> Vec<ApiAttachment> {
...
+    }
+    if (trimmed.starts_with('[') || trimmed.starts_with('{'))
+        && let Ok(value) = serde_json::from_str::<serde_json::Value>(trimmed)
+    {
+        let parsed = parse_search_attachment_json_value(&value);
+        if !parsed.is_empty() {
+            return parsed;
+        }
+    }
+    if let Ok(bytes) = base64::Engine::decode(&base64::engine::general_purpose::STANDARD, trimmed) {
```
huylexuan44, line 585: STANDARD base64 rejects URL-safe/unpadded input that React accepts; attachments/mentions can silently vanish from hits. Use a lenient decoder.

crates/mezon-client/src/transport.rs
```diff
@@ -565,6 +556,172 @@ fn parse_message_attachments(bytes: &[u8]) -> Vec<ApiAttachment> {
...
+    items
+        .iter()
+        .filter_map(parse_search_mention_json_item)
+        .collect()
+}
+
+fn parse_search_mention_json_item(item: &serde_json::Value) -> Option<ApiEntityMention> {
+    let s = item.get("s").and_then(json_to_i32).unwrap_or(0);
+    let e = item.get("e").and_then(json_to_i32)?;
+    if e <= s {
```
huylexuan44, line 635: Drops mentions with missing e or e<=s that React keeps; loses mention rendering/highlight parity in search hits.

crates/mezon-store/src/message.rs
```diff
@@ -928,7 +928,14 @@ pub fn parse_spans(content: &ApiMessageContent) -> Vec<MessageSpan> {
                             &content.cvtt,
                         ));
                     }
-                    "lk_ogp" => {}
+                    "lk_ogp" => {
```
huylexuan44, line 931: lk_ogp emits a Link with empty text when the marker sits past end-of-text (shared parse_spans). Guard empty inner to avoid a zero-width clickable link.

crates/mezon-client/src/search_message.rs
```diff
@@ -0,0 +1,868 @@
...
+        pages.extend((1..=5).map(SearchPageToken::Page));
+        if total_pages > 5 {
+            pages.push(SearchPageToken::Ellipsis);
+            pages.push(SearchPageToken::Page(total_pages));
+        }
+    } else if current_page >= total_pages - 2 {
+        pages.push(SearchPageToken::Page(1));
+        pages.push(SearchPageToken::Ellipsis);
+        pages.extend(
+            ((total_pages - 5)..=total_pages)
```
huylexuan44, line 61: total_pages==6: (total_pages-5)..= starts at 1 and duplicates the explicit Page(1), rendering '1 ... 1 2 3 4 5 6'.

PR #82, opened 2026-07-13
Feat/invite people

duocphungvanNCC:
> `Description:`

**env development:** window
- Implement invite people into clan
- alow search name of friend to invite.
- send direct message of link invite 
- copy invite link
- QR code for invite link 
image


crates/mezon-ui/src/clan/invite_people_modal.rs
```diff
@@ -0,0 +1,1001 @@
...
+        Some(code.to_string())
+    }
+}
+
+fn invite_origin_from_config(config: Option<&AppConfig>) -> String {
+    let Some(config) = config else {
+        return String::new();
+    };
+
+    invite_origin_from_host(&config.api_host, config.api_secure)
```
huylexuan44: api_host is the REST endpoint, not the web origin. In prod (.env.example: api.mezon.ai) every invite link becomes https://api.mezon.ai/invite/... and is dead. Use config.redirect_uri (NX_CHAT_APP_REDIRECT_URI, same source as React) then domain_url. The domain_url fallback here is unreachable since api_host is never empty.

crates/mezon-store/src/direct.rs
```diff
@@ -383,6 +383,59 @@ impl DirectMessageStore {
...
+        cx.spawn(async move |this, cx| {
+            let (channel_id, mode) = match existing {
+                Some(existing) => existing,
+                None => {
+                    let desc = api.create_direct_channel(&[user_id.0]).await?;
+                    let channel_id = ChannelId(desc.channel_id);
+                    let kind = DirectKind::from_raw(desc.channel_type);
+                    let mode = kind.stream_mode();
+                    this.update(cx, |this, cx| {
+                        let channel = direct_from_created(&desc, user_id, "", "", "");
```
huylexuan44: New DM channel is inserted with empty label, avatar and username, so the DM sidebar shows a blank row. create_dm_with_user resolves these from UsersByUserStore; pass the friend label/avatar you already have in InviteFriendRow.

crates/mezon-store/src/direct.rs
```diff
@@ -383,6 +383,59 @@ impl DirectMessageStore {
...
+                        this.channels.upsert_created(channel);
+                        this.freshness.mark_fetched();
+                        // cx.emit(DirectEvent::Changed);
```
huylexuan44: Dead commented emit. Every other path here emits DirectEvent::Changed { channel_id: None } after upsert_created. If send_channel_message fails the new channel is never announced. Emit it or delete the line.

crates/mezon-store/src/clan.rs
```diff
@@ -702,6 +707,27 @@ impl ClanList {
...
+        clan_id: ClanId,
+        channel_id: Option<ChannelId>,
+        cx: &mut Context<Self>,
+    ) -> Task<Result<ClanInviteLink, String>> {
+        let api = self.api.clone();
+        let clan = clan_id.get();
+        let channel = channel_id.map(ChannelId::get).unwrap_or_default();
+        cx.spawn(async move |_, _| {
+            let link = api
+                .create_link_invite_user(clan, channel, 0)
```
huylexuan44: expiry_time is hardcoded 0 but React passes 10 (modalInvite.tsx:61). Confirm the server semantics for 0 — if it means never expires, this ships non-expiring clan invites.

crates/mezon-ui/src/clan/invite_people_modal.rs
```diff
@@ -0,0 +1,1001 @@
...
+        );
+    }
+
+    #[test]
+    fn invite_link_uses_api_invite_id_instead_of_clan_id() {
+        let cfg = AppConfig::dev_defaults();
+        let link = ClanInviteLink {
+            id: 2076277531916374016,
+            invite_link: String::new(),
+            ..Default::default()
```
huylexuan44: clippy needless_update: all fields are already specified, so ..Default::default() has no effect. This fails just lint (-D warnings). cargo fmt --check also fails on this file (indentation at 187/260/355, missing trailing newline).

crates/mezon-ui/src/clan/invite_people_modal.rs
```diff
@@ -0,0 +1,1001 @@
...
+        })
+            .detach();
+    }
+
+    fn copy_invite_link(&mut self, cx: &mut Context<Self>) {
+        if !self.invite_link_ready() {
+            return;
+        }
+        cx.write_to_clipboard(ClipboardItem::new_string(self.invite_link()));
+        self.copied = true;
```
huylexuan44, line 387: copied is never reset, so the button stays in the Copied state for the modal lifetime. React clears it after 1500ms (modalInvite.tsx:185-197).

crates/mezon-ui/src/clan/invite_people_modal.rs
```diff
@@ -0,0 +1,1001 @@
...
+        let invite_loading = self.invite_link_loading;
+        let invite_error = self.invite_link_error.clone();
+        let invite_label = SharedString::from(tr(&self.locale, "inviteToChannel.btnInvite"));
+        let sent_label = SharedString::from(tr(&self.locale, "inviteToChannel.btnSent"));
+        let sending_label = SharedString::from(tr(&self.locale, "forwardMessage.modal.sending"));
+
+        const ROW_HEIGHT: f32 = 56.;
+        const MAX_VISIBLE_ROWS: usize = 4;
+
+        let rows = self.rows.clone();
```
huylexuan44: Hot path: rows Vec plus both HashSets are cloned on every render to feed the uniform_list closure. Capture the entity and read it inside the closure instead, as member_list.rs:812-815 does.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -453,9 +453,23 @@ impl Render for ChannelSidebar {
                 locale.clone(),
             )
         });
+        let active_clan_for_menu = self.clan_list.read(cx).active_clan().map(|clan| {
```
huylexuan44: Computed on every sidebar render (two String clones) but only used when clan_menu_open. Move it inside the .then() closure.

crates/mezon-ui/src/clan/invite_people_modal.rs
```diff
@@ -0,0 +1,1001 @@
...
+
+    Some(QrInviteImage { render, clipboard })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn invite_origin_uses_dev_when_runtime_hosts_are_dev() {
```
huylexuan44: This test passes only because dev_defaults().api_host happens to equal the dev web host, so it encodes the api_host bug as expected behavior. Assert the redirect_uri-derived origin and add a case where API host and web origin differ.

crates/mezon-ui/src/clan/invite_people_modal.rs
```diff
@@ -0,0 +1,1001 @@
...
+
+fn invite_send_link_label(locale: &str) -> String {
+    replace_i18n_var(
+        mezon_i18n::t(locale, "invitation.modal.sendLinkText"),
+        "type",
+        mezon_i18n::t(locale, "invitation.modal.clanInvite"),
+    )
+}
+
+fn copy_qr_label(locale: &str) -> String {
```
huylexuan44, line 952: Trailing space in the label and a leftover commented-out second arg. format!("{} ", ...) should just be the t() call.

crates/mezon-ui/src/clan/invite_people_modal.rs
```diff
@@ -0,0 +1,1001 @@
...
+                .child(button_label),
+        )
+        .into_any_element()
+}
+
+fn tr(locale: &str, key: &'static str) -> String {
+    mezon_i18n::t(locale, key).to_string()
+}
+
+fn replace_i18n_var(template: &str, key: &str, value: &str) -> String {
```
huylexuan44: Reinvents the house idiom: 18 call sites do t(locale, key).replace("{{var}}", val) directly (e.g. root.rs:329). This also rebuilds the {{key}} literal at runtime, allocating an extra String per call.

PR #85, opened 2026-07-13
Create Topic Discussion

giachuvan:


crates/mezon-store/src/topics.rs
```diff
@@ -92,14 +141,761 @@ impl TopicsStore {
...
+                cx,
+            );
+        });
+    }
+
+    fn handle_topic_in_message_event(&mut self, event: &RealtimeEvent, cx: &mut Context<Self>) {
+        let RealtimeEvent::TopicInMessageEvent(ev) = event else {
+            return;
+        };
+        let tp_id = if !ev.tp_id.is_empty() {
```
huylexuan44: Does not compile: TopicInMessageEvent has no tp_id field (only message_id, rpl, lsnt) in the pinned mezon-protocol. Drop the branch and always use resolve_topic_id_for_origin, or bump the protocol.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -1030,11 +1055,100 @@ impl ChannelMessages {
...
             row_memo: Rc::new(RefCell::new(RowMemo::default())),
             row_memo_day: None,
+            is_topic_box: false,
+            topic_align_timeline: None,
+            topic_list_layout: None,
+            topic_spacer_h: None,
+        }
+    }
+
+    pub fn new_topic_box(
```
huylexuan44, line 1246: Topic box drives the MAIN channel. new_topic_box reuses Self::new, which installs the scroll handler at line 871; nothing in it is gated on is_topic_box, so it calls set_viewing_older / sync_channel_seen / schedule_pagination_check on active_channel_id. Scrolling the topic panel paginates the channel and marks it read. Early-return when is_topic_box.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -826,6 +848,7 @@ impl ChannelMessages {
                     }));
                 }
                 MessagesEvent::ReplyTargetChanged => return,
+                MessagesEvent::TopicUpdated { .. } => {}
```
huylexuan44, line 982: Main list no-ops on TopicUpdated while Updated does remeasure_items and RemovedAt does splice. Combined with the store emitting TopicUpdated for main-channel messages whenever the panel is open, a delete leaves item_count stale (rows below render the wrong message) and an edit leaves the row height unmeasured.

crates/mezon-store/src/messages.rs
```diff
@@ -1582,27 +1608,40 @@ impl MessagesStore {
         msg.is_edited = true;
         patch_reply_previews_after_update(&mut channel.messages, message_id, &content);
         self.editing = None;
-        cx.emit(MessagesEvent::Updated {
-            message_id: Some(message_id),
-        });
+        if let Some(topic_id) = self.active_topic_id {
```
huylexuan44: Branches on "is the panel open", not on which bucket the message lives in. storage_id already resolves that correctly. Editing a main-channel message with the panel open emits TopicUpdated so the main list never repaints. Use self.active_topic_id == Some(storage_id).

crates/mezon-store/src/messages.rs
```diff
@@ -1582,27 +1608,40 @@ impl MessagesStore {
...
+            });
+        }
         cx.notify();
 
         let api = self.api.clone();
         let clan_id = self.active_clan_id.map_or(0, |c| c.get());
-        let channel_num = channel_id.get();
         let message_num = message_id.get();
+        let (api_channel_id, api_topic_id, is_update_msg_topic) =
+            if let Some(topic_id) = self.active_topic_id {
```
huylexuan44: Same bug: sends the edit to api_channel_id = topic_id with is_update_msg_topic = true even when the message lives in the parent channel. Gate on storage_id, not on the panel being open.

crates/mezon-store/src/messages.rs
```diff
@@ -1615,21 +1654,41 @@ impl MessagesStore {
...
-        self.apply_message_remove(channel_id, message_id, cx);
+        self.apply_message_remove(storage_id, message_id, cx);
 
         let api = self.api.clone();
         let clan_id = self.active_clan_id.map_or(0, |c| c.get());
-        let channel_num = channel_id.get();
+        let mode = self.mode;
+        let is_public = self.is_public;
         let message_num = message_id.get();
+        let (api_channel_id, api_topic_id) = if let Some(topic_id) = self.active_topic_id {
```
huylexuan44: Same bug: sends api_topic_id = topic_id when deleting a main-channel message with the panel open. Gate on storage_id.

crates/mezon-store/src/messages.rs
```diff
@@ -2745,7 +2913,12 @@ impl MessagesStore {
             );
         }
         patch_reply_previews_after_update(&mut channel.messages, message_id, &preview);
-        if is_active {
+        if let Some(topic_id) = self.active_topic_id {
```
huylexuan44: Same bug in the realtime path: any incoming edit to a main-channel message emits TopicUpdated while the panel is open, so the main list goes stale. Gate on storage_id.

crates/mezon-store/src/messages.rs
```diff
@@ -2770,18 +2943,61 @@ impl MessagesStore {
...
             return;
         };
         patch_reply_previews_after_delete(&mut channel.messages, message_id);
-        if let Some(index) = channel.messages.remove_id(message_id)
-            && is_active
-        {
-            cx.emit(MessagesEvent::RemovedAt { index, message_id });
+        if let Some(index) = channel.messages.remove_id(message_id) {
+            self.on_message_deleted(message_id, deleted_topic_id, cx);
+            if let Some(topic_id) = self.active_topic_id {
```
huylexuan44: Same bug: an incoming delete in the main channel emits TopicUpdated instead of RemovedAt, so the row is never spliced out. Gate on storage_id.

crates/mezon-store/src/messages.rs
```diff
@@ -2867,13 +3090,19 @@ impl MessagesStore {
         let is_public = self.is_public;
         let message_sender_id = current_uid.get();
         let emoji_id_num = emoji_id.parse::<i64>().unwrap_or(0);
-        let channel = channel_id.get();
+        let api_channel = parent_channel_id.get();
+        let api_topic_id = self
+            .active_topic_id
```
huylexuan44: api_topic_id = storage_id sends the PARENT CHANNEL ID in the topic_id slot when reacting to a main-channel message with the panel open. Should be 0 unless storage_id == active_topic_id.

crates/mezon-ui/src/chat/message/user_row.rs
```diff
@@ -308,7 +311,10 @@ fn render_go_to_topic_footer(msg: &Message, ctx: &RowCtx) -> AnyElement {
         .border_color(theme.tokens.border_divider)
         .cursor_pointer()
         .hover(move |s| s.bg(hover_bg))
-        .on_click(move |_, _, cx| coming_soon_toast(&locale, cx))
+        .on_click(move |_, _, cx| {
+            let origin = origin.clone();
+            TopicsStore::global(cx).update(cx, |store, cx| store.start_create(origin, cx));
```
huylexuan44: Go to topic on a card message creates a SECOND topic. start_create only resolves existing_topic_id when origin.code == Topic, but a card has topic_id set and code != Topic, so the panel opens empty and the first reply calls create_sd_topic again. Exclude msg.topic_id.is_some() or resolve the existing topic id.

crates/mezon-store/src/topics.rs
```diff
@@ -92,14 +141,761 @@ impl TopicsStore {
...
+        rpl: i32,
+        lsnt: i64,
+        cx: &mut Context<Self>,
+    ) {
+        let lsnt = normalize_unix_seconds(lsnt);
+        let merged = self
+            .topic_meta
+            .get(&topic_id)
+            .map(|existing| TopicMessageMeta {
+                rpl: existing.rpl.max(rpl),
```
huylexuan44: rpl can never decrease: React overwrites the server value verbatim (upsertOne) and decrements on delete. max() here permanently ratchets the count and discards server corrections.

crates/mezon-store/src/topics.rs
```diff
@@ -194,3 +990,55 @@ impl TopicsStore {
...
+        clan_id: ev.clan_id,
+        channel_id: ev.channel_id,
+        status: 0,
+        create_time_seconds: 0,
+        update_time_seconds: ev
+            .last_sent_message
+            .as_ref()
+            .map(|h| h.timestamp_seconds as u32)
+            .unwrap_or(0),
+        content: String::new(),
```
huylexuan44: Realtime SdTopicEvent synthesizes content: "" / create_time 0 and upsert_topic full-replaces the stored topic, blanking a topic already loaded from listSdTopic (the Inbox Topics tab renders content). React uses addOne, which never clobbers. The event carries the message field — use it, or patch metadata only.

crates/mezon-store/src/topics.rs
```diff
@@ -92,14 +141,761 @@ impl TopicsStore {
...
+    }
+
+    fn handle_sd_topic_event(&mut self, event: &RealtimeEvent, cx: &mut Context<Self>) {
+        let RealtimeEvent::SdTopicEvent(ev) = event else {
+            return;
+        };
+        let topic = topic_discussion_from_api(sd_topic_from_event(ev));
+        self.upsert_topic(topic);
+        let meta = TopicMessageMeta {
+            rpl: 1,
```
huylexuan44: rpl hardcoded to 1; React creates the meta with rpl 0.

crates/mezon-store/src/topics.rs
```diff
@@ -92,14 +141,761 @@ impl TopicsStore {
...
+
+        match (cache_ts, meta_ts) {
+            (Some(a), Some(b)) => Some(a.max(b)),
+            (Some(a), None) => Some(a),
+            (None, Some(b)) => Some(b),
+            (None, None) => None,
+        }
+    }
+
+    pub fn topic_display_reply_count(
```
huylexuan44: React reads only topicMeta.rpl for the reply count — no local message count, no max. This cached-count-first precedence undercounts (topic messages are fetched one 50-item page with no pagination).

crates/mezon-store/src/topics.rs
```diff
@@ -92,14 +141,761 @@ impl TopicsStore {
...
+            .zip(self.compose.active_topic_id.map(ChannelId))
+            .is_some_and(|(tid, active)| tid == active);
+        is_origin || is_topic_anchor
+    }
+
+    pub fn close_panel_from_messages(&mut self, cx: &mut Context<Self>) {
+        self.close_panel_state(cx);
+    }
+
+    fn close_panel_state(&mut self, cx: &mut Context<Self>) {
```
huylexuan44, line 706: close_panel_state never clears init_topic_message_id, and channel_delete_blocked returns true for it. Open a topic, close without sending, and Delete Message is gone for that message for the rest of the session.

crates/mezon-ui/src/chat/message/message_context_menu.rs
```diff
@@ -32,14 +30,304 @@ pub(crate) fn resolve_forward_group(
...
+    menu = menu.separator();
+
+    {
+        let message_id = msg.id;
+        menu = menu.item_icon(
+            t("contextMenu.reply"),
+            IconName::ReplyRightClick,
+            move |_, cx| {
+                MessagesStore::global(cx)
+                    .update(cx, |store, cx| store.set_reply_to(message_id, cx));
```
huylexuan44: Reply inside the topic panel sets the GLOBAL reply target, but TopicsStore::submit_reply takes no reply reference and never reads it. The reply bar appears on the main composer, the topic reply is sent without the reference, and the stale target attaches to your next channel message.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -1180,27 +1288,43 @@ impl ChannelMessages {
...
-        let sender_and_poll = {
-            let store = MessagesStore::global(cx);
-            let store = store.read(cx);
-            store
-                .viewport_message_by_id(message_id)
-                .map(|m| (m.sender_id.clone(), m.code == MessageCode::Poll))
+        let sender_and_poll = self
+            .find_local_message(message_id, cx)
+            .map(|m| (m.sender_id.clone(), m.code == MessageCode::Poll));
+        let forward_messages = if self.is_topic_box {
```
huylexuan44: Right-click deep-clones the entire channel message buffer just to test len() > 1. resolve_forward_group still exists and borrows the slice — keep using it for the non-topic branch.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -2147,10 +2277,219 @@ impl ChannelMessages {
...
+            .active_channel()
+            .map(|c| (Some(c.channel_type), c.parent_id.is_none()))
+            .unwrap_or((None, true));
+        let is_clan_owner = self.cached_is_clan_owner;
+        let emoji_recent = self.emoji_recent.clone();
+        let (editing_id, edit_input) = match &self.edit_input {
+            Some((id, input)) => (Some(*id), Some(input.clone())),
+            None => (None, None),
+        };
+        let messages = Self::collect_topic_messages(cx);
```
huylexuan44: collect_topic_messages clones, sorts and regroups every topic message on every render, and runs a second time via find_local_message while a context menu is open. Read from the entity inside the list closure (see member_list.rs:812-815).

crates/mezon-ui/src/components/primitives/context_menu.rs
```diff
@@ -55,21 +66,32 @@ impl ContextMenu {
     ) -> Self {
         self.items.push(Item::Entry {
             label: label.into(),
-            icon: Some(icon),
+            leading_icon: None,
+            trailing_icon: Some(icon),
```
huylexuan44: item_icon now routes the icon to a trailing slot and the row uses justify_between, so all 13 existing callers render the icon right-aligned instead of leading. Voice participant menu and image viewer menu are visually regressed. leading_icon is set by no constructor, so that branch and its gap_2 are dead.

crates/mezon-ui/src/components/primitives/context_menu.rs
```diff
@@ -2,8 +2,9 @@ use std::rc::Rc;
 
 use gpui::{
     App, ClickEvent, MouseDownEvent, Pixels, Point, SharedString, Window, anchored, deferred, div,
-    prelude::*, px,
+    img, prelude::*, px,
 };
+use mezon_store::{MessageId, MessagesStore};
```
huylexuan44: A generic UI primitive should not depend on the message domain. Take a callback instead of calling MessagesStore::add_reaction directly.

crates/mezon-ui/src/chat/create_topic_panel.rs
```diff
@@ -0,0 +1,215 @@
...
+            window,
+            |this, _, event: &MentionInputEvent, window, cx| match event {
+                MentionInputEvent::Submit => this.submit(window, cx),
+                MentionInputEvent::SendSticker { url, filename } => {
+                    this.send_sticker(url.clone(), filename.clone(), cx);
+                }
+                MentionInputEvent::SendSound { url, filename } => {
+                    this.send_sound(url.clone(), filename.clone(), cx);
+                }
+                MentionInputEvent::Cancel => {}
```
huylexuan44: Cancel is a no-op here, but area.rs uses it to clear_reply — so Escape cannot clear a reply target set from the topic panel.

crates/vendor/gpui/src/elements/list.rs
```diff
@@ -10,8 +10,8 @@
 use crate::{
     AnyElement, App, AvailableSpace, Bounds, ContentMask, DispatchPhase, Edges, Element, EntityId,
     FocusHandle, GlobalElementId, Hitbox, HitboxBehavior, InspectorElementId, IntoElement,
-    Overflow, Pixels, Point, ScrollWheelEvent, Size, Style, StyleRefinement, Styled,
-    Window, point, px, size,
+    Overflow, Pixels, Point, ScrollWheelEvent, Size, Style, StyleRefinement, Styled, Window, point,
```
huylexuan44: cargo fmt was run over the vendored crates: 10 files under crates/vendor are reformatted with zero semantic change. CLAUDE.md forbids formatting vendored crates (they carry upstream drift). Please revert all 10.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -278,6 +281,21 @@ impl ChatLayout {
         .detach();
         cx.observe(&MessageSearchStore::global(cx), |_, _, cx| cx.notify())
             .detach();
+        cx.observe(&TopicsStore::global(cx), |this, store, cx| {
+            if store.read(cx).is_panel_open() {
+                ThreadsStore::global(cx).update(cx, |threads, cx| threads.cancel_create(cx));
+                this.reset_message_search(cx);
```
huylexuan44, line 289: Level-triggered: cancel_create and reset_message_search re-run on EVERY TopicsStore notify, which fires on realtime topic events for any topic in the clan. Type a message search with the panel open and someone replies in an unrelated topic — your search is wiped mid-typing. Move these to the open transition.

huylexuan44:
Reviewed by building and linting the branch, and cross-checking the React reference. Inline comments have the specifics. Four things block merge:

- The branch does not compile. topics.rs reads ev.tp_id, but TopicInMessageEvent has only message_id, rpl, lsnt in the pinned mezon-protocol (same pin on develop, so this is not a local artifact). mezon-ui was never type-checked as a result, so there may be more errors behind it.
- cargo fmt was run over the vendored crates: 10 files under crates/vendor reformatted, zero semantic change. Please revert them.
- Six sites in messages.rs branch on "is the topic panel open" instead of which bucket the message lives in, even though storage_id already resolves that correctly (and lines 2879 and 3075 use the right test). Edits, deletes and reactions on main-channel messages get routed to the topic, and the main list stops repainting.
- The topic box inherits the main list's scroll handler, so scrolling the panel paginates the channel, flags it viewing_older, and marks it read.

Worth a look beyond that: Go to topic on a card message creates a second topic; the reply-count rpl can never decrease (React overwrites verbatim and decrements on delete); a realtime SdTopicEvent full-replaces and blanks an already-loaded topic; Reply inside the panel sets the global reply target that the topic send path ignores.

Confirmed as matching React, no action needed: lazy topic creation on first reply, the non-optimistic attachment send, and the App/Voice/Stream channel-type gate.

The feature design and the store/UI split are sound — most of the above follows from one question ("is this message actually in the topic bucket?") and the fixes are small.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -635,8 +641,24 @@ impl ChannelMessages {
             }
         });
 
+        cx.observe(&TopicsStore::global(cx), |this, _, cx| {
```
huylexuan44: PERF: this observe is installed in new(), so it runs on EVERY timeline, and TopicsStore notifies on every realtime TopicInMessageEvent — i.e. any topic reply by anyone in any channel. Users who never open a topic get a full main-list re-render per topic message. Fingerprint-gate it or drop it for non-topic timelines.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -278,6 +281,21 @@ impl ChatLayout {
...
+            }
+            cx.notify();
```
huylexuan44: PERF: cx.notify() is outside the is_panel_open() guard, so the whole ChatLayout re-renders on every TopicsStore notify — including clan-wide topic traffic with no panel open. Move it inside the guard.

crates/mezon-store/src/topics.rs
```diff
@@ -92,14 +141,761 @@ impl TopicsStore {
...
+            });
+        self.topic_meta.insert(topic_id.clone(), merged);
+        if lsnt > 0
+            && let Some(idx) = self.topic_index.get(&topic_id).copied()
+            && let Some(topic) = self.topics.get_mut(idx)
+        {
+            topic.last_message_timestamp = lsnt.clamp(0, i64::from(u32::MAX)) as u32;
+        }
+        cx.emit(TopicsEvent::Updated);
+        cx.notify();
```
huylexuan44: PERF: no coalescing here. upsert_topic_meta emits + notifies per realtime topic event, unlike DirectMessageStore which coalesces on a 100ms timer (CHANGED_NOTIFY_COALESCE) precisely to avoid per-message notify floods. Add the same coalescing.

crates/mezon-ui/src/chat/message/parts.rs
```diff
@@ -27,15 +28,37 @@ use crate::theme::Theme;
 const DELETED_REPLY_PREVIEW: &str = "Original message was deleted";
 const FILE_NAME_COLOR: u32 = 0x3b_82_f6;
 
+pub fn resolve_message_display_name(msg: &Message, ctx: &RowCtx, cx: &App) -> SharedString {
```
huylexuan44, line 31: PERF: this replaced msg.sender_name.clone() (a free Arc bump) and is called twice per head row per frame (here and line 137). Each call does a sender_id.parse(), a ClanMembersStore lookup and a SharedString heap alloc. row_memo only caches avatars, so none of it is memoized. Precompute the display name onto Message at insert, or add it to row_memo.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -960,6 +983,8 @@ impl ChannelMessages {
                 cx.notify();
             }
         });
+        let channel_permissions_observe =
+            cx.observe(&ChannelPermissionsStore::global(cx), |_, _, cx| cx.notify());
```
huylexuan44: PERF: unconditional notify re-renders the whole message list on any permission-store change, and messages.rs now calls ensure_loaded on every channel activation — so this fires on every channel switch.

crates/mezon-ui/src/chat/message/topic_view_button.rs
```diff
@@ -23,23 +22,41 @@ pub fn render_topic_view_button(msg: &Message, ctx: &RowCtx) -> AnyElement {
...
+    let reply_count = msg
         .topic_id
-        .and_then(|topic_id| {
-            let store = TopicsStore::try_global(ctx.app)?;
-            let store = store.read(ctx.app);
+        .map(|topic_id| {
+            TopicsStore::try_global(ctx.app).map_or(0, |store| {
+                store
+                    .read(ctx.app)
+                    .topic_display_reply_count(topic_id, msg.id, ctx.app)
```
huylexuan44: PERF: topic_display_reply_count and topic_last_reply_timestamp each do a full messages_in_channel(topic_id) scan — two O(n) scans per topic-anchor row per frame. Cache both in TopicMessageMeta instead of rescanning on render.

crates/mezon-ui/src/chat/message/topic_view_button.rs
```diff
@@ -63,10 +87,16 @@ pub fn render_topic_view_button(msg: &Message, ctx: &RowCtx) -> AnyElement {
...
+            &creator_name,
+            &creator_avatar,
+            msg,
+            ctx,
+            ctx.app,
+        ))
         .child(meta);
 
-    let locale = SharedString::from(ctx.locale);
+    let origin = msg.clone();
```
huylexuan44: PERF: deep-clones the whole Message (attachments, reactions, spans) per row per frame just to feed the click closure. Capture msg.id and resolve from the store at click time. Same at user_row.rs:297.

crates/mezon-store/src/messages.rs
```diff
@@ -1735,22 +1794,116 @@ impl MessagesStore {
...
+    pub fn messages_in_channel(&self, channel_id: ChannelId) -> &[Message] {
+        self.cache
+            .get(&channel_id)
+            .map(|c| c.messages.as_slice())
+            .unwrap_or(&[])
+    }
+
+    /// Mark the topic bucket the discussion panel is watching so realtime replies
+    /// to it (a non-active channel) still notify the panel.
+    pub fn set_active_topic(&mut self, topic_id: Option<i64>, cx: &mut Context<Self>) {
```
huylexuan44, line 2042: PERF/memory: the topic bucket lives in the same KeyedCache as channels (MAX_CACHED_CHANNELS = 30) and is never removed when the panel closes — set_active_topic(None) only clears the id. Every topic opened permanently evicts a real channel, which then refetches from the network. Drop the bucket on close, or give topics their own cache.

crates/mezon-store/src/messages.rs
```diff
@@ -2495,6 +2648,11 @@ impl MessagesStore {
         self.is_public = is_public;
         self.is_dm = is_dm;
         self.mode = mode;
+        if !is_dm && !clan_id.is_zero() {
+            ChannelPermissionsStore::global(cx).update(cx, |store, cx| {
```
huylexuan44: PERF: ensure_loaded on every channel activation adds an API round-trip to the channel-switch path, and channel_messages.rs:987 re-renders the full message list when it lands.

huylexuan44:
Follow-up: performance pass, since this must not regress app perf.

The important finding is that several costs are paid by EVERY user, even those who never open a topic:

- TopicsStore notifies on every realtime TopicInMessageEvent (any topic reply, any channel, no filter, no coalescing). channel_messages.rs:644 installs the observer in new(), so it is on every timeline, and layout.rs:289 notifies outside the is_panel_open guard. Result: a busy clan with active topics continuously re-renders the main message list and the whole ChatLayout for people who never opened the feature. DirectMessageStore already solved exactly this with a 100ms coalescing timer — please reuse that.
- parts.rs:31 replaced msg.sender_name.clone() (a free Arc bump) with a parse + ClanMembersStore lookup + SharedString heap alloc, called twice per header row per frame and not covered by row_memo.
- channel_messages.rs:1294 deep-clones the entire channel message buffer on every right-click, only to test len() > 1. resolve_forward_group still exists and borrows the slice.
- messages.rs:1808 — topic buckets share the 30-entry channel KeyedCache and are never evicted on close, so each topic opened permanently evicts a real channel.

Topic-panel-only costs (still worth fixing): render_topic_box clones/sorts/regroups every topic message per frame (twice while a context menu is open), two O(n) store scans per topic-anchor row per frame, per-row Message deep clones, and list_state.reset() on every reply discarding all measured heights.

Inline comments have the specifics and the suggested fix for each. None of these require redesigning the feature.

crates/mezon-store/src/topics.rs
giachuvan: update latest mezon-protocol has tp_id field in TopicInMessageEvent

PR #87, opened 2026-07-14
feat: show status user invoice

phuocnguyenncc:
image
image


crates/mezon-store/src/channel.rs
```diff
@@ -473,6 +483,9 @@ impl ChannelList {
                         this.loading.remove(&clan_id);
                         this.app_channels_cache.insert(clan_id, app_channels);
                         this.merge_pending_badges(&mut categories);
+                        if seed_in_voice_from_categories(&mut this.in_voice, clan_id, &categories) {
```
huylexuan44, line 494: Seed only adds; on resync a peer who left during a dropped event stays a ghost "In voice". Clear this clan (retain clan_id != clan_id) before re-seeding.

crates/mezon-store/src/channel.rs
```diff
@@ -68,6 +68,12 @@ pub struct VoiceMember {
     pub avatar_url: String,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct InVoiceInfo {
```
huylexuan44, line 72: No status field — React renders "Sharing screen" for screen-shares; DM row/header always show "In voice". Add an EInvoice-style status.

crates/mezon-store/src/channel.rs
```diff
@@ -1273,14 +1290,15 @@ impl ChannelList {
                 for cats in self.cache.values_mut() {
                     removed |= remove_channel(cats, id);
                 }
+                let in_voice_changed = remove_in_voice_in_channel(&mut self.in_voice, id);
```
huylexuan44, line 1308: UserChannelRemoved handler does not clear in_voice like this does; call remove_in_voice_in_channel there too or entries go stale.

PR #98, opened 2026-07-17
Fix icons for DM header, bug in ctrlK

giachuvan:


crates/mezon-ui/src/command_palette/filter.rs
```diff
@@ -216,6 +218,22 @@ mod tests {
         assert_eq!(indices, vec![0]);
     }
 
+    #[test]
+    fn channel_search_does_not_match_clan_subtext() {
```
huylexuan44, line 222: This test passes even if the items.rs filter_blob fix is reverted: matches_search never reads subtext, and the mock builds filter_blob from a separate param. Cover build_palette_items instead.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -504,6 +535,66 @@ pub fn render_palette_row(
...
+        .min_w_0()
+        .overflow_hidden()
+        .child(label)
+        .child(
+            div()
+                .flex_shrink_0()
+                .text_size(px(15.))
+                .italic()
+                .text_color(theme.status_dnd)
+                .child("(busy)"),
```
huylexuan44, line 571: Hardcoded English "(busy)" — no i18n key; use mezon_i18n::t and thread the locale. Also no equivalent in the React SearchModal; confirm this is intentional.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -109,13 +115,41 @@ fn palette_channel_subtext(
...
+    clan_id: ClanId,
+    channel_id: ChannelId,
+    channel_type: ChannelType,
+) -> bool {
+    if channel_type != ChannelType::Voice {
+        return false;
+    }
+    channels
+        .channel(clan_id, channel_id)
+        .is_some_and(|channel| channel.voice_members.len() >= 2)
```
huylexuan44, line 150: Why >= 2 and not >= 1 for busy? Extract a named const and confirm this matches React voice-occupancy semantics.

crates/mezon-ui/src/chat/channel_header.rs
```diff
@@ -340,6 +358,202 @@ impl ChannelHeader {
...
+        bg_active: gpui::Rgba,
+        members_action: bool,
+        members_active: bool,
+        on_toggle_members: Option<ToggleHandler>,
+        pin_handle: Option<PopoverMenuHandle<PinnedPopoverPanel>>,
+        settings: Option<Entity<Settings>>,
+        locale: &str,
+        cx: &App,
+    ) -> Vec<AnyElement> {
+        let header = ChannelHeader {
```
huylexuan44, line 375: Building a throwaway ChannelHeader just to call dm_action_buttons is fragile — every new field must be added here. Make it a static fn taking the 5 params it needs.

PR #109, opened 2026-07-19
Add copy image and fix image viewer ontop when click image

giachuvan:


crates/vendor/gpui_linux/src/linux/wayland/window.rs
```diff
@@ -1281,7 +1281,11 @@ impl PlatformWindow for WaylandWindow {
             let serial = state.client.get_serial(SerialKind::MousePress);
             token.set_app_id(app_id);
             token.set_serial(serial, &state.globals.seat);
-            token.set_surface(&state.surface);
+            let interaction_surface = state
+                .client
+                .activation_interaction_surface(state.surface.id())
+                .unwrap_or_else(|| state.surface.clone());
+            token.set_surface(&interaction_surface);
```
huylexuan44, line 1286: serial (last MousePress) and surface (current focus) can be from different events; on Escape-close MousePress is stale since Escape only bumps KeyPress, so the compositor may reject the token. Capture (serial, source_surface) atomically at input time, including KeyPress.

crates/mezon-ui/src/chat/message/parts.rs
```diff
@@ -609,7 +609,7 @@ fn render_album(
             tile_element = tile_element.when(
                 !att.uploading && !att.upload_failed && !raw_url.is_empty(),
                 |d| {
-                    d.cursor_pointer().on_click(move |_, _window, cx| {
+                    d.cursor_pointer().on_mouse_down(MouseButton::Left, move |_, _window, cx| {
```
huylexuan44: on_mouse_down changes behavior on all OSes: it opens on press even if the user drags out and releases to cancel, and fires before a drag-select starts. Revert to on_click — the mouse-up release also refreshes the MousePress serial, so activation still gets a fresh serial. Same change at lines 641, 761, 839.

PR #122, opened 2026-07-21
Fix render bot embeded message

giachuvan:


crates/mezon-client/src/transport.rs
```diff
@@ -770,10 +770,99 @@ fn json_to_i32(value: &serde_json::Value) -> Option<i32> {
...
+            tokens.call_log = serde_json::from_value(call_log.clone()).ok();
+        }
+        tokens.is_card = map
+            .get("isCard")
+            .or_else(|| map.get("is_card"))
+            .and_then(|value| value.as_bool())
+            .unwrap_or(false);
+    }
+    if tokens.t.is_empty() && tokens.embed.is_empty() && tokens.components.is_empty() {
+        tokens.t = trimmed.to_string();
```
huylexuan44: Recovery leaks raw JSON into t for an object with no t/embed/components — the exact bug this PR fixes. Guard as message_search.rs:551 does: only fall back when trimmed does not start with { or ".

crates/mezon-store/src/messages.rs
```diff
@@ -5142,11 +5142,17 @@ fn edit_content_spans(content: &str, content_tokens: OutgoingContent) -> EditTra
...
         return None;
     }
-    u32::from_str_radix(raw, 16).ok().map(gpui::rgb)
+    let hex = raw.trim_start_matches('#');
+    if hex.len() == 6 && hex.bytes().all(|b| b.is_ascii_hexdigit()) {
+        return u32::from_str_radix(hex, 16).ok().map(gpui::rgb);
+    }
+    raw.parse::<u32>()
+        .ok()
+        .map(|value| gpui::rgb(value & 0x00_ff_ff))
```
huylexuan44: 0x00_ff_ff is a 16-bit mask and drops the red channel: 16711680 (#FF0000) renders black, 5793266 (#5865F2) renders #0065F2. Use value & 0x00ff_ffff.

crates/mezon-store/src/messages.rs
```diff
@@ -5142,11 +5142,17 @@ fn edit_content_spans(content: &str, content_tokens: OutgoingContent) -> EditTra
...
 fn parse_embed_accent(color: Option<&str>) -> Option<Rgba> {
-    let raw = color?.trim().trim_start_matches('#');
-    if raw.len() != 6 || !raw.bytes().all(|b| b.is_ascii_hexdigit()) {
+    let raw = color?.trim();
+    if raw.is_empty() {
         return None;
     }
-    u32::from_str_radix(raw, 16).ok().map(gpui::rgb)
+    let hex = raw.trim_start_matches('#');
+    if hex.len() == 6 && hex.bytes().all(|b| b.is_ascii_hexdigit()) {
```
huylexuan44: Decimal 123456 is 6 hex digits so it parses as 0x123456, not 0x01E240 — every decimal color in 100000..999999 is misread. Normalize numbers to #RRGGBB in a transport deserializer instead of guessing here.

crates/mezon-client/src/transport.rs
```diff
@@ -8073,6 +8158,39 @@ mod tests {
...
+            Some("49151")
+        );
+        assert_eq!(
+            parsed.content_tokens.embed[0].title.as_deref(),
+            Some("Saved")
+        );
+    }
+
+    #[test]
+    fn parse_message_content_tokens_recovers_embed_when_struct_parse_fails() {
```
huylexuan44, line 8382: color is a string here, so the struct parse succeeds and recovery never runs — this test does not cover the path it names. Use a real type mismatch, e.g. {"t":123,"embed":[{"title":"Recovered"}]}.

crates/mezon-client/src/transport.rs
```diff
@@ -770,10 +770,99 @@ fn json_to_i32(value: &serde_json::Value) -> Option<i32> {
...
+    }
+    if let Ok(serde_json::Value::String(inner)) = serde_json::from_str::<serde_json::Value>(trimmed)
+        && let Ok(tokens) = serde_json::from_str::<ApiMessageContent>(&inner)
+    {
+        return sanitize_embed_only_content_tokens(tokens, trimmed);
+    }
+    recover_message_content_tokens(trimmed)
+}
+
+fn sanitize_embed_only_content_tokens(
```
huylexuan44: Both branches are unreachable: t can never equal the whole raw JSON. This is an identity function — the real fix is string_or_number on ApiEmbed::color. Delete it and the mirrored else-if at line 834.

crates/mezon-client/src/transport.rs
```diff
@@ -770,10 +770,99 @@ fn json_to_i32(value: &serde_json::Value) -> Option<i32> {
...
+    if !tokens.embed.is_empty() && tokens.t.trim().is_empty() {
+        return tokens;
+    }
+    if !tokens.embed.is_empty() && tokens.t.trim() == raw.trim() {
+        tokens.t.clear();
+    }
+    tokens
+}
+
+fn recover_message_content_tokens(trimmed: &str) -> ApiMessageContent {
```
huylexuan44, line 839: Recovery drops mentions, mk, lk, hg, ej, fwd and all poll fields. It also duplicates message_search.rs:512 with a disjoint field set — consolidate both on this pub fn.

crates/mezon-store/src/messages.rs
```diff
@@ -5675,6 +5681,33 @@ mod tests {
     use crate::ids::UserId;
     use crate::message::MessageSpan;
 
+    #[test]
+    fn parse_embed_accent_accepts_decimal_embed_colors() {
+        let accent = parse_embed_accent(Some("49151")).expect("decimal color");
```
huylexuan44: 49151 is one of the rare values the 16-bit mask preserves, so this passes while the code is broken. Add a color with a red channel, e.g. 16711680 -> #FF0000.

crates/mezon-store/src/messages.rs
```diff
@@ -5142,11 +5142,17 @@ fn edit_content_spans(content: &str, content_tokens: OutgoingContent) -> EditTra
...
+    let raw = color?.trim();
+    if raw.is_empty() {
         return None;
     }
-    u32::from_str_radix(raw, 16).ok().map(gpui::rgb)
+    let hex = raw.trim_start_matches('#');
+    if hex.len() == 6 && hex.bytes().all(|b| b.is_ascii_hexdigit()) {
+        return u32::from_str_radix(hex, 16).ok().map(gpui::rgb);
+    }
+    raw.parse::<u32>()
```
huylexuan44: Parses raw, not hex, so "#49151" fails where "49151" works.

huylexuan44:
Review notes not tied to a line:

- CI is red from a stale base, not from this change. Lint/Windows fail on mezon-ui E0308 at channel_sidebar/mod.rs:1066 (`t(locale, ..)` should be `t(&locale, ..)`, already fixed on develop) and Test additionally hits a missing libsoup-3.0 on the runner. The branch is 28 commits behind develop — please rebase.
- Because the build died in mezon-ui, mezon-client and mezon-store never compiled in CI, so none of the five new tests have actually run.
- The `string_or_number` attribute on `ApiEmbed::color` is the real fix and it is correct. The problems are all in the decimal-color path in `parse_embed_accent` and in the hand-rolled recovery fallback — see inline comments.
- React (`EmbedMessage.tsx:24`) passes `color` straight to CSS, so it renders no accent for a numeric color. Decimal support is an improvement over the reference rather than 1:1 parity — worth a line in the PR description.

crates/mezon-client/src/transport.rs
```diff
@@ -770,10 +770,165 @@ fn json_to_i32(value: &serde_json::Value) -> Option<i32> {
...
+        }
+    }
+    out
+}
+
+fn recover_message_content_tokens(trimmed: &str) -> ApiMessageContent {
+    let root = match serde_json::from_str::<serde_json::Value>(trimmed) {
+        Ok(serde_json::Value::String(inner)) => serde_json::from_str(&inner).ok(),
+        Ok(value) => Some(value),
+        Err(_) => {
```
huylexuan44: Regression: this early return leaks raw JSON and makes extract_json_t_field_loose unreachable, so truncated search content now renders as raw. Fall through to the shared tail instead.

crates/mezon-client/src/transport.rs
```diff
@@ -1652,6 +1807,81 @@ pub fn is_here_user_id(user_id: &str) -> bool {
...
+    fn normalize_string(raw: &str) -> Option<String> {
+        let trimmed = raw.trim();
+        if trimmed.is_empty() {
+            return None;
+        }
+        let body = trimmed.strip_prefix('#').unwrap_or(trimmed);
+        if body.is_empty() {
+            return None;
+        }
+        if body.bytes().all(|b| b.is_ascii_digit()) {
```
huylexuan44: normalize_string strips # before this check, so any hex without A-F is read as decimal: #123456 becomes #01E240. Skip the decimal branch when a # prefix is present.

crates/mezon-client/src/transport.rs
```diff
@@ -770,10 +770,165 @@ fn json_to_i32(value: &serde_json::Value) -> Option<i32> {
...
+            tokens.t = text;
+        }
+        tokens.mentions = recover_content_token_array(map.get("mentions"));
+        tokens.hg = recover_content_token_array(map.get("hg"));
+        tokens.ej = recover_content_token_array(map.get("ej"));
+        tokens.mk = recover_content_token_array(map.get("mk"));
+        tokens.lk = recover_content_token_array(map.get("lk"));
+        tokens.vk = recover_content_token_array(map.get("vk"));
+        tokens.lky = recover_content_token_array(map.get("lky"));
+        tokens.embed = recover_embed_array(map.get("embed"));
```
huylexuan44, line 857: Recovery still drops answers, answer_counts, expire_at, is_closed, total_votes, poll_type, tp, cid and cvtt, so a poll that hits this path renders empty.

crates/mezon-client/src/transport.rs
```diff
@@ -1652,6 +1807,81 @@ pub fn is_here_user_id(user_id: &str) -> bool {
...
+            fn visit_none<E: de::Error>(self) -> Result<Self::Value, E> {
+                Ok(None)
+            }
+
+            fn visit_unit<E: de::Error>(self) -> Result<Self::Value, E> {
+                Ok(None)
+            }
+
+            fn visit_i64<E: de::Error>(self, v: i64) -> Result<Self::Value, E> {
+                Ok((v != 0).then(|| normalize_decimal(v.max(0) as u32)))
```
huylexuan44: Negative color yields #000000 because of v.max(0). Return None for v < 0.

crates/mezon-store/src/messages.rs
```diff
@@ -5675,6 +5675,48 @@ mod tests {
...
+    }
+
+    #[test]
+    fn parse_embed_accent_preserves_red_channel() {
+        let accent = parse_embed_accent(Some("#FF0000")).expect("red color");
+        assert_eq!(accent, gpui::rgb(0xff_00_00));
+    }
+
+    #[test]
+    fn build_embeds_maps_numeric_color_to_accent() {
```
huylexuan44: Both new build_embeds tests pass pre-normalized hex strings, which parse_embed_accent already handled before this PR. Assert from the raw numeric JSON to cover the new deserializer.

crates/mezon-store/src/message_search.rs
```diff
@@ -510,61 +510,7 @@ fn parse_search_hit_content(
...
-
-fn recover_content_token_array(value: Option<&serde_json::Value>) -> Vec<ContentToken> {
-    let Some(serde_json::Value::Array(items)) = value else {
-        return Vec::new();
-    };
-    items
-        .iter()
-        .filter_map(|item| serde_json::from_value::<ContentToken>(item.clone()).ok())
-        .collect()
+    parse_message_content_tokens(trimmed)
```
huylexuan44: parse_search_content_tokens is now a one-line alias - inline parse_message_content_tokens at the call site.

PR #141, opened 2026-07-23
Fix scroll bar for voice participants list

giachuvan:


crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1862,81 +1932,95 @@ fn render_focus_layout(
...
-            (0, total.min(16))
+        let total = strip_cells.len();
+        let strip_h = f32::from(strip_bounds.size.height);
+        let aside_h = carousel_aside_height(strip_h);
+        let tile_w = carousel_tile_width(aside_h);
+        let content_w = carousel_content_width(total, tile_w, gap);
+        let overflows = carousel_overflows(viewport_w, total, aside_h, gap);
+        let avatar_size = px((aside_h * 0.6).clamp(24., 80.));
+
+        let strip_tiles = {
```
phuocnguyenncc, line 1991: Virtualization was dropped here: `strip_tiles` now renders every participant each frame, where the old code only built the visible `start..end` window plus spacers. In a large call this rebuilds ~100 tiles per draw while the voice view repaints at video-frame rate. Please re-add the visible-range windowing — simpler now that `tile_w` is fixed (`tile_step = tile_w + gap`).

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -2206,8 +2207,9 @@ impl ChatLayout {
                         settings.camera_device_id.clone(),
                     )
                 };
+                let voice_theme = theme.clone();
```
phuocnguyenncc: `theme.clone()` runs on every render and `Theme` carries the full `tokens` struct, so this is a multi-KB copy per frame — the repo perf rules explicitly forbid cloning `Theme` per render. Resolve the theme inside `render_voice_channel` via `cx.theme()` once the `&mut` borrows end, or cache it on `ChatLayout`, instead of cloning at the call site.

phuocnguyenncc (approved):

PR #145, opened 2026-07-24
add: agent, sound join voice and fix UI issue

phuocnguyenncc:


crates/mezon-store/src/channel.rs
```diff
@@ -1907,7 +1928,7 @@ fn channel_from_desc(
     let voice_members = voice_ids
         .into_iter()
         .map(|uid| VoiceMember {
-            display_name: uid.to_string(),
+            display_name: String::new(),
```
huylexuan44, line 1931: Dropping the user_id fallback: members not in ClanMembersStore and not self now render blank name and avatar. Confirm store is loaded first or keep a fallback.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1146,20 +1147,28 @@ impl ChatLayout {
...
+                    match this.update(cx, |_, cx| VoiceStore::global(cx).read(cx).frame_store()) {
+                        Ok(store) => store,
+                        Err(_) => break,
+                    };
+                let Some(store) = store else {
+                    cx.background_executor().timer(VOICE_FRAME_FALLBACK).await;
+                    continue;
+                };
+                {
+                    let mut rx = store.frame_watch();
```
huylexuan44: Re-subscribes to frame_watch each iteration; a frame between the seq read and re-subscribe waits for the 200ms fallback. Subscribe once per acquired store.

crates/mezon-store/src/voice.rs
```diff
@@ -1587,6 +1660,7 @@ impl VoiceStore {
                     };
                 }
                 self.call_status = VoiceCallStatus::Stable;
+                self.play_join_sound(cx);
```
huylexuan44: play_join_sound sits in the Connected arm; if Connected re-fires on reconnect the chime replays. Gate on a first-connect flag.

crates/mezon-store/src/voice.rs
```diff
@@ -1325,6 +1364,40 @@ impl VoiceStore {
         self.moderate_participant(identity, ModerationAction::Kick, cx);
     }
 
+    pub fn agent_active(&self) -> bool {
+        self.participants.iter().any(|p| p.is_agent)
+    }
+
+    pub fn toggle_agent(&mut self, cx: &mut Context<Self>) {
```
huylexuan44, line 1373: toggle_agent has no in-flight guard; a rapid second click fires another add/disconnect. Disable or track a pending flag.

crates/mezon-voice/src/lib.rs
```diff
@@ -1101,7 +1101,11 @@ fn emit_participants(
         identity: local.identity().as_str().to_string(),
         name: display_name(&local.name(), local.identity().as_str()),
         is_local: true,
-        is_agent: local.kind() == ParticipantKind::Agent,
+        is_agent: {
+            #[allow(deprecated)]
+            let has_agent_permission = local.permission().is_some_and(|p| p.agent);
```
huylexuan44: Deprecated permission() flags any user with the agent flag as an agent. Add a comment on why, or rely on kind() == Agent.

assets/i18n/en.json
```diff
@@ -1025,7 +1025,10 @@
   "channelVoice.connecting": "Connecting…",
   "channelVoice.reconnecting": "Reconnecting…",
   "channelVoice.weakNetwork": "Weak network",
-  "channelVoice.everyoneWaiting": "Everyone's waiting inside",
+  "channelVoice.everyoneWaiting": "Everyone is waiting for you inside",
+  "channelVoice.addAgent": "Add AI agent",
```
huylexuan44, line 1029: New agent keys are en-only; the other 10 locales fall back to English. Add translations for parity.

PR #150, opened 2026-07-27
feat: event list view

duocphungvanNCC:
> `Description:`

**env development:** window
- list of event user can join
- check status interest of event
- check type of event
- badge count event

image


crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,332 @@
...
+        cx.spawn(async move |this, cx| {
+            loop {
+                cx.background_executor().timer(Duration::from_secs(1)).await;
+                if this
+                    .update(cx, |this, cx| {
+                        let changed = this.remove_expired();
+                        for clan_id in changed {
+                            cx.emit(EventsEvent::Changed { clan_id });
+                        }
+                        cx.notify();
```
huylexuan44: notify() is outside the changed check, so this fires every second for the whole session from app start, even with zero events. ChannelSidebar observes this store, so that is a full sidebar re-render at 1Hz. Move it inside if !changed.is_empty(), and prefer sleeping until the next end_time_seconds over a 1s poll.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -1132,6 +1139,48 @@ fn render_banner_and_events(
                 crate::router::navigate(cx, crate::router::Route::ClanMembers { clan_id });
             }
         });
+    let current_user = BadgeService::global(cx).read(cx).current_user_id(cx);
+    let event_count = members_clan_id.map_or(0, |clan_id| {
+        EventsStore::global(cx)
+            .read(cx)
+            .visible_events(clan_id, current_user, cx)
```
huylexuan44: visible_events ends in .cloned().collect(), so every sidebar render deep-clones 5 Strings + a Vec per event just to call .len(). Add a count method that runs the same filter with .count().

crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,332 @@
...
+            events: HashMap::new(),
+            loaded: HashSet::new(),
+            loading: HashSet::new(),
+            api,
+            _connection_watch: connection_watch,
+            _clock_task: clock_task,
+        }
+    }
+
+    pub fn global(cx: &App) -> Entity<Self> {
```
huylexuan44, line 176: EventsStore has no reset() and no try_global(), and is not in LoginStore::logout_teardown, so a previous account keeps its clan events after logout. Add both and wire it in next to WebhookStore.

crates/mezon-ui/src/chat/clan_events_page.rs
```diff
@@ -0,0 +1,673 @@
...
+impl ClanEventsModal {
+    pub fn new(
+        clan_id: ClanId,
+        settings: Entity<Settings>,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) -> Self {
+        let focus_handle = cx.focus_handle();
+        window.focus(&focus_handle, cx);
+        cx.observe(&EventsStore::global(cx), |_, _, cx| cx.notify())
```
huylexuan44: A modal is transient, so .detach() leaks one subscription per open for all three observers. Store them in Subscription fields like ChannelSidebar does in this same PR.

crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,332 @@
...
+        let RealtimeEvent::ClanEventCreated(event) = realtime else {
+            return;
+        };
+        let clan_id = ClanId(event.clan_id);
+        let events = self.events.entry(clan_id).or_default();
+        match event.action {
+            EVENT_CREATED => {
+                let item = ClanEventItem::from_realtime(event);
+                if let Some(existing) = events.iter_mut().find(|item| item.id == event.event_id) {
+                    *existing = item;
```
huylexuan44: from_realtime resets user_ids to just the creator, so a re-delivered CREATED wipes the interested list. Use apply_realtime here like the UPDATED arm does.

crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,332 @@
...
+    ) -> Vec<ClanEventItem> {
+        let channels = ChannelList::global(cx);
+        let channels = channels.read(cx);
+        let now = chrono::Utc::now().timestamp().max(0) as u32;
+        self.events(clan_id)
+            .iter()
+            .filter(|event| {
+                event.event_status != EVENT_COMPLETED
+                    && (event.end_time_seconds == 0 || event.end_time_seconds > now)
+                    && (!event.is_private || Some(event.creator_id) == current_user)
```
huylexuan44: React eventManagement.slice.ts has no such filter. This hides private events from invited members, making the privateRoom / invite / copyLink branch of event_card dead for everyone but the creator. Confirm what list_events already scopes server-side.

crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,332 @@
...
+                        let items = items
+                            .into_iter()
+                            .map(ClanEventItem::from_api)
+                            .filter(|event| event.event_status != EVENT_COMPLETED)
+                            .collect();
+                        this.events.insert(clan_id, items);
+                        this.loaded.insert(clan_id);
+                        this.remove_expired();
+                    }
+                    Err(error) => tracing::warn!(%error, %clan_id, "failed to load events"),
```
huylexuan44: loaded is only inserted on Ok, so every later rebuild_items re-fetches after a failure with no backoff and no failed-state UI. Also no TTL: other server stores here gate on CACHE_TTL 20min.

crates/mezon-ui/src/chat/clan_events_page.rs
```diff
@@ -0,0 +1,673 @@
...
+            details = details.child(
+                img(event.logo.clone())
+                    .w(px(200.))
+                    .h(px(110.))
+                    .rounded(px(3.))
+                    .object_fit(gpui::ObjectFit::Contain),
+            );
+        }
+
+        let action = |icon, label: &'static str, group: SharedString| {
```
huylexuan44: Interested, Share, End event, ThreeDot, openLink, invite and copyLink all render with cursor_pointer and hover but have no on_click. Wire them, or drop the affordances so they do not look live.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -1132,6 +1139,48 @@ fn render_banner_and_events(
...
+            .len()
+    });
+    let settings = sidebar
+        .upgrade()
+        .map(|entity| entity.read(cx).settings.clone());
+    let event_label = if event_count == 1 {
+        "1 Event".to_string()
```
huylexuan44: Hardcoded English. eventCreator.actions.event_one and event_other exist and are already used in the modal.

crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,332 @@
...
+
+use crate::realtime::{RealtimeDispatch, RealtimeKind};
+use crate::{ChannelId, ChannelList, ClanId, UserId};
+
+const EVENT_CREATED: i32 = 1;
+const EVENT_UPDATED: i32 = 2;
+const EVENT_DELETED: i32 = 3;
+const EVENT_INTERESTED: i32 = 4;
+const EVENT_UNINTERESTED: i32 = 5;
+const EVENT_COMPLETED: i32 = 3;
```
huylexuan44: EVENT_DELETED is also 3 but belongs to the action enum, not status. Correct today, but a trap sitting in one flat const block. Split into two named enums.

crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,332 @@
...
+                }
+            }
+            EVENT_UNINTERESTED => {
+                if let Some(existing) = events.iter_mut().find(|item| item.id == event.event_id) {
+                    existing.user_ids.retain(|user| user.0 != event.user_id);
+                }
+            }
+            _ => self.fetch(clan_id, cx),
+        }
+        self.remove_expired();
```
huylexuan44: Return value discarded here and in fetch, so expiries in other clans skip their Changed emit.

huylexuan44:
Review summary — well scoped, no unrelated changes, and the domain constants are right: EEventAction 1-5 and EEventStatus.COMPLETED = 3 both match libs/utils/src/lib/types/index.ts, and all 15 eventCreator/eventMenu i18n keys resolve in assets/i18n/en.json.

Two to fix before merge:

- spawn_clock notifies unconditionally at 1Hz from app start, and ChannelSidebar observes the store, so the sidebar re-renders every second all session even with zero events.
- EventsStore has no reset/try_global and is missing from logout_teardown, so clan events survive an account switch.

Then the cheap ones: Subscription fields instead of .detach() in the modal, a non-cloning count for the sidebar badge, and apply_realtime on the CREATED collision path.

Minor, not blocking: hardcoded hex colors (a855f7, ef4444, f97316, 3b82f6, 22d3ee, facc15) bypass theme.tokens so they will not track the active theme; img() loads the raw logo URL instead of going through util::imgproxy for a 200x110 slot; and there is a commented-out create-event Button that rustfmt has mangled into a trailing comment block.

I did not run a build — nothing looked compile-suspicious and every referenced method and IconName variant resolves.

crates/mezon-store/src/events.rs
```diff
@@ -0,0 +1,431 @@
...
+                    if this.update(cx, |this, cx| this.refresh_loaded(cx)).is_err() {
+                        break;
+                    }
+                }
+                connected_before = connected;
+            }
+        })
+    }
+
+    fn spawn_clock(cx: &mut Context<Self>) -> Task<()> {
```
huylexuan44: Drop the 1s poll: React has no client timer here - the server pushes event_status transitions and views check time at render. Arm a single timer to the next start/end boundary instead, re-armed on fetch/realtime changes, so the store does zero work when idle.

huylexuan44: For reference this codebase already has both alternatives: PresenceStore::schedule_typing_sweep (1s sweep that starts on first typing entry and stops when the map is empty) and StreamStore bump_controls_visible (timer armed to an exact deadline). Either beats a session-long unconditional tick.

PR #151, opened 2026-07-27
Feat/join streaming channel

giachuvan:


crates/mezon-store/src/config.rs
```diff
@@ -98,21 +98,21 @@ impl AppConfig {
...
         Self {
-            api_host: "dev-mezon.nccsoft.vn".into(),
-            api_port: 8088,
+            api_host: "api.mezon.ai".into(),
+            api_port: 443,
             api_secure: true,
-            api_key: "defaultkey".into(),
-            api_gw_host: "dev-mezon.nccsoft.vn".into(),
-            api_gw_port: 8088,
+            api_key: "HTTP3m3zonPr0dkey".into(),
```
huylexuan44: Live prod API key committed to a PUBLIC repo. These fields were deliberately emptied in PR #76; inject via NX_CHAT_APP_API_KEY at build time instead.

crates/mezon-store/src/config.rs
```diff
@@ -130,7 +130,7 @@ impl AppConfig {
             imgproxy_base_url: "https://imgproxy.komu.vn".into(),
             imgproxy_key: "_AEhOrrckkG-NjqIdVLtzc-dtLFuE4u6ClM0P46ICEY".into(),
 
-            klipy_key: String::new(),
+            klipy_key: "j8Mh0W38PmhdHoX7mQYJ0VSwkHNMJZ9tkPeeZuz3hukOcAxueqUjOuTpiJdrar6p".into(),
```
huylexuan44: Live Klipy key committed to a PUBLIC repo. Revert to String::new() and inject via NX_* at build time.

crates/mezon-store/src/config.rs
```diff
@@ -98,21 +98,21 @@ impl AppConfig {
     /// Development defaults (matches pre-env hardcoded values).
     pub fn dev_defaults() -> Self {
         Self {
-            api_host: "dev-mezon.nccsoft.vn".into(),
-            api_port: 8088,
+            api_host: "api.mezon.ai".into(),
```
huylexuan44: dev_defaults rewritten to prod but tests were not updated: dev_defaults_match_legacy_constants and dev_defaults_use_dev_tcp_port now fail. just test is red.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1181,6 +1244,55 @@ impl ChatLayout {
...
+        }
+        if self._stream_frame_pump.is_some() {
+            return;
+        }
+        self._stream_frame_pump = Some(cx.spawn(async move |this, cx| {
+            loop {
+                cx.background_executor().timer(STREAM_FRAME_INTERVAL).await;
+                let stepped = this.update(cx, |this, cx| {
+                    let seq = this.stream_store.read(cx).frame_store().publish_seq();
+                    if seq > 0 {
```
huylexuan44: publish_seq is monotonic, so seq > 0 is true forever after the first frame: this notifies ChatLayout at 30Hz even with no new frame. Track last_seq and use frame_store.frame_watch() like sync_voice_frame_pump does.

crates/mezon-stream/src/session.rs
```diff
@@ -0,0 +1,384 @@
...
+    runtime.block_on(async {
+        let key = 1u64;
+        let mut stream = NativeVideoStream::new(video_track);
+        while let Some(frame) = stream.next().await {
+            let buffer = frame.buffer.to_i420();
+            let width = buffer.width();
+            let height = buffer.height();
+            let (sy, su, sv) = buffer.strides();
+            let (y, u, v) = buffer.data();
+            let mut bgra = vec![0u8; width as usize * height as usize * 4];
```
huylexuan44: Allocates a fresh multi-MB Vec every frame (~8MB x30/s at 1080p). Reuse the buffer via clear()+resize() like mezon-voice/src/lib.rs:1026.

crates/mezon-stream/src/session.rs
```diff
@@ -0,0 +1,384 @@
...
+                y,
+                u,
+                v,
+                sy as usize,
+                su as usize,
+                sv as usize,
+                width as usize,
+                height as usize,
+            );
+            frame_store.publish(key, width, height, bgra);
```
huylexuan44: publish() returns the recycled buffer; discarding it drops the recycling pool. Assign it back into bgra.

crates/mezon-stream/src/session.rs
```diff
@@ -0,0 +1,384 @@
...
+async fn run_session(
+    config: StreamSessionConfig,
+    frame_store: Arc<VideoFrameStore>,
+    audio: Arc<StreamAudioOutput>,
+    stop_rx: Receiver<()>,
+    event_tx: Sender<StreamEvent>,
+) {
+    if let Err(err) = run_session_inner(config, frame_store, audio, stop_rx, event_tx.clone()).await
+    {
+        tracing::warn!("stream session ended: {err:#}");
```
huylexuan44: err may embed the ws URL, which carries the auth token in a query param. This is logged and also sent as StreamEvent::Error into the on-screen error panel. Map connect failures to a fixed message.

crates/mezon-store/src/stream.rs
```diff
@@ -0,0 +1,646 @@
...
+            cx.notify();
+            return;
+        };
+        if config.stream_ws_url.is_empty() {
+            self.phase = StreamPhase::Error("Stream server not configured".into());
+            self.error_message = Some("Stream server not configured".into());
+            cx.notify();
+            return;
+        }
+        if self.members_for_channel(channel_id).is_empty() {
```
huylexuan44: Silent return: no phase change, no error, no notify. React handleChannelClick has no such gate. Members come from one clan-scoped fetch, and handle_streaming_started never refetches, so Join stays disabled until you switch clans and back.

crates/mezon-store/src/stream.rs
```diff
@@ -0,0 +1,646 @@
...
+            return;
+        };
+        let channel_id = ChannelId(channel_id);
+        if let Ok(user_id) = e.streaming_user_id.parse::<i64>() {
+            self.members.remove(&(channel_id, UserId(user_id)));
+        }
+        cx.notify();
+    }
+
+    fn handle_streaming_started(&mut self, event: &RealtimeEvent, cx: &mut Context<Self>) {
```
huylexuan44, line 705: Refetch members here so the Join button unsticks when a broadcast starts while the clan is already open.

crates/mezon-store/src/stream.rs
```diff
@@ -0,0 +1,646 @@
...
+                self.phase = StreamPhase::Error(err.to_string());
+                self.error_message = Some(err.to_string());
+                cx.notify();
+                return;
+            }
+        };
+        let events = stream_session.events().clone();
+        self.session = Some(stream_session);
+
+        let _label = channel_label.to_string();
```
huylexuan44: Dead binding, remove.

crates/mezon-store/src/stream.rs
```diff
@@ -0,0 +1,646 @@
...
+    fn disconnect_session(&mut self) {
+        if let Some(session) = self.session.take() {
+            session.disconnect();
+        }
+        self._session_task = None;
+        self._join_timeout = None;
+        self.remote_video = false;
+        self.playback_blocked = false;
+        self.join_started = None;
+        self.frame_store.remove(1);
```
huylexuan44: Magic key 1 duplicated across crates (session.rs:334 publishes with the same literal). Extract a shared named constant.

crates/mezon-stream/src/session.rs
```diff
@@ -0,0 +1,384 @@
...
+    ) -> anyhow::Result<Self> {
+        let audio = Arc::new(
+            StreamAudioOutput::start(output_device_id, volume, muted)
+                .context("stream audio output failed")?,
+        );
+        let (stop_tx, stop_rx) = flume::bounded(1);
+        let (event_tx, event_rx) = flume::unbounded();
+        let audio_for_thread = audio.clone();
+        std::thread::spawn(move || {
+            let runtime = tokio::runtime::Builder::new_multi_thread()
```
huylexuan44, line 80: Spawns a full multi-thread tokio runtime per session, plus one more runtime+thread each in pump_video_track and pump_audio_track. Share one runtime; the pumps need dedicated threads, not their own runtimes.

crates/mezon-stream/src/session.rs
```diff
@@ -0,0 +1,384 @@
...
+            }
+            msg = ws_rx.next() => {
+                let Some(msg) = msg else { break; };
+                let msg = msg.context("websocket read failed")?;
+                if msg.is_close() {
+                    break;
+                }
+                let Message::Text(text) = msg else { continue; };
+                let inbound: InboundMessage = serde_json::from_str(&text)
+                    .context("invalid websocket payload")?;
```
huylexuan44: One malformed frame tears down the whole session. ice_candidate below uses .ok() instead. Log and continue rather than propagating.

crates/mezon-store/Cargo.toml
```diff
@@ -13,6 +13,7 @@ mezon-proto.workspace = true
 mezon-i18n.workspace = true
 mmn-client.workspace = true
 mezon-voice.workspace = true
+mezon-stream = { path = "../mezon-stream" }
```
huylexuan44: Ad-hoc path dep; every other crate uses workspace.dependencies + .workspace = true. Add mezon-stream to the workspace table.

huylexuan44:
Review summary — 2 blockers before merge:

- crates/mezon-store/src/config.rs is unrelated to this feature and re-adds live prod secrets to a PUBLIC repo (emptied on purpose in PR #76), and breaks 2 config tests. Please revert that file entirely and use build-time NX_* vars to test against prod.
- mezon-stream is missing from justfile pkgs and scripts/fmt.sh PKGS, so the whole new crate is outside clippy -D warnings, nextest and the fmt gate. Add -p mezon-stream to both, then run just lint + just test.

Other notes: no reconnect on a transient ws drop; stream_playback.rs largely duplicates mezon-voice/src/audio.rs (track-keyed mixer could take a single-track mode); a lot of view state (show_chat, controls_visible, fullscreen, volume) lives in the global StreamStore, which is what forces the broad ChatLayout notify.

The hardcoded Google STUN matches React StreamContext.tsx:79, so it is faithful parity, not a porting bug. Flagging only because config already carries webrtc_ice_servers_* (TURN relay.mezon.ai) that voice.rs:1675 uses, and no TURN means playback fails behind symmetric NAT. Worth a product decision, not a change in this PR.

A short description and test plan on the PR would help — this is +3120 lines and a new crate.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1181,6 +1244,67 @@ impl ChatLayout {
...
+                    let changed = std::pin::pin!(rx.changed());
+                    let fallback =
+                        std::pin::pin!(cx.background_executor().timer(STREAM_FRAME_FALLBACK));
+                    matches!(
+                        futures::future::select(changed, fallback).await,
+                        futures::future::Either::Left((Ok(()), _))
+                    )
+                };
+                if !frame_published {
+                    break;
```
huylexuan44: Frame pump ends here and never restarts: the outer loop from sync_voice_frame_pump was dropped, and the completed Task stays Some so the is_some() guard blocks respawn. Any 200ms gap freezes video permanently.

crates/mezon-ui/src/chat/stream.rs
```diff
@@ -0,0 +1,1124 @@
...
+                .size_full()
+                .object_fit(ObjectFit::Contain)
+                .into_any_element();
+        }
+    }
+    div().size_full().bg(gpui::rgb(0x111111)).into_any_element()
+}
+
+fn frame_to_render_image(frame: &VideoFrameData) -> Option<Arc<RenderImage>> {
+    let buffer = image::RgbaImage::from_raw(frame.width, frame.height, frame.bgra.clone())?;
```
huylexuan44: New RenderImage per render means a fresh ImageId, and Drop only recycles the buffer - the atlas tile leaks every render. Use take_new + new_recyclable + with_id + queue_atlas_replace like VoiceStore::render_frame.

crates/mezon-ui/src/chat/channel_header.rs
```diff
@@ -995,6 +1003,71 @@ impl Render for ChatHeader {
     }
 }
 
+fn render_stream_chat_sidebar_header(
+    theme: &Theme,
+    name: &SharedString,
+    _cx: &App,
+) -> impl IntoElement {
+    let label = if name.len() > 30 {
+        format!("{}...", &name[..30])
```
huylexuan44: name.len() is bytes - this panics on multibyte channel names (Vietnamese hits it at ~11 chars). Use the chars()-based truncate_label from chat/stream.rs.

crates/mezon-ui/src/chat/stream.rs
```diff
@@ -0,0 +1,1124 @@
...
+) -> AnyElement {
+    let subtitle = if members.is_empty() {
+        mezon_i18n::t(locale, "channelStream.noOneInStream")
+    } else {
+        mezon_i18n::t(locale, "channelStream.everyoneWaiting")
+    };
+    let join_text = if joining {
+        SharedString::from("…")
+    } else {
+        SharedString::from(mezon_i18n::t(locale, "channelStream.joinStream"))
```
huylexuan44, line 534: All 8 new channelStream.* and channelTopbar.tooltips.* keys are missing from every assets/i18n locale including en, so t() returns the raw key and the join button reads channelStream.joinStream.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1339,11 +1473,18 @@ impl Render for ChatLayout {
...
-        } else {
-            px(0.)
-        };
+        let stream_connected_bar = crate::chat::stream::render_stream_connected_bar(
+            cx.theme(),
+            self.stream_store.read(cx),
+            self.stream_store.clone(),
+            cx,
+        );
+        let nav_bottom_pad = self.voice_mini_bar_height
```
huylexuan44: Dropping the voice_mini_bar.is_some() guard regresses nav padding: voice_mini_bar_height is only written from the canvas inside the bar and never reset to 0, so phantom padding stays after a call ends.

crates/mezon-ui/src/chat/stream.rs
```diff
@@ -0,0 +1,1124 @@
...
+    let stream_entity = stream.clone();
+    let auth_entity = auth.clone();
+    let channel_id = channel.id;
+    let clan_id = channel.clan_id;
+    let channel_label = channel.name.clone();
+    let clan_name = channel.clan_name.clone();
+    let error_message = store.error_message().map(str::to_owned);
+    let output_device = output_device_id.clone();
+
+    let body = if let Some(message) = error_message {
```
huylexuan44, line 44: error_message is store-global and checked before anything else, so failing on stream A then navigating to stream B shows A error. Clear it in sync_stream_session on channel change.

crates/mezon-ui/src/chat/stream.rs
```diff
@@ -0,0 +1,1124 @@
...
+    let control_opacity = if controls_visible { 100. } else { 0. };
+
+    div()
+        .relative()
+        .flex_1()
+        .flex()
+        .flex_col()
+        .min_h_0()
+        .on_mouse_move(move |_, _, cx| {
+            stream_bump.update(cx, |store, cx| store.bump_controls_visible(cx));
```
huylexuan44, line 656: bump_controls_visible allocates and installs a fresh cx.spawn timer on every mouse-move event. Store a deadline Instant checked by one long-lived timer instead.

crates/mezon-store/src/stream.rs
```diff
@@ -0,0 +1,644 @@
...
+            username: session.username.clone(),
+            token: session.ws_credential().to_string(),
+            clan_id: clan_id.to_string(),
+            channel_id: channel_id.to_string(),
+            user_id: session.user_id.clone(),
+            stream_id: channel_id.to_string(),
+        };
+        let frame_store = self.frame_store.clone();
+
+        let stream_session = match StreamSession::start(
```
huylexuan44: StreamSession::start runs cpal device enumeration and build_output_stream synchronously on the UI thread. Move audio setup into the session thread and report failure via StreamEvent::Error.

crates/mezon-stream/src/session.rs
```diff
@@ -0,0 +1,399 @@
...
+    .await?;
+    send_ws(&mut ws_tx, OutboundMessage::get_channels()).await?;
+
+    let pc = Arc::new(pc);
+    let mut live = false;
+
+    loop {
+        tokio::select! {
+            _ = stop_rx.recv_async() => {
+                pc.close();
```
huylexuan44: pc.close() only runs on the stop path - the ws-close, None and ? error exits skip it, leaving the pump threads blocked on a runtime that is being dropped. Close unconditionally after the loop.

crates/mezon-store/src/stream.rs
```diff
@@ -0,0 +1,644 @@
...
+use mezon_stream::{STREAM_FRAME_KEY, StreamEvent, StreamSession, StreamSessionConfig};
+use mezon_voice::VideoFrameStore;
+
+use crate::AppConfig;
+use crate::AuthState;
+use crate::ChannelType;
+use crate::ids::{ChannelId, ClanId, UserId};
+use crate::realtime::{RealtimeDispatch, RealtimeKind};
+
+const STREAMING_CHANNEL_TYPE: i32 = 6;
```
huylexuan44, line 20: React sends no state for listStreamingChannelUsers. If state=1 is wrong the list comes back empty and the Join button stays permanently disabled. Confirm with the server or drop the param.

crates/mezon-store/src/stream.rs
```diff
@@ -0,0 +1,644 @@
...
+                    this.phase = StreamPhase::Error("Join timed out".into());
+                    this.error_message = Some("Join timed out".into());
+                    cx.notify();
+                }
+            })
+            .ok();
+        }));
+    }
+
+    pub fn leave_stream(&mut self, cx: &mut Context<Self>) {
```
huylexuan44, line 574: leave_stream does not remove your own id from members or reset show_chat, unlike React handleLeaveChannel - your avatar lingers in the member row and the chat panel stays open.

PR #193, opened 2026-07-31
Fix image resolution in pin

giachuvan:


crates/mezon-ui/src/chat/pinned_popover.rs
```diff
@@ -979,20 +1001,22 @@ fn render_pin_image_attachment(
...
+    } else {
+        att.proxied_src.clone()
+    };
     div()
         .mt_1()
-        .w(px(width))
-        .h(px(height))
-        .max_w_full()
-        .flex_shrink_0()
+        .max_w(px(ATTACHMENT_PREVIEW_MAX))
```
huylexuan44: Box has no reserved size, so it is 0-height until decode and rows jump in the measure_all list; React uses a fixed 120x120 square (rounded 4px) — use w/h(px(120.)) for parity and no jank.

crates/mezon-ui/src/chat/pinned_popover.rs
```diff
@@ -979,20 +1001,22 @@ fn render_pin_image_attachment(
     att: &MessageAttachment,
     image_cache: Entity<LruImageCache>,
 ) -> gpui::AnyElement {
-    let width = att.display_width.clamp(1., 280.);
-    let height = att.display_height.clamp(1., 200.);
+    let src = if att.proxied_src.is_empty() {
```
huylexuan44, line 1008: Dead branch: render_pin_body still filters on !att.proxied_src.is_empty(), so this url fallback never runs — relax that gate or drop the fallback.

crates/mezon-ui/src/chat/message/pin_confirm_modals.rs
```diff
@@ -161,6 +165,15 @@ impl ConfirmPinMessageModal {
                 .to_string()
                 .into(),
             avatar_image_cache: crate::image_cache::shared_avatar_cache(cx),
+            message_image_cache: cx.new(|cx| {
+                LruImageCache::message(
```
huylexuan44: 48-item/32MB timeline budget for a one-message preview is oversized — a small cap (e.g. 8 items / 8MB) fits the modal scope.

PR #200, opened 2026-07-31
Fix add jump to messages inbox

giachuvan:


crates/mezon-store/src/messages.rs
```diff
@@ -1562,6 +1562,7 @@ impl MessagesStore {
                         removed_bottom: dropped_bottom,
                     });
                 }
+                this.try_consume_pending_jump(cx);
```
huylexuan44: The 3 early returns above (fetch error, cache miss, older.is_empty()) skip this - pending_jump stays armed forever.

crates/mezon-store/src/messages.rs
```diff
@@ -1726,6 +1727,7 @@ impl MessagesStore {
                         removed_bottom: 0,
                     });
                 }
+                this.try_consume_pending_jump(cx);
```
huylexuan44: Same as load_more: fetch error, cache miss and newer.is_empty() return before this, leaving pending_jump armed.

crates/mezon-store/src/messages.rs
```diff
@@ -1751,7 +1753,7 @@ impl MessagesStore {
         let Some((channel_id, message_id)) = self.pending_jump else {
             return;
         };
-        if self.active_channel_id != Some(channel_id) || self.loading {
+        if self.active_channel_id != Some(channel_id) || self.loading || self.loading_more {
```
huylexuan44, line 1747: Gate now waits on loading_more, but the initial-fetch Err arm (line 4182) clears loading without consuming.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -720,78 +720,66 @@ fn schedule_inbox_jump(
...
+    let channel_id = channel_id_str.parse::<ChannelId>().ok()?;
+    let clan_id = notification
+        .effective_clan_id()
+        .and_then(|id| id.parse::<ClanId>().ok())
+        .or_else(|| {
+            ChannelList::global(cx)
+                .read(cx)
+                .clan_id_for_channel(channel_id)
+        })?;
+    if clan_id.is_zero() {
```
huylexuan44: Unreachable: effective_clan_id and clan_id_for_channel both exclude 0, so DM notifications get no jump at all.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -903,7 +890,9 @@ fn render_notification_item(
     let category = notification.category;
     let id: SharedString = notification.id.clone().into();
     let copy_text = notification_copy_text(&notification);
-    let show_jump = tab == InboxTab::Mentions;
+    let show_jump = matches!(tab, InboxTab::Mentions | InboxTab::Messages)
+        && notification.effective_message_id().is_some()
+        && notification_jump_route(&notification, cx).is_some();
```
huylexuan44: notification_jump_route runs per row in render and again on click. Precompute it into NotificationRowView.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -903,7 +890,9 @@ fn render_notification_item(
     let category = notification.category;
     let id: SharedString = notification.id.clone().into();
     let copy_text = notification_copy_text(&notification);
-    let show_jump = tab == InboxTab::Mentions;
+    let show_jump = matches!(tab, InboxTab::Mentions | InboxTab::Messages)
```
huylexuan44: React shows Jump only on MENTIONS (AllNotificationItem.tsx:161). Confirm adding it to Messages is intended.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -720,78 +720,66 @@ fn schedule_inbox_jump(
...
+        .and_then(|id| id.parse::<ClanId>().ok())
+        .or_else(|| {
+            ChannelList::global(cx)
+                .read(cx)
+                .clan_id_for_channel(channel_id)
+        })?;
+    if clan_id.is_zero() {
+        Some(Route::DirectMessage {
+            direct_id: channel_id,
+            message_type: notification.channel_type.to_string(),
```
huylexuan44: channel_type may be a stream mode (DM=4/Group=3), not ChannelType, so groups open as DMs. Derive it like React.

crates/mezon-client/src/inbox.rs
```diff
@@ -597,6 +676,51 @@ mod tests {
         assert_eq!(preview.content, "hello");
     }
 
+    #[test]
+    fn enrich_message_preview_reads_channel_message_bytes() {
```
huylexuan44: This never reaches enrich_message_preview_from_raw: parse_notification_content already sets message_id at line 399.

crates/mezon-client/src/inbox.rs
```diff
@@ -231,6 +235,69 @@ fn preview_from_fcm(fcm: api::DirectFcmProto) -> InboxMessagePreview {
...
+        mention_spans: Vec::new(),
+    }
+}
+
+fn enrich_message_preview_from_raw(preview: &mut InboxMessagePreview, raw: &[u8]) {
+    if is_valid_inbox_message_id(&preview.message_id) {
+        return;
+    }
+    if !matches!(raw.first(), Some(b'{') | Some(b'['))
+        && let Ok(message) = api::ChannelMessage::decode(raw)
```
huylexuan44: Unreachable: parse_notification_content:399 already claims these bytes. Drop this block or move the decode here.

crates/mezon-store/src/messages.rs
```diff
@@ -3966,17 +3971,56 @@ impl MessagesStore {
         else {
             return;
         };
+        self.open_channel_record(channel.clan_id, channel_id, &channel, cx);
+    }
+
+    pub fn open_channel_in_clan(
```
huylexuan44, line 4053: 20 lines duplicated from open_channel. Make open_channel delegate to this with the active clan.

huylexuan44:
Two blockers, both undercutting the feature itself:

- pending_jump is only consumed on the happy path. Every early return in load_more (1511/1531/1543), load_more_bottom (1644/1669/1697) and the initial-fetch Err arm (4182) clears the loading flag without consuming, so a Jump queued during an in-flight load never fires until the next channel switch. The empty-page case is the common one. Route all flag clearing through one finish_loading_more helper.
- The Route::DirectMessage branch in notification_jump_route is unreachable, so DM notifications get no jump button and the channel_type change is dead. Confirm whether Messages-tab rows can ever have clan_id 0.

Test job failure is pre-existing and unrelated: crates/vendor/libspa/build.rs:11 uses deprecated cc::Build::shared_flag under -D warnings. It means the two new tests never actually ran in CI.

Good parts: typed ChannelId/MessageId through schedule_inbox_jump, and the clan-explicit channel lookup matching the existing convention.

PR #202, opened 2026-07-31
Fix/mezon icon for window upstream

giachuvan:


crates/mezon-native/src/window_icon.rs
```diff
@@ -0,0 +1,70 @@
...
+    let module = unsafe { GetModuleHandleW(None)? };
+    let load = |cx: i32, cy: i32| -> windows::core::Result<_> {
+        unsafe {
+            LoadImageW(
+                Some(module.into()),
+                PCWSTR(1 as _),
+                IMAGE_ICON,
+                cx,
+                cy,
+                LR_DEFAULTCOLOR | LR_SHARED,
```
huylexuan44: LR_SHARED can return gpui's cached default-size handle for resource 1, defeating the fix. Use LoadIconWithScaleDown instead.

crates/mezon-app/src/main.rs
```diff
@@ -581,7 +581,9 @@ fn run_app(lock: SingleInstance, initial_url: Option<String>) {
                     window.window_handle().ok().map(|handle| handle.as_raw())
                 })
             {
-                mezon_native::badge::set_main_window_hwnd(win32.hwnd.get());
+                let hwnd = win32.hwnd.get();
+                mezon_native::badge::set_main_window_hwnd(hwnd);
+                mezon_native::window_icon::apply_dpi_aware_icons(hwnd);
```
huylexuan44, line 586: Only the main window. Other GPUI windows share class Zed::Window - also SetClassLongPtrW GCLP_HICON/GCLP_HICONSM to cover them.

crates/mezon-native/src/window_icon.rs
```diff
@@ -0,0 +1,70 @@
...
+    use windows::Win32::System::LibraryLoader::GetModuleHandleW;
+    use windows::Win32::UI::HiDpi::{GetDpiForWindow, GetSystemMetricsForDpi};
+    use windows::Win32::UI::WindowsAndMessaging::{
+        ICON_BIG, ICON_SMALL, IMAGE_ICON, LR_DEFAULTCOLOR, LR_SHARED, LoadImageW, SM_CXICON,
+        SM_CXSMICON, SendMessageW, WM_SETICON,
+    };
+    use windows::core::PCWSTR;
+
+    let hwnd = HWND(hwnd as *mut core::ffi::c_void);
+    let dpi = unsafe { GetDpiForWindow(hwnd) };
```
huylexuan44, line 87: Computed once at startup; icons go stale on WM_DPICHANGED when the window moves to a different-DPI monitor. Add a TODO.

crates/mezon-native/src/window_icon.rs
```diff
@@ -0,0 +1,70 @@
...
+    if dpi == 0 {
+        return Err(windows::core::Error::from_win32());
```
huylexuan44: GetDpiForWindow/GetSystemMetricsForDpi do not set last-error; from_win32() can log 'completed successfully'. Same at line 34.

crates/mezon-native/src/window_icon.rs
```diff
@@ -0,0 +1,70 @@
...
+                PCWSTR(1 as _),
+                IMAGE_ICON,
+                cx,
+                cy,
+                LR_DEFAULTCOLOR | LR_SHARED,
+            )
+        }
+    };
+
+    let big_icon = load(big, big)?;
```
huylexuan44: load(big, big) uses the X metric for height; pass SM_CYICON/SM_CYSMICON for cy.

crates/mezon-native/src/window_icon.rs
```diff
@@ -0,0 +1,70 @@
...
+    if big <= 0 || small <= 0 {
+        return Err(windows::core::Error::from_win32());
+    }
+
+    let module = unsafe { GetModuleHandleW(None)? };
+    let load = |cx: i32, cy: i32| -> windows::core::Result<_> {
+        unsafe {
+            LoadImageW(
+                Some(module.into()),
+                PCWSTR(1 as _),
```
huylexuan44: PCWSTR(1 as _) is the resource id from app.rc '1 ICON'. Extract a named const with a comment.

huylexuan44:
Icon regen looks good: 9 frames (16/20/24/32/40/48/64/128/256) vs 6, and 42 KB vs 370 KB. Sizes match GetSystemMetricsForDpi at 100/125/150/200%.

- All frames are PNG-compressed, and packaging/windows/mezon.iss uses this same file as SetupIconFile. ISCC only runs on v* tags, so PR CI never exercises it. Please do one workflow_dispatch release run before merging. If Inno rejects it, keep BMP for sizes <= 64 and PNG for 128/256 (still ~68 KB).
- No frames for 175% (28/56), 250% big (80) or 300% big (96). LoadIconWithScaleDown would downscale cleanly; LoadImage stretches.
- Test job failure is unrelated: crates/vendor/libspa/build.rs:11 uses deprecated cc::Build::shared_flag under -D warnings. Needs a separate fix.
- Windows Check passes, so the unsafe block and windows 0.61 signatures are compile-verified.

PR #211, opened 2026-08-03
fix: UI settings profile and clan

duocphungvanNCC:
> `Description:`

**env development:** window
- fix ui of setting profile and clan pages
- fix bug crash app when none clan found but enter clan profile setting
image

image

===================================================================================

image
 
image


crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -317,20 +392,40 @@ impl ProfilePage {
...
-            .child(h_flex().gap_8().child(form).child(preview))
-            .when(is_dirty || saving, |el| {
+            .child(
+                h_flex()
+                    .gap_8()
+                    .items_start()
+                    .child(div().min_w_0().flex_1().flex_basis(px(0.)).child(form))
+                    .child(div().min_w_0().flex_1().flex_basis(px(0.)).child(preview)),
+            )
+            .when(false && (is_dirty || saving), |el| {
```
huylexuan44: Dead branch: false && (...) makes this whole block unreachable, and clippy overly_complex_bool_expr denies it. Delete the block.

crates/mezon-ui/src/settings/clan_profile_section.rs
```diff
@@ -335,24 +428,48 @@ impl Render for ClanProfileSection {
...
-            .child(h_flex().gap_8().child(form).child(preview))
-            .when(is_dirty || saving, |el| {
+            .child(
+                h_flex()
+                    .gap_8()
+                    .items_start()
+                    .child(div().min_w_0().flex_1().flex_basis(px(0.)).child(form))
+                    .child(div().min_w_0().flex_1().flex_basis(px(0.)).child(preview)),
+            )
+            .when(false && (is_dirty || saving), |el| {
```
huylexuan44: Same dead branch here. false && (...) is unreachable and fails clippy overly_complex_bool_expr. Delete the block.

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -635,30 +805,179 @@ impl ProfilePage {
...
+                        .border_color(theme.border),
                     )
                     .child(
                         div()
-                            .text_xs()
+                            .w_full()
+                            .text_right()
+                            .text_sm()
                             .text_color(theme.text_muted)
                             .child(format!("{}/128", about_me.len())),
```
huylexuan44: len() is bytes but the limit is chars, so Vietnamese text shows ~2x the real count. Use about_me.chars().count().

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -222,12 +296,12 @@ impl ProfilePage {
...
             let about = about.clone();
-            move |this: &mut Self, _, event: &InputEvent, _, cx| {
-                if let InputEvent::Change = event
+            move |this: &mut Self, _, event: &TextAreaEvent, _, cx| {
+                if *event == TextAreaEvent::Change
                     && let Some(state) = &mut this.profile
                     && !state.saving
                 {
-                    let value = about.read(cx).value().to_string();
+                    let value = about.read(cx).value().chars().take(128).collect::<String>();
```
huylexuan44: Truncation is state-only; TextArea has no max_length so text past 128 chars stays visible but is dropped on save. Clamp the widget too.

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -303,6 +377,7 @@ impl ProfilePage {
     }
 
     fn render_user_section(&mut self, theme: &Theme, cx: &mut Context<Self>) -> impl IntoElement {
+        self.refresh_banner_color(cx);
```
huylexuan44: Spawns a task from the render path. account_page.rs and user_profile_modal.rs call this from new() plus the store observer instead.

crates/mezon-ui/src/settings/clan_profile_section.rs
```diff
@@ -278,10 +345,35 @@ impl ClanProfileSection {
...
+                this.banner_color = Some(color);
+                cx.notify();
+            },
+        );
+    }
 }
 
 impl Render for ClanProfileSection {
     fn render(&mut self, window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        self.refresh_banner_color(cx);
```
huylexuan44: Same here: banner refresh spawns a task from render. Drive it from new() and the AccountStore observer.

crates/mezon-ui/src/settings/screen.rs
```diff
@@ -101,6 +101,9 @@ impl SettingsScreen {
     }
 
     pub fn set_page(&mut self, page: SettingsPage, cx: &mut Context<Self>) {
+        if self.current_page == SettingsPage::Profile && page != SettingsPage::Profile {
+            self.profile_page = None;
```
huylexuan44: This makes ProfilePage transient unlike every sibling page, so its detached observers leak per visit and unsaved edits are dropped without a prompt.

crates/mezon-store/src/account.rs
```diff
@@ -396,30 +404,80 @@ impl AccountStore {
...
+
+    pub fn upload_direct_message_icon(&mut self, path: &Path, cx: &mut Context<Self>) {
+        let api = self.api.clone();
+        let path = path.to_path_buf();
+        cx.spawn(async move |this, cx| {
+            let result = cx
+                .background_executor()
+                .spawn(async move {
+                    let url = api.upload_avatar(&path).await?;
+                    api.update_account(None, None, None, Some(&url)).await?;
```
huylexuan44: This commits to the server immediately while every other field on the page is Save-gated, and the icon can never be cleared (empty logo is filtered to None).

crates/mezon-store/src/account.rs
```diff
@@ -396,30 +404,80 @@ impl AccountStore {
...
+                                Self::spawn_persist_cache(account, cx);
+                            }
+                            cx.emit(AccountEvent::UserAvatarUploaded(url));
+                        }
+                        cx.notify();
+                    });
+                }
+                Err(e) => {
+                    let _ = this.update(cx, |_, cx| {
+                        cx.emit(AccountEvent::AvatarUploadFailed(e.to_string()));
```
huylexuan44: Success was split into User/ClanAvatarUploaded but failure was not, so one failed upload fires both page and clan-section toasts.

huylexuan44:
Review summary (9 inline comments posted):

- Blocking: both per-section save bars were switched off with `false && (...)` instead of being deleted after the shared bar moved into ProfilePage::render. That is ~120 lines of unreachable code and it fails clippy overly_complex_bool_expr, which is deny by default.
- About Me counter uses bytes while the limit is chars, so Vietnamese text reads about double. The 128 cap is also state-only, so text past the limit stays visible and is silently dropped on save.
- refresh_banner_color spawns a task from render in both files. account_page.rs and user_profile_modal.rs already do this correctly from new() plus the store observer.
- Clearing profile_page on leave/ESC makes ProfilePage the only transient settings page: its detached observers leak per visit, and unsaved edits are discarded right after the new unsaved-changes bar warns about them.

The crash fix for entering clan profile with no clan looks solid, and all six new i18n keys are present in en.json.

Note on CI: the failures are pre-existing on develop (two clippy errors in mezon-store, plus a libspa build script failure), not from this PR. But because clippy aborts in mezon-store it never reaches mezon-ui, so none of the changes here were actually linted, which is how the `false &&` blocker got through.

PR #213, opened 2026-08-03
fix: realtime data of channel list

duocphungvanNCC:
> `Description:`

**env development:** window
- fix realtime data channel/thread of channel list


crates/mezon-store/src/channel_settings.rs
```diff
@@ -90,6 +96,18 @@ impl ChannelSettingsStore {
         if self.rows.contains_key(&key) || !self.loading.insert(key) {
             return;
         }
+        self.fetch(key, cx);
+    }
+
+    fn reload(&mut self, key: (ClanId, ChannelId), cx: &mut Context<Self>) {
+        if !self.loading.insert(key) {
```
huylexuan44, line 126: Reload during an in-flight fetch is dropped and never re-armed, so the change is lost. Track pending keys and refetch on completion.

crates/mezon-store/src/channel_settings.rs
```diff
@@ -131,4 +149,87 @@ impl ChannelSettingsStore {
...
+            dispatch.on(RealtimeKind::ChannelMessage, &entity, |this, event, cx| {
+                this.handle_message(event, cx)
+            });
+            for kind in [
+                RealtimeKind::ChannelCreated,
+                RealtimeKind::ChannelUpdated,
+                RealtimeKind::ChannelDeleted,
+                RealtimeKind::ChannelArchive,
+            ] {
+                dispatch.on(kind, &entity, |this, _, cx| this.reload_loaded(cx));
```
huylexuan44: Event payload is discarded, so a ChannelCreated in another clan refetches this clan keys. Scope the reload by clan_id.

crates/mezon-store/src/channel_settings.rs
```diff
@@ -131,4 +149,87 @@ impl ChannelSettingsStore {
...
+                *row_clan_id == clan_id && rows.iter().any(|row| row.id == channel_id)
+            })
+            .map(|(&key, _)| key)
+            .collect::<Vec<_>>();
+        for key in keys {
+            self.reload(key, cx);
+        }
+    }
+
+    fn reload_loaded(&mut self, cx: &mut Context<Self>) {
```
huylexuan44: rows is never cleared, so this refetches every loaded key forever, even long after the settings page closes. Add a reset.

crates/mezon-store/src/channel_settings.rs
```diff
@@ -131,4 +149,87 @@ impl ChannelSettingsStore {
...
+                row.message_count = row.message_count.saturating_add(1);
+                row.last_sender_id = UserId(message.sender_id);
+                row.last_sent_seconds = message.create_time_seconds;
+                changed.push(parent_id);
+            }
+        }
+        for parent_id in changed {
+            cx.emit(ChannelSettingsEvent::Changed { clan_id, parent_id });
+        }
+        cx.notify();
```
huylexuan44, line 433: notify fires on every ChannelMessage app-wide, even when no row changed. Return early when changed is empty.

crates/mezon-store/src/channel_settings.rs
```diff
@@ -131,4 +149,87 @@ impl ChannelSettingsStore {
...
+
+    fn handle_message(&mut self, event: &RealtimeEvent, cx: &mut Context<Self>) {
+        let RealtimeEvent::ChannelMessage(message) = event else {
+            return;
+        };
+        let clan_id = ClanId(message.clan_id);
+        let channel_id = ChannelId(message.channel_id);
+        let code = MessageCode::from_raw(message.code);
+
+        if matches!(
```
huylexuan44, line 220: Ephemeral and Unknown codes still increment message_count. Use an allow-list of codes the server actually counts.

crates/mezon-store/src/channel_settings.rs
```diff
@@ -131,4 +149,87 @@ impl ChannelSettingsStore {
...
+        let entity = cx.entity();
+        RealtimeDispatch::global(cx).update(cx, |dispatch, _| {
+            dispatch.on(RealtimeKind::ChannelMessage, &entity, |this, event, cx| {
+                this.handle_message(event, cx)
+            });
+            for kind in [
+                RealtimeKind::ChannelCreated,
+                RealtimeKind::ChannelUpdated,
+                RealtimeKind::ChannelDeleted,
+                RealtimeKind::ChannelArchive,
```
huylexuan44, line 202: Missing UserChannelAdded/UserChannelRemoved, so the Members column stays stale. threads.rs subscribes to both.

huylexuan44:
Review summary (6 inline comments posted):

- Blocking: a reload arriving during an in-flight fetch is silently dropped and never re-armed, which loses exactly the events this PR aims to handle.
- Blocking: reload_loaded refetches every loaded key, for events in any clan, forever, because rows is never cleared. React handles these with in-place reducers (addChannelToSocket / removeChannelFromSocket / updateChannelFromSocket in clanSettingChannel.slice.ts) and never refetches. Note the server serves List* from a cache keyed on (clan_id, uid), so a refetch right after a mutation may return pre-change data.
- Perf/correctness: unconditional notify on every ChannelMessage, and message_count increments for Ephemeral and Unknown codes.
- No tests added. CLAUDE.md requires tests for store transitions and realtime event handling; channel_settings.rs has no test module.

CI is red for pre-existing reasons, not this PR: the two clippy errors (inbox.rs:270, messages.rs:2859) reproduce on develop HEAD, and the Test job fails on a libspa build script. Worth rebasing once develop is green so CI can actually validate this change.

PR #218, opened 2026-08-03
fix: sort thread sidebar by asc alphabet

duocphungvanNCC:
> `Description:`

**env development:** window
- fix sort of thread in left sidebar by esc alphabet


huylexuan44:
selft test new code

PR #227, opened 2026-08-04
Fix search filter archived threads

giachuvan:


crates/mezon-store/src/channel.rs
```diff
@@ -3281,17 +3342,56 @@ fn apply_in_voice_leaved(
...
+    target.creator_id = source.creator_id;
+    target.active = source.active;
+    target.avatar_url = source.avatar_url.clone();
+    if source.last_sent_timestamp >= target.last_sent_timestamp {
+        copy_channel_unread_fields(target, source);
     }
-    false
+}
+
+fn add_user_channel(
```
huylexuan44: add_user_channel is now unused; CI runs -D warnings, so this dead_code warning fails the build. Delete it.

crates/mezon-store/src/threads.rs
```diff
@@ -1068,7 +1138,9 @@ fn filter_threads(threads: Vec<ThreadSummary>) -> Vec<ThreadSummary> {
         .into_iter()
         .filter(|t| {
             if t.channel_private != 0 {
-                t.active == THREAD_STATUS_JOINED || t.active == THREAD_STATUS_ACTIVE_PRIVATE
+                t.active == THREAD_STATUS_JOINED
+                    || t.active == THREAD_STATUS_ACTIVE_PRIVATE
+                    || t.active == THREAD_STATUS_ARCHIVED
```
huylexuan44, line 1100: React keeps only joined/activePrivate here (threads.slice.ts:480). active=0 on a private thread means not-a-member, so this leaks unjoined threads.

crates/mezon-store/src/threads.rs
```diff
@@ -1068,7 +1138,9 @@ fn filter_threads(threads: Vec<ThreadSummary>) -> Vec<ThreadSummary> {
         .into_iter()
         .filter(|t| {
```
huylexuan44, line 1096: React does not filter search results at all (threads.slice.ts:519). Fix the search bug by dropping filter_threads from schedule_search instead.

crates/mezon-store/src/channel.rs
```diff
@@ -1369,13 +1362,74 @@ impl ChannelList {
...
+            cx.notify();
+        }
+    }
+
+    fn merge_user_channels_from_api_descs(
+        &mut self,
+        descs: Vec<mezon_client::transport::ApiChannelDesc>,
+        cx: &mut Context<Self>,
+    ) {
+        for d in descs {
```
huylexuan44, line 1384: Lost the user_channels.clear() the old handler did, so the list only ever grows. Left clans and closed DMs now survive every resync.

crates/mezon-store/src/channel.rs
```diff
@@ -3281,17 +3342,56 @@ fn apply_in_voice_leaved(
...
+fn upsert_user_channel(
     user_channels: &mut HashMap<ChannelId, Channel>,
     order: &mut Vec<ChannelId>,
     channel: Channel,
 ) -> bool {
     let channel_id = channel.id;
-    if user_channels.insert(channel_id, channel).is_none() {
+    if let Some(existing) = user_channels.get_mut(&channel_id) {
+        merge_user_channel_from_structure(existing, &channel);
+        true
```
huylexuan44: Both branches return true, so the return value is meaningless and callers can no longer tell a real change from a no-op.

crates/mezon-store/src/channel.rs
```diff
@@ -1369,13 +1362,74 @@ impl ChannelList {
     }
 
     fn sync_user_channel_from(&mut self, source: &Channel, cx: &mut Context<Self>) {
```
huylexuan44, line 1365: Now notifies on every inbound message via note_channel_message since upsert always returns true; the old unread/badge gate is gone.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -200,6 +200,9 @@ pub fn build_palette_items(cx: &App) -> Vec<PaletteItem> {
     let channels = channel_list.read(cx);
     let clans = clan_list.read(cx);
     for channel in channels.user_channels() {
+        if channel.is_archived() {
```
huylexuan44: forward_modal.rs:303 and mention_input/mod.rs:2386 also read user_channels() and got no archived guard. Filter at the source instead.

crates/mezon-store/src/channel.rs
```diff
@@ -1369,13 +1362,74 @@ impl ChannelList {
...
+                &mut self.user_channels,
+                &mut self.user_channels_order,
+                channel,
+            );
+        }
+        self.sync_user_channels_from_all_loaded_caches(cx);
+    }
+
+    fn sync_user_channels_from_all_loaded_caches(&mut self, cx: &mut Context<Self>) {
+        let snapshots: Vec<Vec<Category>> =
```
huylexuan44: cats.clone() deep-copies every category of every cached clan on each response. Use a split borrow of self instead of snapshotting.

crates/mezon-store/src/threads.rs
```diff
@@ -571,8 +611,12 @@ impl ThreadsStore {
                 {
                     self.apply_channel(channel);
                 } else {
-                    self.list_channel_id = Some(id.to_string());
-                    self.clan_id = Some("0".to_string());
+                    self.list_channel_id = Some(list_channel_id_for_active(id, cx));
+                    self.clan_id = ClanList::global(cx)
```
huylexuan44: Was "0" for DMs. active_clan_id is not cleared when opening a DM, so DM thread fetches now send a stale clan id.

crates/mezon-store/src/threads.rs
```diff
@@ -1009,6 +1071,14 @@ fn list_channel_id_for(channel: &Channel) -> String {
     }
 }
 
+fn list_channel_id_for_active(channel_id: ChannelId, cx: &App) -> String {
```
huylexuan44: Both call sites sit in the else of find_channel_in_active_clan, the same lookup this retries, so it always hits the fallback.

crates/mezon-store/src/threads.rs
```diff
@@ -809,8 +861,18 @@ impl ThreadsStore {
                 }
                 match result {
                     Ok(list) => {
-                        let page_full = page_has_more(list.len());
+                        let raw_count = list.len();
+                        let page_full = page_has_more(raw_count);
                         let batch = filter_threads(list.into_iter().map(thread_from_api).collect());
+                        tracing::info!(
```
huylexuan44: Debug instrumentation on every popover open and load-more page. Downgrade to debug! or drop.

crates/mezon-store/src/channel.rs
```diff
@@ -4878,15 +5039,14 @@ mod tests {
             let channels = init_authenticated_channel_list(cx);
             channels.update(cx, |channels, cx| {
                 channels.apply_clan_structure(ClanId(1), structure_with_a_thread(), cx);
-                assert!(channels.user_channel(ChannelId(9)).is_none());
-
-                channels.handle_event(&added_to_channel(9, 1, &[REMOVED_SELF]), cx);
-
                 assert!(
                     channels.user_channel(ChannelId(9)).is_some(),
```
huylexuan44: This now holds both before and after handle_event, so the handler is no longer tested. Use a thread absent from the clan structure.

crates/mezon-store/src/threads.rs
```diff
@@ -1266,6 +1338,37 @@ mod tests {
...
+        assert_eq!(kept.len(), 2);
+        assert!(kept.iter().any(|t| t.channel_id == "1"));
+        assert!(kept.iter().any(|t| t.channel_id == "2"));
+    }
+
+    #[test]
+    fn filter_threads_drops_private_threads_user_has_not_joined() {
+        let threads = vec![{
+            let mut thread = summary("9", 1);
+            thread.active = THREAD_STATUS_ACTIVE_PUBLIC;
```
huylexuan44, line 1324: channel_private=1 with active=ACTIVE_PUBLIC is a contradictory state. The risky case, private with active=0, stays untested.

huylexuan44:
Reviewed at 04fea92. Tests pass locally (792 in mezon-store), but three blockers:

- CI break: add_user_channel is dead after the switch to upsert_user_channel, and ci.yml sets RUSTFLAGS -D warnings.
- Private-thread exposure: keeping active=0 for private threads contradicts React (threads.slice.ts:479-491), where active encodes membership, not just archival. The parity fix for the search bug is the opposite change - React does not filter search results at all, so drop filter_threads from schedule_search and leave the fetch filter alone.
- user_channels no longer prunes: the clear() on refetch is gone, so left clans and closed DMs survive every resync and stay in Ctrl+K, forward and hashtag autocomplete.

Also: upsert_user_channel always returns true, so ChannelList now notifies on every inbound message; the is_archived guard was added to only one of three user_channels() consumers.

The threads.rs scope resolution (sync_list_scope / resolve_fetch_clan_id) is a real improvement. The channel.rs user_channels rework is much wider than the PR title and may not be needed for this bug at all - worth splitting.

PR #242, opened 2026-08-05
Fix messages forwarded indicator

giachuvan:


crates/mezon-store/src/messages.rs
```diff
@@ -395,11 +395,16 @@ impl MessageList {
     }
 
     fn regroup_row(&mut self, idx: usize) {
-        let combined = {
+        let (combined, show_forwarded_label) = {
```
huylexuan44, line 398: regroup_row now duplicates the derived pair from recompute_message_grouping. Extract compute_grouping(prev, msg) -> (bool, bool) in message.rs so a third field cannot diverge again; that also drops the pub(crate) bump.

crates/mezon-store/src/messages.rs
```diff
@@ -8752,6 +8757,31 @@ mod tests {
...
+        let mut list =
+            MessageList::from_messages(vec![Message::new(MessageId(1), "a", "u1", "U1", 100)]);
+        list.push_grouped(
+            Message::new(MessageId(2), "fwd-a", "u1", "U1", 110).with_forwarded(true),
+        );
+        list.push_grouped(
+            Message::new(MessageId(3), "fwd-b", "u1", "U1", 120).with_forwarded(true),
+        );
+        assert!(list.as_slice()[1].show_forwarded_label);
+        assert!(!list.as_slice()[2].show_forwarded_label);
```
huylexuan44, line 8781: Add a case for the eviction path: regroup_row(0) after a drain should set show_forwarded_label on the new head.

crates/mezon-store/src/message.rs
```diff
@@ -1423,7 +1423,7 @@ pub fn recompute_message_grouping(messages: &mut [Message]) {
     }
 }
 
-fn compute_show_forwarded_label(prev: Option<&Message>, msg: &Message) -> bool {
+pub(crate) fn compute_show_forwarded_label(prev: Option<&Message>, msg: &Message) -> bool {
```
huylexuan44, line 1426: Pre-existing, no action: the 600_000 gap is seconds-based create_time (~6.9 days, not 10 min). React has the same bug, so keep it for parity.

huylexuan44:
Verified on 39b642b: cargo test -p mezon-store --lib passes 818/818, clippy --all-targets and fmt are clean.

- Root cause and fix look right. regroup_row was the only writer of show_forwarded_label that skipped the recompute; every other mutator already routes through recompute_message_grouping, and the eviction branch regroup_row(0) is covered too.
- Matches the React reference in MessageWithUser/index.tsx (shouldShowForwardedText), same four branches in the same order.
- Follow-up, not this PR: merge_sparse_sender does not carry is_forwarded the way it carries references, so a sparse echo omitting fwd would clear the label on your own forward. Same shape as the presign-echo/reply-reference bug.
- PR body is empty. One line saying the forwarded label was missing on realtime-received messages until reload would help.

PR #259, opened 2026-08-06
Fix pagination in search channel

giachuvan:


crates/mezon-ui/src/chat/message_search.rs
```diff
@@ -258,13 +258,41 @@ impl MessageSearchPanel {
...
+        let page_changed = self.last_results_page != state.current_page;
+
+        if content_changed {
+            self.rows_cache_fp = Some(fp);
+            self.cached_rows = Rc::new(build_search_list_rows(&state.results));
+            self.cached_highlights = Rc::new(search_content_highlight_terms(&state.query));
+        }
+
+        if content_changed || page_changed {
+            self.list_state.reset(self.cached_rows.len());
```
huylexuan44, line 272: Revision bumps per OGP preview, so this resets scroll to top mid-browse. Reset only on row count or page change.

crates/mezon-ui/src/chat/message_search.rs
```diff
@@ -1106,49 +1127,240 @@ fn render_rich_search_content(
...
+    let mut mentions = Vec::new();
+    let mut links = Vec::new();
+    for (range, style) in tokens {
+        if range.start >= range.end || range.end > text_len {
+            continue;
+        }
+        if !text.is_char_boundary(range.start) || !text.is_char_boundary(range.end) {
+            continue;
+        }
+        if style.background_color.is_some() {
```
huylexuan44: RichRunKind::Bold has no background or underline, so it falls through both arms and is dropped. Classify by run kind.

crates/mezon-ui/src/chat/message_search.rs
```diff
@@ -288,13 +316,19 @@ impl Render for MessageSearchPanel {
         let is_direct = self.is_direct;
         let state = MessageSearchStore::global(cx).read(cx).state(channel_id);
         let entity = cx.weak_entity();
-        if !state.is_searching {
-            self.set_results_page(state.current_page, cx);
-        }
+        self.set_results_page(state.current_page, cx);
```
huylexuan44: Calls cx.notify() inside render and pre-syncs last_results_page, making page_changed in sync_results_list dead. Drop it.

crates/mezon-store/src/message_search.rs
```diff
@@ -28,8 +28,10 @@ pub struct ChannelSearchState {
     pub results: Vec<SearchHit>,
     pub total: i32,
     pub current_page: i32,
+    pub loaded_page: i32,
     pub is_searching: bool,
     pub has_error: bool,
+    pub page_empty: bool,
```
huylexuan44: page_empty is written but never read - the UI derives show_empty_page itself. Read it here or drop the field.

PR #279, opened 2026-08-10
Fix: sort name in channel list page the Vietnamese alphabet with key d/đ

duocphungvanNCC:


crates/mezon-ui/src/chat/clan_channels_page.rs
```diff
@@ -891,6 +891,10 @@ fn normalize(value: &str) -> String {
         .nfkd()
         .filter(|ch| !is_combining_mark(*ch))
         .flat_map(char::to_lowercase)
+        .map(|character| match character {
```
huylexuan44: Third copy of this fold. Put it in util/text_utils.rs so the Ctrl+K palette stops missing Đ names too.

crates/mezon-ui/src/chat/clan_channels_page.rs
```diff
@@ -891,6 +891,10 @@ fn normalize(value: &str) -> String {
...
+            '\u{0111}' => 'd',
```
huylexuan44: Also fold the identical ETH pair U+00D0/U+00F0; a name typed with Ð still sorts after z and misses a da query.

crates/mezon-ui/src/chat/clan_channels_page.rs
```diff
@@ -891,6 +891,10 @@ fn normalize(value: &str) -> String {
...
+            _ => character,
```
huylexuan44: ł ø ß æ œ þ have no decomposition either, so they still sort after z. A small fold table covers the class.

crates/mezon-ui/src/chat/clan_channels_page.rs
```diff
@@ -891,6 +891,10 @@ fn normalize(value: &str) -> String {
         .nfkd()
```
huylexuan44: nfkd here vs nfd in clan_members_page::normalize_search, so the two settings tables normalize differently.

crates/mezon-ui/src/chat/clan_channels_page.rs
```diff
@@ -891,6 +891,10 @@ fn normalize(value: &str) -> String {
...
         .filter(|ch| !is_combining_mark(*ch))
         .flat_map(char::to_lowercase)
+        .map(|character| match character {
+            '\u{0111}' => 'd',
+            _ => character,
+        })
```
huylexuan44: .map(|ch| if ch == 'đ' { 'd' } else { ch }) is one line, and matches the ch naming on the line above.

huylexuan44:
The fold itself is correct. Five inline comments, plus two notes that have no line in the diff:

- clan_channels_page.rs:1045, search_normalization_removes_diacritics only asserts normalize("Hiền") == "hien", which passes with or without these four lines. Delete them and just test stays green. Add assert_eq!(normalize("Đà Nẵng"), "da nang") here and in the twin test at clan_members_page.rs:1324.
- React normalizeString (libs/utils/src/lib/utils/index.ts:372) is NFD + strip U+0300-U+036F with no đ handling, and SettingChannel does no client-side sort, so desktop now answers a da query differently from web. Fine as a deliberate improvement, but upstream it or note it as a desktop-only divergence.

Scope note: this fixes the clan-settings channel table only. The same sort-after-z bug is still live in the Ctrl+K palette (util/text_utils.rs:33 has no đ fold), friends_page.rs:371 (bare to_lowercase key), the notification-setting modal, and message_search.rs. Worth separate issues rather than growing this PR.

PR #281, opened 2026-08-10
Fix stream connected overlap channels list

giachuvan:


crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1758,7 +1747,6 @@ impl Render for ChatLayout {
                     .flex_col()
                     .w(px(344.0))
                     .h_full()
```
huylexuan44, line 1749: Left column paints no bg, so the footer card gutters now fall through to RootView bg_primary. Add .bg(theme.bg_tertiary) here.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1768,25 +1756,18 @@ impl Render for ChatLayout {
...
-                            ),
+                            .child(div().w(px(272.0)).h_full().child(nav_body)),
                     )
                     .child(
                         div()
-                            .absolute()
-                            .left(px(12.0))
-                            .right(px(8.0))
-                            .bottom(px(12.0))
+                            .flex_none()
```
huylexuan44, line 1769: flex_none with no max_h: a tall footer stack shrinks the channel list and can push the user row off the window bottom.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1768,25 +1756,18 @@ impl Render for ChatLayout {
...
+                            .mx_2()
```
huylexuan44: React FooterProfile is left-[12px] right-2; mx_2 shifts the card 4px left and 8px wider. Use .ml(px(12.)).mr_2().

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1797,24 +1778,7 @@ impl Render for ChatLayout {
                             .bg(theme.tokens.bg_surface)
                             .occlude()
```
huylexuan44: occlude() existed for the floating overlay; in flow it only blocks group_hover inside its own subtree. Drop it.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -897,7 +897,6 @@ impl Render for ChannelSidebar {
             .flex_col()
             .w_full()
             .h_full()
```
huylexuan44, line 899: Dropping pb(68) makes all three sidebars depend on being ChatLayout footer sibling. Note that coupling somewhere.

huylexuan44:
Approach looks right: it removes the per-frame canvas measure loop and also reserves space for the update pills, which nav_bottom_pad never did. Five inline comments above. Two more that fall outside the diff:

- crates/mezon-ui/src/chat/stream.rs: this PR deletes the only call site of stream_connected_bar_height(), so that fn and STREAM_CONNECTED_BAR_HEIGHT are now orphaned. Remove both, otherwise the hardcoded 52px reserve gets reintroduced later.
- crates/mezon-ui/src/components/compositions/user_info_bar.rs:194: the comment still says the chat layout applies absolute positioning and insets. That is no longer true. Update it.

PR #282, opened 2026-08-10
Fix typing indicator issues

giachuvan:


crates/mezon-ui/src/chat/channel_typing.rs
```diff
@@ -78,21 +78,21 @@ impl ChannelTyping {
 
 impl Render for ChannelTyping {
     fn render(&mut self, _window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
-        let (primary, active) = {
-            let theme = cx.theme();
-            (theme.tokens.text_theme_primary, theme.tokens.text_secondary)
-        };
+        let theme = cx.theme();
+        let primary = theme.tokens.text_theme_primary;
+        let name_color = theme.tokens.text_theme_primary_hover;
```
huylexuan44: text_theme_primary_hover is a=0.122 in sunrise/cisher and red in purple_haze. React uses text-secondary; revert.

crates/mezon-ui/src/chat/channel_typing.rs
```diff
@@ -78,21 +78,21 @@ impl ChannelTyping {
...
         let bar = div()
             .pl_3()
-            .h(px(16.))
+            .pr_1()
+            .h_full()
```
huylexuan44: h_full breaks the non-cached mount in create_topic_panel.rs:245 - 0px when idle, overflows when typing. Keep h(px(16.)).

crates/mezon-ui/src/app/window_controls.rs
```diff
@@ -147,20 +147,24 @@ pub fn is_edge_resizable() -> bool {
 }
 
 pub fn render_resize_edges(window: &mut Window) -> impl IntoElement {
-    if !is_edge_resizable() || window.is_maximized() {
+    if !is_edge_resizable() {
         return div().id("window-resize-edges-disabled");
     }
 
     let decorations = window.window_decorations();
+    if matches!(decorations, Decorations::Client { .. }) {
```
huylexuan44: Window::client_inset now stores 8px while X11/Wayland apply 0 when maximized, adding 8px to snap-to-window popovers.

crates/mezon-ui/src/app/window_controls.rs
```diff
@@ -147,20 +147,24 @@ pub fn is_edge_resizable() -> bool {
...
+        window.set_client_inset(px(RESIZE_BORDER_SIZE));
+    }
+
+    if window.is_maximized() {
```
huylexuan44: is_maximized() is false in fullscreen, so resize strips stay live over the outer 8px. Add || window.is_fullscreen().

crates/mezon-ui/src/app/window_controls.rs
```diff
@@ -147,20 +147,24 @@ pub fn is_edge_resizable() -> bool {
...
 
 pub fn render_resize_edges(window: &mut Window) -> impl IntoElement {
-    if !is_edge_resizable() || window.is_maximized() {
+    if !is_edge_resizable() {
         return div().id("window-resize-edges-disabled");
     }
 
     let decorations = window.window_decorations();
+    if matches!(decorations, Decorations::Client { .. }) {
+        window.set_client_inset(px(RESIZE_BORDER_SIZE));
```
huylexuan44: set_client_inset is a platform write now run every frame from render(). Drive it from the decoration callback.

crates/mezon-ui/src/chat/channel_typing.rs
```diff
@@ -78,21 +78,21 @@ impl ChannelTyping {
...
+            .w_full()
```
huylexuan44: w_full/h_full is inert here: the .cached() wrapper at area.rs:667 already pins w_full + h(px(16.)). Change it there.

crates/mezon-ui/src/chat/channel_typing.rs
```diff
@@ -110,16 +110,15 @@ impl Render for ChannelTyping {
                         .flex()
                         .flex_row()
                         .items_center()
-                        .flex_1()
                         .min_w_0()
+                        .overflow_hidden()
```
huylexuan44: overflow_hidden is redundant with min_w_0 + truncate, and it clips the is-typing suffix earlier. Drop it.

crates/mezon-ui/src/app/window_controls.rs
```diff
@@ -147,20 +147,24 @@ pub fn is_edge_resizable() -> bool {
 }
 
 pub fn render_resize_edges(window: &mut Window) -> impl IntoElement {
-    if !is_edge_resizable() || window.is_maximized() {
+    if !is_edge_resizable() {
         return div().id("window-resize-edges-disabled");
```
huylexuan44, line 151: Disabled return is now duplicated at 151 and 160. One let-else on Decorations::Client covers both paths.

huylexuan44:
Eight inline comments above. One more that falls just outside the diff hunk so it cannot be anchored:

- crates/mezon-ui/src/chat/channel_typing.rs:99, self.content(cx) inside render: PresenceStore::typing_users allocates a Vec and clones each name into a new SharedString, and mezon_i18n::t allocates a String, on every frame. This view is notified on every TypingChanged event for the active channel. Cache the locale-derived suffix and the several-people string on the entity when settings.language changes, and read the name without building a Vec.

The two I would block on are the name color at channel_typing.rs:83 and h_full at :87.

PR #286, opened 2026-08-10
Feat archive thread and channel

giachuvan:


crates/mezon-store/src/channel.rs
```diff
@@ -2729,6 +2891,136 @@ impl ChannelList {
...
+            return false;
+        }
+        if self
+            .user_channels
+            .get(&channel_id)
+            .is_some_and(|ch| ch.active == CHANNEL_ACTIVE_ARCHIVED)
+        {
+            return false;
+        }
+        self.cache.iter().any(|(_, cats)| {
```
huylexuan44: Returns false for every DM since DM ids are not in cache, so bind_channel calls clear_draft and the draft is lost.

crates/mezon-store/src/channel.rs
```diff
@@ -1048,8 +1196,15 @@ impl ChannelList {
         desc: ApiChannelDesc,
         cx: &mut Context<Self>,
     ) {
+        let channel_id = ChannelId(desc.channel_id);
+        if self.archived_channel_ids.contains(&channel_id) {
+            return;
+        }
         let badge = desc.badge_count.max(0) as u32;
         let mut channel = channel_from_desc(desc, badge, Vec::new(), false);
+        if !channel.visible_in_sidebar() {
```
huylexuan44, line 1205: Returning without setting channel_detail_failed makes ensure_channel_in_clan refetch on every notify, forever.

crates/mezon-store/src/channel.rs
```diff
@@ -2796,38 +3088,21 @@ impl ChannelList {
...
+            if was_viewing && archived_by_other {
+                cx.emit(ChannelEvent::ArchivedByAdministrator { is_thread });
             }
-            crate::threads::ThreadsStore::global(cx).update(cx, |store, cx| {
-                store.mark_thread_archived(&channel_id.to_string(), cx);
-            });
             return;
         }
 
+        self.archived_channel_ids.remove(&channel_id);
```
huylexuan44, line 3116: Also remove the child threads tombstoned by remove_child_threads_of_channel; the server never archived them.

crates/mezon-store/src/channel.rs
```diff
@@ -166,6 +173,68 @@ impl Channel {
...
 }
 
+pub fn archive_menu_hidden(channel_type: ChannelType, is_welcome_channel: bool) -> bool {
+    matches!(
+        channel_type,
+        ChannelType::Voice | ChannelType::Stream | ChannelType::App
+    ) || is_welcome_channel
+}
+
+pub fn archive_allowed_by_server(
```
huylexuan44, line 183: Diverges from React both ways: manageChannel dropped for channels, manageThread && isCreator dropped for threads.

crates/mezon-store/src/channel.rs
```diff
@@ -842,6 +927,58 @@ impl ChannelList {
...
+                    .and_then(|ch| ch.parent_id)
+                    .unwrap_or(ChannelId(0));
+                Some(Ok((parent_id, this.api.clone())))
+            });
+            let Some(prep) = prep.ok().flatten() else {
+                return Err("archive channel store unavailable".into());
+            };
+            let (parent_id, api) = match prep {
+                Ok(value) => value,
+                Err(err) => {
```
huylexuan44: ARCHIVE_ERR_IN_PROGRESS also runs finish_archiving, releasing the guard of the archive still in flight.

crates/mezon-store/src/channel.rs
```diff
@@ -2729,6 +2891,136 @@ impl ChannelList {
...
+                self.invalidate_channel_index(clan_id);
+            }
+        }
+        if self.active_channel_id == Some(channel_id) {
+            let redirect = (!parent_id.is_zero()).then_some(parent_id);
+            self.active_channel_id = redirect;
+            cx.emit(ChannelEvent::ActiveChannelChanged(redirect));
+        }
+        cx.notify();
+        if let Some(user_channel) = self.user_channels.get_mut(&channel_id) {
```
huylexuan44: active=0 keeps the row, so forward_modal and hashtag suggestions still list it. Use remove_user_channel.

crates/mezon-store/src/channel.rs
```diff
@@ -2729,6 +2891,136 @@ impl ChannelList {
...
+    }
+
+    pub fn apply_local_archive(
+        &mut self,
+        clan_id: ClanId,
+        channel_id: ChannelId,
+        parent_id: ChannelId,
+        cx: &mut Context<Self>,
+    ) {
+        self.archived_channel_ids.insert(channel_id);
```
huylexuan44, line 2931: No sync_clan_after_read here, so the clan badge keeps counting an archived channel. See apply_self_removed.

crates/mezon-store/src/channel.rs
```diff
@@ -2729,6 +2891,136 @@ impl ChannelList {
...
+            cx.emit(ChannelEvent::ActiveChannelChanged(redirect));
+        }
+        cx.notify();
+        if let Some(user_channel) = self.user_channels.get_mut(&channel_id) {
+            user_channel.active = CHANNEL_ACTIVE_ARCHIVED;
+        }
+        crate::threads::ThreadsStore::global(cx).update(cx, |store, cx| {
+            store.mark_thread_archived(&channel_id.to_string(), cx);
+        });
+        self.clear_compose_draft(channel_id, cx);
```
huylexuan44: Drafts are keyed by conversation_channel_id, which is the PARENT for threads, so this clears nothing.

crates/mezon-store/src/channel.rs
```diff
@@ -2729,6 +2891,136 @@ impl ChannelList {
...
+            .get(&channel_id)
+            .is_some_and(|ch| ch.active == CHANNEL_ACTIVE_ARCHIVED)
+        {
+            return false;
+        }
+        self.cache.iter().any(|(_, cats)| {
+            cats.iter()
+                .flat_map(|c| &c.channels)
+                .any(|ch| ch.id == channel_id)
+        })
```
huylexuan44: Linear scan of every clan/category/channel on each switch; clan_id_for_channel(id).is_some() is indexed.

crates/mezon-store/src/channel.rs
```diff
@@ -166,6 +173,68 @@ impl Channel {
...
+        return is_creator;
+    };
+    let permissions = permissions.read(cx);
+    archive_allowed_by_server(
+        is_thread,
+        is_creator,
+        permissions.check(clan_id, None, PERMISSION_CLAN_OWNER, cx),
+        permissions.check(clan_id, None, PERMISSION_ADMINISTRATOR, cx),
+        permissions.check(clan_id, None, PERMISSION_MANAGE_CLAN, cx),
+        permissions.check(clan_id, Some(channel_id), PERMISSION_MANAGE_CHANNEL, cx),
```
huylexuan44: manage-channel is not in OVERRIDDEN_SLUGS, so Some(channel_id) is ignored and this reads the clan-wide level.

crates/mezon-store/src/channel.rs
```diff
@@ -319,6 +389,8 @@ pub struct ChannelList {
     channel_index: RefCell<ChannelLocationCache>,
     reset_generation: u64,
     reactivating: HashSet<ChannelId>,
+    archiving: HashSet<ChannelId>,
+    archived_channel_ids: HashSet<ChannelId>,
```
huylexuan44, line 393: Six read-side guards and one prune site. A single is_hidden() gate at visible_in_sidebar() would centralize it.

crates/mezon-store/src/channel.rs
```diff
@@ -985,12 +1126,19 @@ impl ChannelList {
     /// channel/thread absent from the clan structure, fetch its detail and insert it. Returns
```
huylexuan44: This doc comment belongs to ensure_channel_in_clan; the new fn was inserted between them.

crates/mezon-store/src/channel.rs
```diff
@@ -842,6 +927,58 @@ impl ChannelList {
...
     }
 
+    pub fn archive_channel(
+        &self,
+        clan_id: ClanId,
+        channel_id: ChannelId,
+        cx: &mut Context<Self>,
+    ) -> Task<Result<(), String>> {
+        cx.spawn(async move |this, cx| {
+            let prep = this.update(cx, |this, cx| {
```
huylexuan44: This Option is never None. Result<(ChannelId, Arc<AppApi>), String> plus let Ok(..) else drops a state.

crates/mezon-client/src/transport.rs
```diff
@@ -6559,6 +6559,21 @@ impl MezonTransport {
         Ok(())
     }
 
+    /// Archive channel or thread (sets active = 0 on server).
```
huylexuan44: Repo code style forbids doc comments, and active=0 is already named by CHANNEL_ACTIVE_ARCHIVED.

huylexuan44:
Fourteen inline comments. One more has no line in the diff because the function is untouched but the new tombstone set breaks it:

- channel.rs:2862, apply_thread_reactivated re-inserts the thread into the cache but never removes it from archived_channel_ids. Reactivate an archived thread by sending a message and the next apply_clan_structure deletes it again at the retain on line 1015, so it disappears from the sidebar with no way back this session.

Two I would block on before the archive behaviour itself, because they hit users who never archive anything:

- channel.rs:2917, every DM switch now deletes the unsent draft.
- channel.rs:1205, a route pointing at a server-archived thread refetches list_channel_detail once per notify and never falls back to the parent.

Note also that archive_channel_returns_error_when_already_in_flight passes despite the guard bug at :955 - it asserts the error value but never re-checks is_archiving. Findings at :3105, :2862 and :393 are the same missing invariant, so centralizing the prune would close all three.

PR #296, opened 2026-08-11
Fix scrollbar in topic

giachuvan:


crates/mezon-ui/src/chat/layout.rs
```diff
@@ -2159,15 +2159,10 @@ impl ChatLayout {
...
-            let align_timeline = self.chat_area.timeline.clone();
-            self.topic_panel = Some(cx.new(|cx| {
-                crate::chat::create_topic_panel::TopicPanel::new(
-                    settings,
-                    align_timeline,
-                    window,
-                    cx,
-                )
-            }));
+            self.topic_panel =
```
huylexuan44, line 2162: Nit: the arg list shrank, so this assignment fits on a single line now.

huylexuan44:
LGTM. Notes:

- Removal is complete: no references to topic_spacer / topic_align / message_viewport_top remain, and sync_topic_rows, remeasure_topic_rows and render_topic_box all drop the spacer base together. rustfmt is clean on all three files.
- Good fix beyond the scrollbar itself: sync_topic_spacer spliced the list and scheduled a notify from inside render_topic_box, which is a render-to-notify bounce.
- Matches React: TopicDiscussionBox renders ChannelMessages with isTopicBox and no alignment spacer, so this restores parity. The origin message still shows as row 0 via collect_topic_messages.
- CI Lint is red from pre-existing errors on develop (mezon-store/src/channel.rs:7675 and :7795, confirm_archive_channel_modal.rs:90, account_page.rs:71), not from this PR. Please add a PR description noting the origin-alignment behavior is intentionally dropped.

giachuvan:
## Notes
- Intentionally drops desktop-only origin-message vertical alignment between the channel timeline and the topic panel (top spacer synced to channel scroll). React never had this; keeping it caused a false topic scrollbar and empty top space when scrolling the main channel.
- rustfmt keep current style — nit comment conflict with formatter.

crates/mezon-ui/src/chat/layout.rs
giachuvan: rustfmt keep current style — nit comment conflict with formatter.

PR #299, opened 2026-08-11
Show share screen status in voice call

giachuvan:


crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -1910,25 +1913,42 @@ fn render_sidebar_item(
...
-                                        .flex_1()
-                                        .truncate()
-                                        .text_sm()
-                                        .text_color(theme.text_muted)
-                                        .child(name_text),
-                                )
+                                        .flex()
+                                        .flex_row()
+                                        .items_center()
+                                        .gap_1()
```
huylexuan44: React uses gap-2 (8px) for this avatar and name group; gap_1 is 4px.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -1910,25 +1913,42 @@ fn render_sidebar_item(
...
+                                            div()
+                                                .min_w_0()
+                                                .flex_1()
+                                                .truncate()
+                                                .text_sm()
+                                                .text_color(theme.text_muted)
+                                                .child(name_text),
+                                        ),
+                                );
+                            if sharing_screen {
```
huylexuan44: Prefer .when(sharing_screen, |el| el.child(...)) to match the .when style used a few lines above.

huylexuan44:
LGTM. Notes:

- Verified against the proto: ScreenShareEvent fields and share_screen_ids (repeated string) map correctly, and the parse matches the existing user_ids handling. Notify wiring is complete since VoiceMemberSlot PartialEq now includes sharing_screen. Leave and rejoin clear the flag correctly.
- Matches React: UserListItemVoiceChannel.tsx renders VoiceScreenShareIcon with color #22c55e at 16px on a justify-between row, so the hardcoded hex is correct here.
- Follow-up: React drives three more sites from the same status (MemberList/listMember.tsx, ChannelTopbar, DMListItem) which switch In voice to Share Screen. Those read self.in_voice in the Rust port, and there is no shareScreen i18n key yet.
- CI Lint is red from pre-existing errors on develop (channel.rs:7778 and :7898, confirm_archive_channel_modal.rs:90, account_page.rs:71), not from this PR. Please add a PR description and a screenshot of the sidebar row.

PR #300, opened 2026-08-11
Fix unread badge with user own message

giachuvan:


crates/mezon-store/src/badge.rs
```diff
@@ -395,9 +395,12 @@ impl BadgeService {
                                 ts,
                                 message_id,
                                 cx,
-                            )
+                            );
+                            if from_me && is_new_message {
+                                cl.apply_read(clan_id, channel_id, cx);
```
huylexuan44: Regression: clears badge_count before note_viewport_seen reads it, so last_seen writes 0 and the server never decrements. Capture the badge first.

huylexuan44: sync_clan_after_read has no is_clan_cache_loaded guard, so this clears the unread dot of a clan whose channels never loaded.

huylexuan44: apply_read pushes last_seen_timestamp up to last_sent_timestamp, which can exceed your own message and drop a late mention.

huylexuan44: apply_read also drops pending_channel_badges, the seeded badge and topic badges - too destructive for an own-send echo.

huylexuan44: Perf: full clan scan plus clan_total_badge re-sum and notify on every send. Gate on channel_badge_count > 0 first.

crates/mezon-store/src/badge.rs
```diff
@@ -395,9 +395,12 @@ impl BadgeService {
                                 ts,
                                 message_id,
                                 cx,
-                            )
+                            );
+                            if from_me && is_new_message {
```
huylexuan44: This sits under the topic_id == 0 guard, so replying in a topic never clears the parent badge. React has no topic guard.

crates/mezon-store/src/badge.rs
```diff
@@ -395,9 +395,12 @@ impl BadgeService {
...
+                                cl.apply_read(clan_id, channel_id, cx);
+                            }
                         });
-                        if seen {
+                        if seen || from_me {
```
huylexuan44: is_clan_message_seen already returns true when from_me, so || from_me is dead. Drop it.

crates/mezon-store/src/badge.rs
```diff
@@ -395,9 +395,12 @@ impl BadgeService {
...
                             MessagesStore::global(cx).update(cx, |store, _| {
                                 store.set_last_read_message(channel_id, message_id);
```
huylexuan44, line 436: Missing the !message_id.is_zero() guard used by the LastSeenUpdated arm; a zero id erases the last-read anchor.

crates/mezon-store/src/channel.rs
```diff
@@ -6417,6 +6417,52 @@ mod tests {
...
+                channels.note_channel_message(
+                    ClanId(1),
+                    ChannelId(1),
+                    false,
+                    true,
+                    200,
+                    MessageId(10),
+                    cx,
+                );
+                channels.apply_read(ClanId(1), ChannelId(1), cx);
```
huylexuan44: This test calls apply_read directly, so it passes on the base commit and covers none of the badge.rs change.

PR #317, opened 2026-08-12
Format popup position issue

giachuvan:


crates/mezon-canvas/src/editor.rs
```diff
@@ -2972,6 +2980,158 @@ fn open_canvas_link(url: &str, cx: &mut App) {
...
+        let Some(dropdown) = self.dropdown.as_mut() else {
+            return;
+        };
+
+        let menu_bounds = window.layout_bounds(dropdown_id);
+        let menu_size = menu_bounds.size;
+        let client_inset = window.client_inset().unwrap_or(px(0.));
+        let viewport_bottom = window.viewport_size().height - client_inset;
+        let space_below = viewport_bottom - bounds.bottom();
+        let flip_up = space_below < menu_size.height + CANVAS_BLOCK_DROPDOWN_GAP;
```
huylexuan44: flip_up ignores space above: at the 600px min window height a selection at y~300 pushes the ~315px menu off the top. Clamp desired into the viewport like anchored does.

crates/mezon-canvas/src/editor.rs
```diff
@@ -3051,7 +3211,28 @@ fn render_bubble_menu(
                 ),
         );
 
-    h_flex()
+    let block_dropdown_panel = block_menu_open.then(|| {
```
huylexuan44, line 3225: Link popover at line 3349 still uses hardcoded .absolute().top(40).left(-115) - route it through CanvasFormatToolbar too.

crates/mezon-canvas/src/editor.rs
```diff
@@ -2972,6 +2980,158 @@ fn open_canvas_link(url: &str, cx: &mut App) {
...
+            point(bounds.left(), bounds.top() - CANVAS_BLOCK_DROPDOWN_GAP)
+        } else {
+            point(bounds.left(), bounds.bottom() + CANVAS_BLOCK_DROPDOWN_GAP)
+        };
+        let anchor = if flip_up {
+            Anchor::BottomLeft
+        } else {
+            Anchor::TopLeft
+        };
+        let desired = Bounds::from_anchor_and_size(anchor, attach, menu_size);
```
huylexuan44: Vertical axis only. Clamp desired.origin.x to the viewport so the 220px menu cannot overflow the right edge.

crates/mezon-canvas/src/editor.rs
```diff
@@ -2972,6 +2980,158 @@ fn open_canvas_link(url: &str, cx: &mut App) {
     }
 }
 
+const CANVAS_BLOCK_DROPDOWN_GAP: Pixels = px(0.);
```
huylexuan44: GAP is px(0.), so the menu now sits flush against the bar (doubled border, stacked shadows). Give it 4-6px or drop the constant.

crates/mezon-canvas/src/editor.rs
```diff
@@ -2972,6 +2980,158 @@ fn open_canvas_link(url: &str, cx: &mut App) {
...
+
+struct CanvasFormatToolbar {
```
huylexuan44, line 2985: Add a comment here: this is a custom Element to avoid nesting deferred() inside the outer deferred (depth-panic bug).

crates/mezon-canvas/src/editor.rs
```diff
@@ -2972,6 +2980,158 @@ fn open_canvas_link(url: &str, cx: &mut App) {
...
+
+        let mut child_ids = smallvec::SmallVec::<[LayoutId; 2]>::new();
+        child_ids.push(bar_id);
+        if let Some(dropdown_id) = dropdown_id {
+            child_ids.push(dropdown_id);
+        }
+
+        let layout_id = window.request_layout(
+            Style {
+                position: Position::Relative,
```
huylexuan44: Style::default() is already Position::Relative + Display::Block - drop both fields.

crates/mezon-canvas/src/editor.rs
giachuvan: project rules not allow to add comment to

PR #326, opened 2026-08-13
Fix list files after update doc type

giachuvan:


crates/mezon-store/src/files.rs
```diff
@@ -6,14 +6,15 @@ use mezon_client::AppApi;
...
+    LoadDirection, initial_page_has_more, next_page_has_more, resolve_attachment_uploader,
+};
 use crate::ids::{ChannelId, ClanId, MessageId, UserId};
 
 pub const FILES_CACHE_TTL: Duration = Duration::from_secs(60 * 60);
 pub const FILES_PAGE_SIZE: i32 = 100;
-pub const FILES_TYPED_QUERY: &str = "FILE";
-pub const FILES_BROAD_QUERY: &str = "";
-const FILES_QUERY_GEN: u8 = 3;
+pub const FILES_QUERY: &str = "doc";
```
huylexuan44, line 16: Backend only special-cases "FILE"; any other value falls through to no type filter. Confirm the API knows "doc".

crates/mezon-store/src/files.rs
```diff
@@ -219,35 +256,73 @@ impl FilesStore {
...
-                        entry.documents = docs;
+                        entry.fetch_error = false;
+                        let added =
+                            merge_documents(&mut entry.documents, &mut entry.ids, docs, reset);
+                        if reset {
+                            entry.has_more_before =
+                                initial_page_has_more(raw_count, FILES_PAGE_SIZE);
+                        } else {
+                            entry.has_more_before =
+                                next_page_has_more(raw_count, FILES_PAGE_SIZE, added);
```
huylexuan44, line 322: raw_count is pre-filter but added is post-filter, so a page with no documents sets has_more_before=false forever.

crates/mezon-ui/src/chat/files_popover.rs
```diff
@@ -156,10 +160,40 @@ impl FilesPopoverPanel {
...
         this.refresh_rows(cx);
         this
     }
 
+    fn install_scroll_handler(&self, cx: &mut Context<Self>) {
+        let weak = cx.weak_entity();
+        let clan = self.clan_id;
+        let channel = self.channel_id;
+        self.list_state
+            .set_scroll_handler(move |event, _window, cx| {
```
huylexuan44, line 173: render_body drops the list when rows are empty, so this handler never runs and the empty state cannot recover.

crates/mezon-store/src/files.rs
```diff
@@ -219,35 +256,73 @@ impl FilesStore {
...
+            if let Some(oldest) = channel.and_then(|c| c.documents.last()) {
+                before = oldest.create_time_seconds;
+            }
+        }
+
         let entry = self.ensure_channel(channel_id);
         entry.is_loading = true;
         entry.fetch_error = false;
+        if reset {
+            entry.documents.clear();
```
huylexuan44, line 279: Clearing before the request blanks the list on every open and loses it on failure; merge_documents already resets.

crates/mezon-store/src/files.rs
```diff
@@ -70,7 +71,13 @@ pub fn is_document(filetype: &str) -> bool {
     if lower == "sticker" {
         return false;
     }
-    if lower.starts_with("image/") || lower.starts_with("video/") {
+    if lower.starts_with("image/")
+        || lower.starts_with("video/")
+        || lower.starts_with("audio/")
```
huylexuan44, line 76: React counts audio as a document. Excluding audio/* hides voice files and makes the popover audio player dead code.

crates/mezon-ui/src/chat/files_popover.rs
```diff
@@ -156,10 +160,40 @@ impl FilesPopoverPanel {
...
+                        if this.viewport_autofill_remaining == 0 {
+                            return;
+                        }
+                        this.viewport_autofill_remaining -= 1;
+                    }
+                    FilesStore::global(cx).update(cx, |store, cx| {
+                        store.fetch_page(clan, channel, LoadDirection::Before, cx);
+                    });
+                });
+            });
```
huylexuan44, line 194: Scroll handler fires on wheel only; scrollbar drag bypasses it. Run the pagination check from render instead.

crates/mezon-store/src/files.rs
```diff
@@ -219,35 +256,73 @@ impl FilesStore {
...
+                    channel_id.0,
+                    FILES_QUERY,
+                    0,
+                    FILES_PAGE_SIZE,
+                    before,
+                    0,
+                )
+                .await;
             let mapped = result.map(|list| {
+                let raw_count = list.len();
```
huylexuan44, line 300: transport already drops empty-url attachments, so raw_count under-reports the page and can cap the list at 99.

crates/mezon-store/src/files.rs
```diff
@@ -219,35 +256,73 @@ impl FilesStore {
...
+        channel_id: ChannelId,
+        _direction: LoadDirection,
+        reset: bool,
+        cx: &mut Context<Self>,
+    ) {
+        let mut before = 0u32;
+        if !reset {
+            let channel = self.by_channel.get(&channel_id);
+            if let Some(oldest) = channel.and_then(|c| c.documents.last()) {
+                before = oldest.create_time_seconds;
```
huylexuan44, line 271: Backend filters create_time_second < ?, so items tied on this second are skipped; a 0 timestamp drops the cursor.

crates/mezon-ui/src/chat/files_popover.rs
```diff
@@ -173,14 +207,17 @@ impl FilesPopoverPanel {
...
-        ));
-        if self.list_state.item_count() != self.rows.len() {
-            self.list_state.reset(self.rows.len());
+        let new_rows =
+            Self::compute_rows(self.channel_id, &self.settings, &self.debounced_query, cx);
+        let old_count = self.rows.len();
+        let new_count = new_rows.len();
+        let prefix = file_rows_common_prefix(&self.rows, &new_rows);
+        self.rows = Rc::new(new_rows);
+        if old_count == 0 || prefix == 0 {
```
huylexuan44, line 216: Uploader enrichment rewrites row 0, so prefix==0 and reset() snaps a scrolled user back to the top.

crates/mezon-ui/src/chat/files_popover.rs
```diff
@@ -156,10 +160,40 @@ impl FilesPopoverPanel {
...
+                let user_scrolled = event.visible_range.start > 0;
+                let Some(this) = weak.upgrade() else {
+                    return;
+                };
+                this.update(cx, |this, cx| {
+                    if !user_scrolled {
+                        if this.viewport_autofill_remaining == 0 {
+                            return;
+                        }
+                        this.viewport_autofill_remaining -= 1;
```
huylexuan44, line 188: Budget is decremented before fetch_page, which no-ops while is_loading; smooth-wheel frames drain it to 0.

crates/mezon-ui/src/chat/files_popover.rs
```diff
@@ -147,6 +150,7 @@ impl FilesPopoverPanel {
             list_state,
             focus_handle,
             rows: Rc::new(Vec::new()),
+            viewport_autofill_remaining: MAX_VIEWPORT_AUTOFILL_PAGES,
```
huylexuan44, line 153: Gallery resets this budget on every store reset; the refresh and retry paths here leave it at 0 permanently.

crates/mezon-ui/src/chat/files_popover.rs
```diff
@@ -173,14 +207,17 @@ impl FilesPopoverPanel {
...
-        self.rows = Rc::new(Self::compute_rows(
-            self.channel_id,
-            &self.settings,
-            &self.debounced_query,
-            cx,
-        ));
-        if self.list_state.item_count() != self.rows.len() {
-            self.list_state.reset(self.rows.len());
+        let new_rows =
+            Self::compute_rows(self.channel_id, &self.settings, &self.debounced_query, cx);
```
huylexuan44, line 211: Search now filters only the loaded pages, so a file on a later page returns No files with no way to load it.

crates/mezon-store/src/files.rs
```diff
@@ -219,35 +256,73 @@ impl FilesStore {
...
             let _ = this.update(cx, |this, cx| {
                 let succeeded = match mapped {
-                    Ok(docs) => {
+                    Ok((docs, raw_count)) => {
                         let entry = this.ensure_channel(channel_id);
                         entry.is_loading = false;
-                        entry.documents = docs;
+                        entry.fetch_error = false;
+                        let added =
+                            merge_documents(&mut entry.documents, &mut entry.ids, docs, reset);
```
huylexuan44, line 316: documents is now unbounded, so enrich_channel plus refresh_rows rebuild every loaded page on each member notify.

crates/mezon-store/src/files.rs
```diff
@@ -336,9 +380,27 @@ fn sort_desc_in_place(items: &mut [ChannelDocument]) {
     });
 }
 
-fn dedupe_by_id(docs: Vec<ChannelDocument>) -> Vec<ChannelDocument> {
-    let mut seen = std::collections::HashSet::new();
-    docs.into_iter().filter(|doc| seen.insert(doc.id)).collect()
+fn merge_documents(
```
huylexuan44, line 383: Re-sorts all accumulated docs per page; reuse gallery's merge_two_desc_sorted to merge the two sorted runs.

crates/mezon-store/src/files.rs
```diff
@@ -219,35 +256,73 @@ impl FilesStore {
             .expect("channel just ensured")
     }
 
-    fn fetch(&mut self, clan_id: ClanId, channel_id: ChannelId, cx: &mut Context<Self>) {
+    fn fetch(
+        &mut self,
+        clan_id: ClanId,
+        channel_id: ChannelId,
+        _direction: LoadDirection,
```
huylexuan44, line 263: _direction is ignored here and fetch_page no-ops on After. Drop the parameter or implement the After branch.

crates/mezon-ui/src/chat/files_popover.rs
```diff
@@ -156,10 +160,40 @@ impl FilesPopoverPanel {
             _debounce_task: Task::ready(()),
             _subs: subs,
         };
+        this.install_scroll_handler(cx);
         this.refresh_rows(cx);
         this
     }
 
+    fn install_scroll_handler(&self, cx: &mut Context<Self>) {
```
huylexuan44, line 168: Verbatim copy of gallery's scroll handler with different constants and a dropped reset. Share one helper.

crates/mezon-store/src/files.rs
giachuvan: update latest backend logic change FILE to dock type.

crates/mezon-store/src/files.rs
giachuvan: next_page_has_more requires both a full page and added > 0 on purpose, so we do not loop on a page of already-seen or filtered-out rows. Files now requests file_type=doc, so the server already returns documents; the client is_document filter almost never empties a page. Stopping when added == 0 is the correct end condition, not a stuck cursor.

crates/mezon-ui/src/chat/files_popover.rs
giachuvan: The list is replaced by the empty state when search matches nothing in the pages already loaded (see 11). A truly empty first page with file_type=doc means there are no documents, so there is nothing to paginate. We do not keep a hidden list mounted just to fire scroll events.

crates/mezon-store/src/files.rs
giachuvan: The new attachment contract is five lowercase categories: image, video, doc, sticker, audio. The Files tab is doc only. Voice/audio belongs in audio, not in this list. Matching old React (treat audio as a document) would fight the API change. The in-popover audio player is leftover from when FILE returned mixed types.

crates/mezon-ui/src/chat/files_popover.rs
giachuvan: GPUI list only dispatches set_scroll_handler from wheel / smooth-wheel, same as Gallery. Driving load-more from render would refetch on every frame while the viewport sits near the bottom. We are not diverging Files from that Gallery pattern in this PR.

crates/mezon-store/src/files.rs
giachuvan: list_channel_attachment filters url.is_empty() before Files (and Gallery) see the page. That can under-count by a few failed attachments and stop one page early. Same transport behavior Gallery already lives with. Changing the shared API client is out of scope for this Files/doc PR.

crates/mezon-ui/src/chat/files_popover.rs
giachuvan: Search is client-side over the pages already fetched (page size 100). A file on a later page will not match until the user scrolls and loads that page. React Files does the same on its loaded list and has no server search. Full-channel search would need a new API; not part of this change.

crates/mezon-store/src/files.rs
giachuvan: The buffer is meant to grow with load-more. Capping it would hide files the user already paged in. Uploader enrich + row rebuild on member updates is existing store behavior; we are not adding an arbitrary cap in this PR.

crates/mezon-ui/src/chat/files_popover.rs
giachuvan: The handler is similar, not a bug. Files dropped Gallery-only date-filter reset and uses different constants. Extracting a shared helper would couple two features without fixing a defect. Not doing that in this PR.

PR #328, opened 2026-08-13
Delete channel

giachuvan:


crates/mezon-ui/src/chat/channel_settings/mod.rs
```diff
@@ -435,6 +436,10 @@ impl ChannelSettingScreen {
         } else {
             mezon_i18n::t(locale, "channelSetting.fields.channelDelete.delete")
         };
+        let can_delete_channel = !ctx.is_thread && !ctx.is_welcome_channel;
```
huylexuan44: Delete gate ignores permissions. Use mezon_store::can_delete_channel(clan_id, channel_id, cx) like the sidebar does.

crates/mezon-store/src/channel.rs
```diff
@@ -3453,6 +3498,65 @@ impl ChannelList {
...
+    }
+
+    fn remove_deleted_child_threads_of_channel(
+        &mut self,
+        clan_id: ClanId,
+        parent_channel_id: ChannelId,
+        cx: &mut Context<Self>,
+    ) {
+        let child_ids = self.child_thread_ids(clan_id, parent_channel_id);
+        crate::threads::ThreadsStore::global(cx).update(cx, |store, cx| {
```
huylexuan44: ThreadsStore::global panics when the store is unregistered. Move this below the child_ids.is_empty() early return.

crates/mezon-store/src/channel.rs
```diff
@@ -3453,6 +3498,65 @@ impl ChannelList {
...
+                &mut self.user_channels,
+                &mut self.user_channels_order,
+                child_id,
+            );
+            self.clear_compose_draft(child_id, cx);
+        }
+        if active_cleared {
+            cx.emit(ChannelEvent::ActiveChannelChanged(None));
+        }
+        if removed_any {
```
huylexuan44, line 3556: Child thread badges are pruned after sync_clan_after_read runs. Re-sync here or the clan badge stays inflated.

crates/mezon-ui/src/channel_navigation.rs
```diff
@@ -17,36 +17,161 @@ pub fn navigate_after_thread_removed(
...
+    }
+
+    ChannelList::global(cx).update(cx, |list, cx| {
+        list.clear_compose_draft(removed_id, cx);
+    });
+
+    let target = fallback_channel_after_delete(cx, clan_id, removed_id);
+    match target {
+        Some(channel_id) => open_clan_channel(cx, clan_id, channel_id),
+        None => router::navigate(cx, Route::Chat),
```
huylexuan44: Use router::replace here. navigate pushes the deleted channel onto the back stack and clears forward history.

crates/mezon-store/src/channel.rs
```diff
@@ -246,6 +262,34 @@ fn archive_permission_for(
...
+    let is_creator = BadgeService::try_global(cx)
+        .and_then(|badges| badges.read(cx).current_user_id(cx))
+        .is_some_and(|me| me == channel.creator_id);
+    let Some(permissions) = PermissionStore::try_global(cx) else {
+        return is_creator;
+    };
+    let permissions = permissions.read(cx);
+    delete_allowed_by_server(
+        is_creator,
+        permissions.check(clan_id, None, PERMISSION_CLAN_OWNER, cx),
```
huylexuan44, line 286: Permissions are checked clan-scoped; React scopes them to the channel. Pass Some(channel_id) to honor overrides.

crates/mezon-store/src/threads.rs
```diff
@@ -405,6 +424,39 @@ impl ThreadsStore {
...
+    fn apply_parent_channel_deleted(&mut self, parent_id: &str, cx: &mut Context<Self>) {
+        let before = self.threads.len();
+        self.threads.retain(|t| t.parent_id != parent_id);
+        let mut changed = self.threads.len() != before;
+        if let Some(results) = self.search_results.as_mut() {
+            let results_before = results.len();
+            results.retain(|t| t.parent_id != parent_id);
+            changed |= results.len() != results_before;
+        }
+        if self.list_channel_id.as_deref() == Some(parent_id) {
```
huylexuan44, line 449: Also clear list_channel_id. Leaving it set lets ensure_loaded() re-fetch threads for the deleted parent.

crates/mezon-ui/src/chat/channel_settings/mod.rs
```diff
@@ -452,11 +457,26 @@ impl ChannelSettingScreen {
...
+                        .on_click(move |_, window, cx| {
+                            let locale = delete_locale.clone();
+                            Shell::global(cx).update(cx, |shell, cx| {
+                                shell.confirm_delete_channel(
+                                    clan_id, channel_id, &locale, window, cx,
+                                );
+                            });
+                        })
+                })
+                .when(!can_delete_channel, |el| {
```
huylexuan44, line 474: React only disables delete for the welcome channel; threads keep it. Thread delete is unreachable here.

crates/mezon-ui/src/channel_navigation.rs
```diff
@@ -17,36 +17,161 @@ pub fn navigate_after_thread_removed(
...
+    }
+}
+
+fn fallback_channel_after_delete(
+    cx: &App,
+    clan_id: ClanId,
+    removed_id: ChannelId,
+) -> Option<ChannelId> {
+    let channels = ChannelList::global(cx).read(cx);
+    let welcome = ClanList::global(cx)
```
huylexuan44, line 89: Fallback order welcome/default/remembered contradicts ensure_active_channel_for_clan, which resolves remembered first.

crates/mezon-ui/src/app/shell/confirm_delete_channel_modal.rs
```diff
@@ -0,0 +1,162 @@
...
+
+impl Render for ConfirmDeleteChannelModal {
+    fn render(&mut self, _window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        let theme = cx.theme();
+        let submitting = self.submitting;
+
+        v_flex()
+            .track_focus(&self.focus_handle)
+            .key_context("menu")
+            .on_action(cx.listener(|this, _: &::menu::Cancel, _window, cx| {
```
huylexuan44, line 98: React DeleteModal confirms on Enter. Add a menu::Confirm action so this destructive modal is keyboard-operable.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1220,16 +1237,84 @@ impl ChatLayout {
...
+                    Some(DeletedRedirect::Channel { id }) => {
+                        crate::channel_navigation::navigate_after_channel_removed(cx, clan_id, id);
+                    }
+                    None => return,
+                }
+                self.focused_channel_id = None;
+                self.dismiss_threads_popover(cx);
+                self.dismiss_topic_panel(cx);
+            }
+            Route::ChannelSettings {
```
huylexuan44: Byte-for-byte copy of the Route::Channel arm. Merge into one or-pattern arm; all three expose clan_id/channel_id.

crates/mezon-store/src/channel.rs
```diff
@@ -3453,6 +3498,65 @@ impl ChannelList {
...
+                cats.iter()
+                    .flat_map(|c| &c.channels)
+                    .filter(|ch| ch.parent_id == Some(parent_channel_id))
+                    .map(|ch| ch.id)
+                    .collect()
+            })
+            .unwrap_or_default()
+    }
+
+    fn remove_deleted_child_threads_of_channel(
```
huylexuan44, line 3514: Near-duplicate of remove_child_threads_of_channel. Parameterize a single cascade so the two cannot drift.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -1220,16 +1237,84 @@ impl ChatLayout {
...
+                    match list.deleted_channel_parent(channel_id) {
+                        Some(parent)
+                            if parent != channel_id && list.channel_in_clan(clan_id, parent) =>
+                        {
+                            Some(DeletedRedirect::Thread { parent })
+                        }
+                        Some(parent) if parent != channel_id && list.is_locally_deleted(parent) => {
+                            Some(DeletedRedirect::Channel { id: parent })
+                        }
+                        _ => Some(DeletedRedirect::Channel { id: channel_id }),
```
huylexuan44: This arm makes redirect always Some, so the None => return below is dead. Drop the Option wrapper.

crates/mezon-store/src/channel.rs
```diff
@@ -209,12 +209,28 @@ pub fn archive_allowed_by_server(
...
+    is_creator || has_owner || has_administrator || has_manage_clan || has_manage_channel
+}
+
 pub fn can_archive_channel(clan_id: ClanId, channel_id: ChannelId, cx: &App) -> bool {
     ChannelList::global(cx)
         .read(cx)
         .can_archive_channel_for(clan_id, channel_id, cx)
 }
 
+pub fn can_delete_channel(clan_id: ClanId, channel_id: ChannelId, cx: &App) -> bool {
```
huylexuan44, line 228: can_delete_channel is pub but never called. Wire it into the Channel Settings delete gate or drop it.

crates/mezon-ui/src/channel_navigation.rs
```diff
@@ -17,36 +17,161 @@ pub fn navigate_after_thread_removed(
...
+        .filter(|id| *id != removed_id && channels.channel_in_clan(clan_id, *id));
+    pick_fallback_channel_after_delete(removed_id, welcome, default_text, remembered)
+}
+
+fn pick_fallback_channel_after_delete(
+    removed_id: ChannelId,
+    welcome: Option<ChannelId>,
+    default_text: Option<ChannelId>,
+    remembered: Option<ChannelId>,
+) -> Option<ChannelId> {
```
huylexuan44, line 106: Callers already filter removed_id, so this reduces to welcome.or(default_text).or(remembered).

crates/mezon-ui/src/channel_navigation.rs
```diff
@@ -17,36 +17,161 @@ pub fn navigate_after_thread_removed(
...
+                        && list.deleted_channel_parent(active) == Some(removed_id))
+            }
+            _ => false,
+        }
+    };
+    if !needs_nav {
+        return;
+    }
+
+    ChannelList::global(cx).update(cx, |list, cx| {
```
huylexuan44: apply_local_delete already clears this draft, and this runs only when needs_nav. Keep draft clearing in the store.

crates/mezon-store/src/channel.rs
giachuvan: PermissionStore::check only uses channel_id for overridden slugs (manage-thread, send-message, delete-message). manage-channel, manage-clan, owner, and admin are clan-scoped. Passing Some(channel_id) here is a no-op in this codebase and would not honor channel overrides. We keep the existing clan-scoped checks, consistent with the sidebar menu.

crates/mezon-ui/src/chat/channel_settings/mod.rs
giachuvan: This is intentional and matches develop. Context-menu Delete Thread is already shipped (confirm_delete_thread). Channel Settings still renders the Delete Thread label as a stub (no on_click), same as develop. This PR only wires Settings Delete Channel. Wiring Settings Delete Thread is a separate follow-up so we do not mix it into this change.

crates/mezon-ui/src/channel_navigation.rs
giachuvan: These are different paths. After deleting the channel you are viewing, we resolve: welcome (if set and still in the clan) → first remaining top-level text channel → remembered → Route::Chat. A clan with no welcome channel skips step 1 and lands on the next text channel. ensure_active_channel_for_clan (remembered → welcome → default) is for clan switch / opening Chat, not delete-nav. Welcome-first on delete is a locked product choice: we are not porting Electron’s sibling / channels[0] fallback.

crates/mezon-store/src/channel.rs
giachuvan: remove_child_threads_of_channel (archive) and remove_deleted_child_threads_of_channel (delete) look similar but differ on purpose: tombstones (archived_* vs deleted_*), active-channel redirect (parent vs clear), ThreadsStore (mark_thread_archived vs remove_threads_of_parent), and draft keys. A shared helper would risk drifting archive behavior. We keep them separate.

PR #337, opened 2026-08-16
Give flowers in voice call

giachuvan:


crates/mezon-client/src/transport.rs
```diff
@@ -220,6 +222,7 @@ impl TryFrom<realtime::envelope::Message> for RealtimeEvent {
             realtime::envelope::Message::VoiceJoinedEvent(m) => Ok(Self::VoiceJoined(m)),
             realtime::envelope::Message::VoiceLeavedEvent(m) => Ok(Self::VoiceLeaved(m)),
             realtime::envelope::Message::VoiceReactionSend(m) => Ok(Self::VoiceReaction(m)),
+            realtime::envelope::Message::VoiceInteractiveEvent(m) => Ok(Self::VoiceInteractive(m)),
```
huylexuan44, line 225: mezon-protocol submodule is unchanged from develop and has no VoiceInteractiveEvent — bump it, this does not compile.

crates/mezon-store/src/voice.rs
```diff
@@ -1011,6 +1043,139 @@ impl VoiceStore {
...
+    ) {
+        let key = flower_effect_key(&giver_id, &receiver_id, timestamp);
+        if self
+            .displayed_flowers
+            .iter()
+            .any(|flower| flower.key == key)
+        {
+            return;
+        }
+        self.displayed_flowers.clear();
```
huylexuan44, line 1216: displayed_flowers.clear() right after the dedup check kills dedup — the sender own echo replays the burst.

crates/mezon-ui/src/app/shell/mod.rs
```diff
@@ -900,33 +903,29 @@ impl Shell {
...
+            window.focus(&focus_handle, cx);
         }
     }
 
-    pub fn dismiss_modal(&mut self, window: &mut Window, cx: &mut Context<Self>) {
-        if self.modal.take().is_none() {
-            return;
-        }
-        if let Some((underlay, fullscreen, command_palette_open, focus_handle)) =
+    pub fn close_modal(&mut self, cx: &mut Context<Self>) -> Option<gpui::FocusHandle> {
```
huylexuan44: close_modal now restores the underlay instead of dismissing it, and callers drop the returned handle — Esc/Enter go dead.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -2420,12 +2610,24 @@ fn participant_menu_trigger(
     }
 }
 
+fn flower_menu_disabled(cx: &App) -> bool {
```
huylexuan44, line 2623: Unknown balance (None) reads as cannot afford — Give Flowers stays greyed out for a funded wallet.

crates/mezon-ui/src/app/shell/wallet_not_available_modal.rs
```diff
@@ -94,18 +136,28 @@ impl Render for WalletNotAvailableModal {
...
+                                this.hover(|this| this.bg(rgb(ENABLE_BUTTON_BG_HOVER)))
+                                    .cursor_pointer()
+                            })
+                            .when(enabling, |this| this.opacity(0.7))
                             .text_color(white())
                             .font_weight(gpui::FontWeight::MEDIUM)
-                            .child(self.enable_label.clone())
-                            .on_click(|_, window, cx| {
+                            .child(if enabling {
+                                SharedString::from("Enabling…")
```
huylexuan44: enable_wallet emits nothing when the generation changes — enabling stays true and the button is stuck here forever.

crates/mezon-store/src/voice.rs
```diff
@@ -1446,6 +1611,172 @@ impl VoiceStore {
...
+                            card_text,
+                            cx,
+                        )
+                    });
+                    card.detach();
+                    this.show_flower_effect(local_id, identity, timestamp, cx);
+                })
+                .ok();
+                if let Err(error) = api
+                    .write_voice_interactive(
```
huylexuan44, line 1905: Fire-and-forget after the transfer succeeded — a failed write burns 50k tokens with no flower and no retry.

crates/mezon-store/src/voice.rs
```diff
@@ -1446,6 +1611,172 @@ impl VoiceStore {
...
+        match can_give_flower(
+            identity == local_id,
+            wallet_available,
+            pending,
+            self.last_flower_send,
+            Instant::now(),
+            balance.as_deref(),
+        ) {
+            Err(
+                GiveFlowerDeny::SelfTarget
```
huylexuan44, line 1807: Deny branches return with no toast, and set_pending_give_flower is applied without cx.notify() so menus stay stale.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1228,6 +1232,189 @@ fn reactions_overlay(
     )
 }
 
+fn flower_burst(flower: &DisplayedFlower, theme: &Theme, locale: &str) -> AnyElement {
```
huylexuan44: Particles use started_at but the caption uses with_animation (first paint) — they desync if the view was unmounted.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1228,6 +1232,189 @@ fn reactions_overlay(
...
+    "icons/flower-daisy.svg",
+    "icons/flower-stems.svg",
+    "icons/flower-lotus.svg",
+    "icons/flower-tied.svg",
+    "icons/flower-sun.svg",
+    "icons/flower-ring.svg",
+    "icons/flower-bud.svg",
+];
+
+fn paint_flower_burst(
```
huylexuan44, line 1346: 128 paint_svg per frame at varying sizes fills the never-evicted Monochrome atlas — quantize sizes and cull offscreen.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1228,6 +1232,189 @@ fn reactions_overlay(
...
 
+fn flower_burst(flower: &DisplayedFlower, theme: &Theme, locale: &str) -> AnyElement {
+    let label = SharedString::from(
+        mezon_i18n::t(locale, "channelVoice.giveFlowerGiven")
+            .replace("{{giver}}", &flower.giver_name)
+            .replace("{{receiver}}", &flower.receiver_name),
+    );
+    let id = flower.timestamp.unsigned_abs();
+    let duration = FLOWER_ANIMATION_TTL;
+    let particles = flower.particles.clone();
```
huylexuan44, line 1238: Clones 128 particles and rebuilds the caption every frame for 6s — precompute the SharedString and pass a slice.

crates/mezon-store/src/gifts.rs
```diff
@@ -0,0 +1,494 @@
...
+struct FlowerExtraAttribute {
+    gift_type: String,
+    voice_channel_id: String,
+}
+
+pub fn flower_price() -> i64 {
+    FLOWER_PRICE
+}
+
+pub fn format_flower_amount(n: i64) -> String {
```
huylexuan44, line 154: format_flower_amount duplicates format_thousands (messages.rs:7274) — reuse it; flower_price() only wraps the const.

crates/mezon-store/src/voice.rs
```diff
@@ -1011,6 +1043,139 @@ impl VoiceStore {
...
+            receiver_name,
+            timestamp,
+            particles: flower_particles(timestamp.unsigned_abs()),
+            started_at: Instant::now(),
+            _remove_timer: remove_timer,
+        });
+        cx.notify();
+    }
+
+    fn resolve_flower_name(&self, user_id: &str, cx: &App) -> String {
```
huylexuan44, line 1255: resolve_flower_name duplicates resolve_reaction_name, which returns an empty name for DM/self — extract one resolver.

crates/mezon-ui/src/components/primitives/context_menu.rs
```diff
@@ -445,17 +474,28 @@ impl RenderOnce for ContextMenu {
                             })
                             .when(leading_icon.is_some(), |row| row.gap_2())
                             .when_some(leading_icon, |row, icon| {
-                                row.child(Icon::new(icon).size_4().text_color(icon_color))
+                                let icon_el = if icon == IconName::Flower {
```
huylexuan44, line 477: Hard-coded IconName::Flower branch in shared menu code — add an icon size option on the item builder instead.

crates/mezon-store/src/voice.rs
```diff
@@ -252,6 +265,18 @@ pub struct DisplayedReaction {
     _remove_timer: Task<()>,
 }
 
+pub struct DisplayedFlower {
```
huylexuan44, line 282: displayed_flowers can never exceed len 1 (cleared on insert) — use Option; giver_id/receiver_id are never read.

crates/mezon-ui/src/app/shell/wallet_not_available_modal.rs
```diff
@@ -17,11 +18,52 @@ pub(super) struct WalletNotAvailableModal {
...
+        enable_label: SharedString,
+        cancel_label: SharedString,
+        cx: &mut Context<Self>,
+    ) -> Self {
+        let wallet = WalletStore::global(cx);
+        let wallet_sub = cx.subscribe(&wallet, |this, _, event, cx| match event {
+            WalletEvent::Enabled => {
+                this.enabling = false;
+                Shell::global(cx).update(cx, |shell, cx| {
+                    shell.success("Wallet enabled", cx);
```
huylexuan44: Wallet enabled here and Enabling on line 147 are hardcoded English — route through mezon_i18n like the other labels.

crates/mezon-store/src/voice.rs
giachuvan: Incorrect. Dedup compares key before clear(). The sender echo reuses the same timestamp/params, so it returns and does not replay. clear() only replaces a different gift so one burst is on screen at a time.

crates/mezon-store/src/voice.rs
giachuvan: Pay-then-notify is the correct order — writing first would fan out an unpaid gift. The local burst already runs after a successful transfer. Retrying the write is unsafe: a timeout after the server accepted the event would double-broadcast. This matches voice reactions (log and move on).

crates/mezon-store/src/voice.rs
giachuvan: Those deny arms are unreachable from the UI (self is hidden, pending disables the item, missing wallet opens the modal). Rate-limit and insufficient already toast. Pending is set after the participant menu closes; the next open reads the new flag.

crates/mezon-ui/src/chat/voice.rs
giachuvan: Performance suggestion, not a logic bug. Quantizing sizes would change the 1.0→1.5 grow. The burst lives 7s then is removed. Offscreen culling can land later if the monochrome atlas actually grows.

crates/mezon-store/src/gifts.rs
giachuvan: format_thousands is private in a 10k-line messages.rs. Promoting it just to DRY flowers risks that message path. flower_price() is the test/API seam so callers do not scatter 50_000.

crates/mezon-store/src/voice.rs
giachuvan: resolve_reaction_name intentionally returns empty for DM/self so reaction chips stay nameless. Flower captions always need a name (clan nick → display → username). One shared resolver would change reaction UX.

crates/mezon-ui/src/components/primitives/context_menu.rs
giachuvan: An icon_size option on every menu item is a wide API change for one glyph. The Flower branch is isolated and does not change other items.

crates/mezon-store/src/voice.rs
giachuvan: Vec plus clear/push/retain matches the reaction overlay. Switching to Option is a refactor of the timer and overlay with no behavior change. Ids are already in the dedup key; keeping the fields is free.

crates/mezon-ui/src/app/shell/mod.rs
```diff
@@ -900,33 +905,37 @@ impl Shell {
...
+        let focus = if let Some((underlay, fullscreen, command_palette_open, focus_handle)) =
             self.modal_underlay.take()
         {
             self.modal = Some(underlay);
             self.modal_fullscreen = fullscreen;
             self.command_palette_open = command_palette_open;
-            if let Some(focus_handle) = focus_handle {
-                window.focus(&focus_handle, cx);
-            }
+            focus_handle
```
huylexuan44, line 922: close_modal pops to the underlay instead of clearing it — navigate_from_external_trigger now leaves that modal on screen.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1228,6 +1232,174 @@ fn reactions_overlay(
...
+    );
+    for particle in particles {
+        let pose = flower_particle_pose(particle, elapsed, ttl);
+        if pose.opacity < 0.02 {
+            continue;
+        }
+        let position = point(origin.x + px(pose.x * unit), origin.y + px(pose.y * unit));
+        let slot = (particle.palette as usize) % colors.len();
+        let color = colors[slot].opacity(pose.opacity);
+        let size = particle.size * pose.scale;
```
huylexuan44: size_px varies continuously; monochrome atlas tiles are never evicted, so ~1300 tiles leak. Round size_px to whole pixels.

crates/mezon-store/src/voice.rs
```diff
@@ -1446,6 +1660,172 @@ impl VoiceStore {
...
+                    let card = DirectMessageStore::global(cx).update(cx, |store, cx| {
+                        store.create_dm_and_send_token_card(
+                            receiver_id,
+                            receiver_name,
+                            card_text,
+                            cx,
+                        )
+                    });
+                    card.detach();
+                    this.show_flower_effect(local_id, identity, timestamp, cx);
```
huylexuan44: No room re-check after the transfer — leaving mid-flight still plays the burst and sound out of call.

crates/mezon-store/src/voice.rs
```diff
@@ -1446,6 +1660,172 @@ impl VoiceStore {
...
+                })
+                .ok();
+                if let Err(error) = api
+                    .write_voice_interactive(
```
huylexuan44, line 1905: write_voice_interactive still targets the old channel if the user left during the transfer.

crates/mezon-store/src/gifts.rs
```diff
@@ -0,0 +1,578 @@
...
+    }
+    if pending {
+        return Err(GiveFlowerDeny::Pending);
+    }
+    if last_send.is_some_and(|sent| now.duration_since(sent) < FLOWER_RATE_LIMIT) {
+        return Err(GiveFlowerDeny::RateLimited);
+    }
+    match balance {
+        Some(value) if can_afford(value, flower_price()) => Ok(()),
+        _ => Err(GiveFlowerDeny::Insufficient),
```
huylexuan44: balance None denies here but flower_menu_blocked allows it — funded wallet with unknown balance always toasts Insufficient.

crates/mezon-store/src/voice.rs
```diff
@@ -1011,6 +1082,149 @@ impl VoiceStore {
...
+            receiver_id,
+            giver_name,
+            receiver_name,
+            timestamp,
+            particles: Arc::new(flower_particles(timestamp.unsigned_abs())),
+            started_at: Instant::now(),
+            label,
+            _remove_timer: remove_timer,
+        });
+        self.play_flower_sound(cx);
```
huylexuan44: No receive-side throttle; emoji reactions use last_emoji_at and sounds use sound_throttle. Add one for flowers.

crates/mezon-store/src/voice.rs
```diff
@@ -1011,6 +1082,149 @@ impl VoiceStore {
...
+            msg.user_id,
+            msg.voice_channel_id,
+            msg.clan_id,
+            &msg.params,
+            joined_channel,
+            joined_clan,
+        ) else {
+            return;
+        };
+        self.show_flower_effect(giver_id, receiver_id, timestamp, cx);
```
huylexuan44, line 1191: giver_id and receiver_id are not checked against self.participants — any id in params renders and plays the sound.

crates/mezon-store/src/gifts.rs
```diff
@@ -0,0 +1,578 @@
...
+    voice_channel_id: i64,
+    clan_id: i64,
+    params: &str,
+    joined_channel_id: i64,
+    joined_clan_id: i64,
+) -> Option<(String, String, i64, String)> {
+    if VoiceInteractiveEventType::from_i32(event_type) != Some(VoiceInteractiveEventType::Gift) {
+        return None;
+    }
+    if voice_channel_id != joined_channel_id || clan_id != joined_clan_id {
```
huylexuan44: clan_id equality adds nothing over the unique voice_channel_id and drops the event if the server sends clan_id 0.

crates/mezon-ui/src/app/shell/mod.rs
```diff
@@ -900,33 +905,37 @@ impl Shell {
...
             self.modal_fullscreen = false;
-        }
+            None
+        };
         cx.notify();
+        if let Some(focus_handle) = focus.clone()
+            && let Some(window_handle) =
+                crate::app::main_window::handle(cx).or_else(|| cx.active_window())
+        {
+            let _ = cx.update_window(window_handle, |_, window, cx| {
```
huylexuan44: update_window fails re-entrantly, so this focus restore no-ops for callers already inside a window update.

crates/mezon-store/src/wallet.rs
```diff
@@ -474,12 +496,20 @@ impl WalletStore {
                 ),
             };
             this.update(cx, |this, cx| {
-                if this.enabling_user.as_deref() == Some(user_id.as_str()) {
-                    this.enabling_user = None;
-                }
                 if this.reset_generation != generation {
+                    if report_failure {
+                        cx.emit(WalletEvent::EnableFailed {
+                            message: "Wallet enable was cancelled".to_string(),
```
huylexuan44: Wallet enable was cancelled and Wallet is not configured are hardcoded English; route through mezon_i18n.

crates/mezon-store/src/voice.rs
giachuvan: Correct and required. The event belongs to the room the gift was paid for, not wherever the sender is now. Retargeting would fan out to the wrong call. Skipping the write after leave means the room never sees flowers after a 50k debit. Local burst/sound is already gated in #3.

crates/mezon-store/src/voice.rs
giachuvan: The overlay is room-wide, not tile-bound. The giver can leave right after send and the room must still see it. The receiver may not be in participants yet (join race) or may have just left. A participant check would drop valid gifts. Channel + gift_type + the 1s receive throttle are the right filters. A forged id only affects the resolved caption, it does not crash.

crates/mezon-store/src/voice.rs
```diff
@@ -1011,6 +1084,153 @@ impl VoiceStore {
...
+            .displayed_flowers
+            .iter()
+            .any(|flower| flower.key == key)
+        {
+            return;
+        }
+        let now = Instant::now();
+        if self
+            .last_flower_effect_at
+            .is_some_and(|last| now.duration_since(last) < FLOWER_RATE_LIMIT)
```
huylexuan44: Global 1s throttle drops a second giver flower entirely; key it per giver or throttle only the sound.

crates/mezon-store/src/voice.rs
```diff
@@ -1446,6 +1666,179 @@ impl VoiceStore {
...
+                            cx,
+                        )
+                    });
+                    card.detach();
+                    let still_in_room = this
+                        .connection
+                        .connected_channel()
+                        .is_some_and(|(channel, _)| channel == gift_channel);
+                    if still_in_room {
+                        this.show_flower_effect(local_id, identity, timestamp, cx);
```
huylexuan44, line 1900: Another flower within 1s swallows the sender own burst and its echo after a 50k debit.

crates/mezon-ui/src/app/shell/mod.rs
```diff
@@ -900,33 +905,40 @@ impl Shell {
...
+        cx.notify();
+        focus
+    }
+
+    pub fn close_modal(&mut self, cx: &mut Context<Self>) -> Option<gpui::FocusHandle> {
+        let focus = self
+            .modal_underlay
+            .take()
+            .and_then(|(_, _, _, focus_handle)| focus_handle);
+        self.modal.take();
```
huylexuan44, line 937: Lost the is_some() guard, so notify and state clearing now run even with no modal open.

crates/mezon-ui/src/app/shell/mod.rs
```diff
@@ -900,33 +905,40 @@ impl Shell {
...
+        self.command_palette_open = false;
+        self.modal_fullscreen = false;
         cx.notify();
+        focus
```
huylexuan44: Returns the underlay focus handle but discards the underlay, and every caller drops it. Return unit.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1228,6 +1232,202 @@ fn reactions_overlay(
...
+        let pose = flower_particle_pose(particle, elapsed, ttl);
+        if pose.opacity < 0.02 {
+            continue;
+        }
+        let position = point(origin.x + px(pose.x * unit), origin.y + px(pose.y * unit));
+        let visual = particle.size * pose.scale;
+        if visual < 1.0 {
+            continue;
+        }
+        let radius = px(visual * 0.5);
```
huylexuan44: Rotated sprite reaches visual*0.707 from centre, so cull radius visual*0.5 pops edge particles early.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1228,6 +1232,202 @@ fn reactions_overlay(
...
+        bounds.origin.y + bounds.size.height * 0.46,
+    );
+    for particle in particles {
+        let pose = flower_particle_pose(particle, elapsed, ttl);
+        if pose.opacity < 0.02 {
+            continue;
+        }
+        let position = point(origin.x + px(pose.x * unit), origin.y + px(pose.y * unit));
+        let visual = particle.size * pose.scale;
+        if visual < 1.0 {
```
huylexuan44: Dead branch: visual is always >= 24.

crates/mezon-store/src/voice.rs
giachuvan: Agreed. The global 1s gate dropped the whole overlay, so a second giver (or the sender’s own burst after a 50k debit) vanished and the echo could not land either. Overlay always replaces now; FLOWER_RATE_LIMIT applies to the SFX only. Echo stays key-deduped; send stays 1s via last_flower_send. We did not add a per-giver map: the overlay is already single-slot, local send is already limited, and a HashMap would only add allocations.

crates/mezon-store/src/voice.rs
giachuvan: Agreed. The global 1s gate dropped the whole overlay, so a second giver (or the sender’s own burst after a 50k debit) vanished and the echo could not land either. Overlay always replaces now; FLOWER_RATE_LIMIT applies to the SFX only. Echo stays key-deduped; send stays 1s via last_flower_send. We did not add a per-giver map: the overlay is already single-slot, local send is already limited, and a HashMap would only add allocations.

crates/mezon-ui/src/app/shell/mod.rs
giachuvan: Agreed. Restored an early return when both modal and modal_underlay are empty so navigation does not notify for nothing. We still hard-clear a leftover underlay (prior review).

crates/mezon-ui/src/app/shell/mod.rs
giachuvan: Agreed. close_modal returns (). Esc focus still goes through dismiss_modal → pop_modal.

crates/mezon-ui/src/chat/voice.rs
giachuvan: Agreed. Cull radius is now visual / √2 so a 45° sprite is not clipped at the edge.

crates/mezon-ui/src/chat/voice.rs
giachuvan: Agreed. Removed the dead visual < 1.0 branch (size is 24–32, scale 1.0–1.5).

PR #357, opened 2026-08-18
feat: button voice interactive app in voice channel

duocphungvanNCC:


crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +856,120 @@ impl VoiceStore {
...
+                event.voice_channel_id,
+                &event.params,
+                joined_channel,
+            ) else {
+                return;
+            };
+            self.show_flower_effect(giver_id, receiver_id, timestamp, cx);
+            return;
+        }
+        let targets_current_user = event.receiver_id == 0
```
huylexuan44: receiver_id == 0 lets any peer open a browser window on every participant. Gate on receiver_id == me, or confirm before launch.

crates/mezon-client/src/transport.rs
```diff
@@ -4415,15 +4416,37 @@ impl MezonTransport {
...
-        let (code, _response) = self.send(cid, encode_envelope_cid_last(envelope)).await?;
+        tracing::info!(
+            target: "socket",
+            cid = i32::from(cid),
+            clan_id,
+            voice_channel_id,
+            event_type,
+            "sending VoiceInteractiveEvent"
+        );
+        let (code, response) = self
```
huylexuan44: response is only read for its length; if the server returns params on this cid they are discarded and the launch never fires.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +856,120 @@ impl VoiceStore {
...
-                |this, event, cx| this.handle_voice_interactive(event, cx),
+                |this, event, cx| {
+                    if let RealtimeEvent::VoiceInteractive(event) = event {
+                        tracing::info!(
+                            clan_id = event.clan_id,
+                            voice_channel_id = event.voice_channel_id,
+                            sender_id = event.sender_id,
+                            receiver_id = event.receiver_id,
+                            event_type = event.event_type,
+                            params = %event.params,
```
huylexuan44: params is the signed web_app_data credential; CLAUDE.md forbids logging it at any level. Drop this field.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +856,120 @@ impl VoiceStore {
...
+                },
             );
         });
     }
 
+    fn handle_voice_interactive(&mut self, event: &RealtimeEvent, cx: &mut Context<Self>) {
+        let RealtimeEvent::VoiceInteractive(event) = event else {
+            return;
+        };
+        if event.event_type == VoiceInteractiveEventType::Gift as i32 {
```
huylexuan44, line 882: Protocol bump moved event_type 4 to 5 and added receiver_id at 4, so gifts from old peers decode as 0 and are silently dropped.

crates/mezon-client/src/transport.rs
```diff
@@ -4415,15 +4416,37 @@ impl MezonTransport {
...
+            .send_with_timeout(
+                cid,
+                encode_envelope_cid_last(envelope),
+                Duration::from_millis(2000),
```
huylexuan44: 2000 ms also applies to give_flower, which used the 10 s default; a slow ack now loses the gift after the wallet was debited.

crates/mezon-store/src/voice.rs
```diff
@@ -2670,6 +2786,85 @@ impl VoiceStore {
...
+                        sender_id = user_id,
+                        receiver_id = user_id,
+                        event_type = app.event_type(),
+                        "VoiceInteractiveEvent acknowledged; waiting for realtime params"
+                    );
+                }
+                Err(error) => {
+                    tracing::warn!("write_voice_interactive_event failed: {error:#}");
+                    let _ = this.update(cx, |store, cx| {
+                        store.open_interactive_app_url(app, None, clan_id, cx)
```
huylexuan44: Timeout fallback opens the bare base URL with no data param, and the late cid reply then opens a second window with the real params.

crates/mezon-store/build.rs
```diff
@@ -14,6 +14,9 @@ const BAKED_ENV_VARS: &[&str] = &[
     "NX_CHAT_APP_STREAM_WS_URL",
     "NX_CHAT_APP_MEET_WS_URL",
     "NX_CHAT_APP_NOTIFICATION_WS_URL",
+    "NX_CHAT_APP_BLACKBOARD_URL",
+    "NX_CHAT_APP_QUIZ_URL",
+    "NX_CHAT_APP_INTERACTIVE_URL",
```
huylexuan44, line 19: These three are not set in release.yml or build.yml, so shipped builds bake the .vn dev defaults. Add them like STREAM_WS_URL.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -3124,6 +3124,37 @@ fn control_bar(
...
+            neutral_bg,
+            neutral_hover,
+            IconName::Joystick,
+            theme.text_muted,
+        )
+        .tooltip(Tooltip::text(mezon_i18n::t(
+            locale,
+            "channelVoice.openInteractiveApp",
+        )))
+        .on_click(move |event, _, cx| {
```
huylexuan44: The menu backdrop closes on mouse-down and this on_click re-opens on mouse-up, so the button can never close its own menu.

crates/mezon-store/src/voice.rs
```diff
@@ -215,6 +241,7 @@ pub struct VoiceStore {
     frame_store: Option<Arc<VideoFrameStore>>,
     camera_devices: Vec<CameraDeviceInfo>,
     device_menu: Option<DeviceMenuKind>,
+    interactive_app_menu_position: Option<gpui::Point<gpui::Pixels>>,
```
huylexuan44: teardown() clears device_menu, device_submenu and participant_menu but not this field, so the menu reopens on the next join.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +856,120 @@ impl VoiceStore {
...
+        cx: &App,
+    ) {
+        let config = AppConfig::global(cx);
+        let base_url = match app {
+            VoiceInteractiveApp::Quiz => &config.quiz_url,
+            VoiceInteractiveApp::Blackboard => &config.blackboard_url,
+            VoiceInteractiveApp::Interactive => &config.interactive_url,
+        };
+        let clan_id_string = clan_id.to_string();
+        let clan_name = crate::ClanList::global(cx)
```
huylexuan44: ClanList::global panics if the global is unset, and this runs from an unsolicited push. Use try_global like PlatformStore below.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -3124,6 +3124,37 @@ fn control_bar(
...
+            voice_for_click.update(cx, |store, cx| {
+                let position = event.position();
+                store.toggle_interactive_app_menu(
+                    point(position.x - px(110.), position.y - px(160.)),
```
huylexuan44: event.position() is the cursor, not the button bounds, and -160 assumes a fixed menu height. Anchor to the button instead.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -3449,6 +3481,30 @@ fn circle_button(
...
 
+fn interactive_app_menu(voice: &Entity<VoiceStore>) -> ContextMenu {
+    let dismiss = {
+        let voice = voice.clone();
+        move |_window: &mut Window, cx: &mut App| {
+            voice.update(cx, |store, cx| store.close_interactive_app_menu(cx));
+        }
+    };
+    [
+        ("Quiz", VoiceInteractiveApp::Quiz),
```
huylexuan44: Quiz/Blackboard/Interactive are hard-coded English while the tooltip got an i18n key in all 11 locales. Route them through t().

crates/mezon-store/src/voice.rs
```diff
@@ -47,6 +47,32 @@ pub enum DeviceKind {
     VideoInput,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VoiceInteractiveApp {
```
huylexuan44: Duplicates VoiceInteractiveEventType AppQuiz/AppBlackboard/AppInteractive in gifts.rs, which already has from_i32 and tests.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +856,120 @@ impl VoiceStore {
...
+            return;
+        }
+        let Some(app) = VoiceInteractiveApp::from_event_type(event.event_type) else {
+            return;
+        };
+        let params = (!event.params.is_empty()).then_some(event.params.as_str());
+        self.open_interactive_app_url(app, params, event.clan_id, cx);
+    }
+
+    fn open_interactive_app_url(
```
huylexuan44: This launch tail duplicates channel_app::open_in_browser; move it onto PlatformStore so both call sites share one helper.

crates/mezon-store/src/voice.rs
```diff
@@ -47,6 +47,32 @@ pub enum DeviceKind {
...
+impl VoiceInteractiveApp {
+    pub fn event_type(self) -> i32 {
+        match self {
+            Self::Quiz => 10,
+            Self::Blackboard => 11,
+            Self::Interactive => 12,
+        }
+    }
+
+    fn from_event_type(value: i32) -> Option<Self> {
```
huylexuan44: No tests for from_event_type or the new receiver-targeting branch; CLAUDE.md requires tests for realtime event handling.

crates/mezon-store/src/voice.rs
```diff
@@ -2674,6 +2801,69 @@ impl VoiceStore {
...
+            return;
+        };
+        let api = self.api.clone();
+        cx.spawn(async move |this, cx| {
+            match api
+                .write_voice_interactive_event(
+                    clan_id,
+                    voice_channel_id,
+                    user_id,
+                    0,
```
huylexuan44, line 2803: receiver_id changed from user_id to 0, so one click force-opens a browser on every participant. Confirm broadcast is intended.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -3131,6 +3131,27 @@ fn control_bar(
         })
     };
 
+    let interactive_app_button = {
```
huylexuan44, line 3134: No permission gate, unlike record_button (can_record) and agent_button (PERMISSION_MANAGE_CHANNEL). Any guest can broadcast.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +860,128 @@ impl VoiceStore {
...
+                .active_channel_id()
+                .and_then(|id| id.parse::<i64>().ok())
+                != Some(event.voice_channel_id)
+        {
+            return;
+        }
+        let Some(app) = interactive_app_type(event.event_type) else {
+            return;
+        };
+        self.open_interactive_app_url(app, &event.params, event.clan_id, cx);
```
huylexuan44: params never checked non-empty; an echoed ack opens an empty data= URL, then the realtime push opens a second window.

crates/mezon-client/src/transport.rs
```diff
@@ -78,6 +78,75 @@ struct PromiseExecutor {
...
+) -> Result<Option<realtime::VoiceInteractiveEvent>> {
+    if response.is_empty() {
+        return Ok(None);
+    }
+    if let Ok(envelope) = realtime::Envelope::decode(response)
+        && let Some(realtime::envelope::Message::VoiceInteractiveEvent(event)) = envelope.message
+    {
+        return Ok(Some(event));
+    }
+    if let Ok(event) = realtime::VoiceInteractiveEvent::decode(response) {
```
huylexuan44: VoiceInteractiveEvent::decode succeeds on almost any proto3 payload, so a cid-only ack becomes a bogus event. Return Ok(None).

crates/mezon-client/src/transport.rs
```diff
@@ -78,6 +78,75 @@ struct PromiseExecutor {
...
+    match realtime::Envelope::decode(payload) {
+        Ok(envelope) => Ok(envelope),
+        Err(modern_error) => {
+            let Ok(legacy_envelope) = LegacyVoiceInteractiveEnvelope::decode(payload) else {
+                return Err(modern_error);
+            };
+            let Some(legacy) = legacy_envelope.voice_interactive_event else {
+                return Err(modern_error);
+            };
+            if legacy.event_type != 1 {
```
huylexuan44: event_type != 1 rejects the new app types 10/11/12, so legacy-layout app pushes are dropped. Accept app event types too.

crates/mezon-client/src/transport.rs
```diff
@@ -4458,11 +4527,26 @@ impl MezonTransport {
...
+            code,
+            response_bytes = response.len(),
+            "received VoiceInteractiveEvent CID response"
+        );
         if code != 0 {
-            anyhow::bail!("write_voice_interactive error: code={code}");
+            anyhow::bail!("write_voice_interactive_event error: code={code}");
         }
-        Ok(())
+        decode_voice_interactive_response(&response)
```
huylexuan44, line 4560: code==0 now fails when the ack will not parse, so a successful flower send at voice.rs:2034 can report an error. Ignore unparsable acks.

crates/mezon-store/src/voice.rs
```diff
@@ -215,6 +245,7 @@ pub struct VoiceStore {
     frame_store: Option<Arc<VideoFrameStore>>,
     camera_devices: Vec<CameraDeviceInfo>,
     device_menu: Option<DeviceMenuKind>,
+    last_interactive_launch: Option<(i32, u64, Instant)>,
```
huylexuan44: Single dedup slot: an interleaved event for another app type overwrites it and lets the duplicate through. Key by app type.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +860,128 @@ impl VoiceStore {
...
+            .last_interactive_launch
+            .is_some_and(|(last_type, last_fingerprint, at)| {
+                last_type == app as i32
+                    && last_fingerprint == fingerprint
+                    && now.duration_since(at) < INTERACTIVE_LAUNCH_DEDUP_TTL
+            })
+        {
+            return;
+        }
+        self.last_interactive_launch = Some((app as i32, fingerprint, now));
```
huylexuan44: Fingerprint stored before build_channel_app_url. Move it after a successful build so a failed launch does not burn the TTL.

crates/mezon-store/src/voice.rs
```diff
@@ -62,6 +64,34 @@ const SOUND_REACTION_TAIL: Duration = Duration::from_millis(300);
...
+        | Some(app @ VoiceInteractiveEventType::AppInteractive) => Some(app),
+        _ => None,
+    }
+}
+
+fn interactive_event_targets_user(event_type: i32, receiver_id: i64, user_id: i64) -> bool {
+    interactive_app_type(event_type).is_some() && (receiver_id == 0 || receiver_id == user_id)
+}
+
+fn redact_interactive_app_url(url: &str) -> String {
```
huylexuan44, line 79: Substring match on data= also hits appdata=, leaking the signed payload into logs. Parse the query with the url crate.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +860,128 @@ impl VoiceStore {
...
+            Ok(url) => url,
+            Err(error) => {
+                tracing::warn!("failed to build voice interactive app URL: {error:#}");
+                return;
+            }
+        };
+        tracing::info!(
+            url = %redact_interactive_app_url(&url),
+            params_present = !params.is_empty(),
+            params_bytes = params.len(),
```
huylexuan44: CLAUDE.md forbids logging credential material even as a length. Drop params_bytes and params_present.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -3462,6 +3484,157 @@ fn circle_button(
...
+
+impl InteractiveAppPopoverPanel {
+    fn new(
+        voice: Entity<VoiceStore>,
+        locale: String,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) -> Self {
+        let focus_handle = cx.focus_handle();
+        cx.on_blur(&focus_handle, window, |_, _, cx| cx.emit(DismissEvent))
```
huylexuan44, line 3559: Missing .on_mouse_down_out(DismissEvent). The voice view has no focusable divs, so blur never fires and outside clicks do not close.

crates/mezon-store/src/voice.rs
```diff
@@ -2674,6 +2801,69 @@ impl VoiceStore {
         self.device_menu
     }
 
+    pub fn request_interactive_app(
+        &mut self,
+        app: VoiceInteractiveEventType,
```
huylexuan44: Taking the wire enum lets a caller pass Gift or Recording. Use a dedicated app type so the invariant is unrepresentable.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +860,128 @@ impl VoiceStore {
...
+                event.voice_channel_id,
+                &event.params,
+                joined_channel,
+            ) else {
+                return;
+            };
+            self.show_flower_effect(giver_id, receiver_id, timestamp, cx);
+            return;
+        }
+        let Some(current_user_id) = crate::BadgeService::try_global(cx)
```
huylexuan44: Broadcast events need no user id, but this early return drops them when current_user_id is unavailable. Keep the lookup lazy.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +860,128 @@ impl VoiceStore {
...
+        let url = match build_channel_app_url(
+            base_url,
+            ChannelAppLaunchParams {
+                web_app_data: params,
+                clan_id: &clan_id_string,
+                clan_name,
+            },
+        ) {
+            Ok(url) => url,
+            Err(error) => {
```
huylexuan44: build_channel_app_url returns Ok from both arms, so this Err handler is unreachable. Change its signature to -> String.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +860,128 @@ impl VoiceStore {
...
+        if !interactive_event_targets_user(event.event_type, event.receiver_id, current_user_id)
+            || self
+                .connection
+                .active_channel_id()
+                .and_then(|id| id.parse::<i64>().ok())
+                != Some(event.voice_channel_id)
+        {
+            return;
+        }
+        let Some(app) = interactive_app_type(event.event_type) else {
```
huylexuan44: interactive_app_type already ran inside interactive_event_targets_user, so this else is unreachable, as is the _ => return at 951.

crates/mezon-store/src/voice.rs
```diff
@@ -828,11 +860,128 @@ impl VoiceStore {
             dispatch.on(
                 RealtimeKind::VoiceInteractive,
                 &entity,
-                |this, event, cx| this.handle_voice_interactive(event, cx),
+                |this, event, cx| {
```
huylexuan44: Re-matches the event just to log, duplicating handle_voice_interactive. Also fires info for every gift; use debug.

crates/mezon-store/src/voice.rs
```diff
@@ -2674,6 +2801,69 @@ impl VoiceStore {
         self.device_menu
     }
 
+    pub fn request_interactive_app(
+        &mut self,
+        app: VoiceInteractiveEventType,
+        cx: &mut Context<Self>,
+    ) {
+        let Some((voice_channel_id, clan_id)) = self.connection.connected_channel() else {
+            cx.notify();
```
huylexuan44: cx.notify() is dead here, the function mutates no store state. The branch at 2818 omits it, which makes the intent unreadable.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -3462,6 +3484,157 @@ fn circle_button(
...
+        self.on_click = Some(Box::new(handler));
+        self
+    }
+
+    fn cursor_style(self, _cursor_style: CursorStyle) -> Self {
+        self
+    }
+}
+
+impl RenderOnce for InteractiveAppTrigger {
```
huylexuan44, line 3528: Duplicates circle_button line-for-line. Wrap circle_button and add only the tooltip and click handler.

crates/mezon-ui/src/channel_app/mod.rs
```diff
@@ -1,21 +1,8 @@
-use gpui::{App, AppContext};
+use gpui::App;
 use mezon_store::PlatformStore;
 
 fn open_in_browser(url: String, cx: &App) {
```
huylexuan44: open_in_browser is now a pass-through. Call PlatformStore::open_app_window directly and drop the indirection.

assets/i18n/en.json
```diff
@@ -1075,6 +1075,10 @@
   "channelVoice.muteModal.mute": "Mute",
   "channelVoice.muteModal.title": "Mute member",
   "channelVoice.noDisplay": "No display channel voice",
+  "channelVoice.openInteractiveApp": "Open interactive app",
+  "channelVoice.interactiveApp.quiz": "Quiz",
```
huylexuan44: Breaks the sorted order documented in crates/mezon-i18n/README.md; interactiveApp.* belongs before muteModal.*. Same in 11 locales.

assets/i18n/vi.json
```diff
@@ -1070,6 +1070,10 @@
   "channelVoice.muteModal.mute": "Tắt tiếng",
   "channelVoice.muteModal.title": "Tắt tiếng thành viên",
   "channelVoice.noDisplay": "Không hiển thị kênh thoại",
+  "channelVoice.openInteractiveApp": "Mở ứng dụng tương tác",
+  "channelVoice.interactiveApp.quiz": "Quiz",
```
huylexuan44: Values are English in all 11 locales while openInteractiveApp is translated. Translate these or drop the i18n indirection.

crates/mezon-store/src/voice.rs
```diff
@@ -3227,10 +3418,74 @@ mod tests {
...
+        ));
+        assert!(!interactive_event_targets_user(
+            VoiceInteractiveEventType::Gift as i32,
+            user_id,
+            user_id,
+        ));
+    }
+
+    #[test]
+    fn interactive_app_log_url_redacts_signed_data() {
```
huylexuan44, line 3417: No case with an earlier param ending in data=, and the dedup slot has no test at all. Both are pure functions and cheap to cover.

crates/mezon-store/src/voice.rs
duocphungvanNCC: This is intentional. In the first demo phase, selecting an app should open it for every participant receiving the event.

crates/mezon-ui/src/chat/voice.rs
duocphungvanNCC: This is intentional for the first demo phase. The button is available to everyone, so no permission check is yet.

PR #358, opened 2026-08-18
Fix26c/desktop/create thread with mentions

giachuvan:


crates/mezon-store/src/threads.rs
```diff
@@ -964,7 +969,13 @@ impl ThreadsStore {
         cx.notify();
     }
 
-    pub fn submit_create(&mut self, name: String, message: String, cx: &mut Context<Self>) {
+    pub fn submit_create(
+        &mut self,
+        name: String,
+        message: String,
+        content_tokens: OutgoingContent,
```
huylexuan44, line 975: Line 983 message.trim() shifts these s/e offsets - they index the untrimmed raw. Use trim_end or rebase them.

crates/mezon-store/src/threads.rs
```diff
@@ -1119,6 +1170,67 @@ fn thread_create_fail_reason(err: &anyhow::Error) -> ThreadCreateFailReason {
...
+    mentioned: &[UserId],
+    parent_members: Option<&[UserId]>,
+    self_id: Option<UserId>,
+) -> Vec<UserId> {
+    let mut invitees = Vec::new();
+    for user_id in mentioned {
+        if Some(*user_id) == self_id || invitees.contains(user_id) {
+            continue;
+        }
+        if parent_members.is_some_and(|members| !members.contains(user_id)) {
```
huylexuan44: Fail-open: an unknown parent roster skips the gate, so @role can add non-members into a thread under a private channel.

crates/mezon-store/src/threads.rs
```diff
@@ -1056,16 +1088,42 @@ impl ThreadsStore {
...
+            {
+                tracing::warn!("join_chat after thread create failed: {e}");
+            }
+
+            if !invite_ids.is_empty() {
+                let user_ids: Vec<String> = invite_ids
+                    .iter()
+                    .map(|user_id| user_id.to_string())
+                    .collect();
+                if let Err(e) = api.add_channel_users(thread_id, user_ids).await {
```
huylexuan44, line 1155: add_channel_users failure is only logged - the thread still reports success and the starter message still pings those users.

crates/mezon-store/src/threads.rs
```diff
@@ -1056,16 +1088,42 @@ impl ThreadsStore {
...
+                    tracing::error!("add mentioned users to new thread failed: {e}");
+                } else {
+                    let _ = this.update(cx, |_this, cx| {
+                        if let Some(members) = ChannelMembersStore::try_global(cx) {
+                            members.update(cx, |members, cx| {
+                                members.apply_members_added(ChannelId(thread_id), &invite_ids, cx);
```
huylexuan44: apply_members_added returns early when the channel has no cache bucket, and a just-created thread never has one.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -782,6 +791,16 @@ impl MentionInput {
         cx.notify();
     }
 
+    pub fn current_content(&self, cx: &App) -> (String, OutgoingContent) {
```
huylexuan44: current_content ignores pending_attachments, so images pasted into the create-thread box are silently dropped.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -2235,8 +2240,18 @@ impl ChatLayout {
...
             let ph = mezon_i18n::t(&locale, "chat.messagePlaceholder");
-            self.create_thread_message_input =
-                Some(cx.new(|cx| InputState::new(window, cx).placeholder(ph).embedded(true)));
+            let settings = self.settings.clone();
+            let input = cx.new(|cx| MentionInput::new_compact(ph, settings, window, cx));
+            self._create_thread_mention_sub = Some(cx.subscribe_in(
+                &input,
+                window,
+                |this, _, event: &MentionInputEvent, window, cx| {
+                    if matches!(event, MentionInputEvent::Submit) {
```
huylexuan44: MentionInputEvent::Cancel is dropped, so Escape no longer closes the panel. Handle it like create_topic_panel.rs:88.

crates/mezon-store/src/threads.rs
```diff
@@ -1119,6 +1170,67 @@ fn thread_create_fail_reason(err: &anyhow::Error) -> ThreadCreateFailReason {
     }
 }
 
+fn mention_invite_candidates(
```
huylexuan44: Duplicates mentioned_thread_candidates and plan_thread_membership in messages.rs - make those pub(crate) and reuse.

crates/mezon-ui/src/chat/create_thread_panel.rs
```diff
@@ -257,12 +256,7 @@ pub fn render_create_thread_panel(params: CreateThreadPanelParams<'_>) -> AnyEle
                         .px_3()
                         .py_2()
                         .min_h(px(45.))
```
huylexuan44: Wrapper keeps px_3/py_2/min_h(45) while compact MentionInput adds its own px(10)/py(9)/min_h(40) - double padding.

crates/mezon-ui/src/chat/create_thread_panel.rs
```diff
@@ -257,12 +256,7 @@ pub fn render_create_thread_panel(params: CreateThreadPanelParams<'_>) -> AnyEle
...
-                        .child(
-                            Input::new(&message_input)
-                                .w_full()
-                                .text_sm()
-                                .text_color(tokens.text_theme_message),
-                        ),
+                        .child(message_input.clone()),
```
huylexuan44, line 256: Dropping Input also drops .text_sm() and text_theme_message; the field now renders 16px with text_primary.

crates/mezon-store/src/threads.rs
```diff
@@ -990,6 +1001,27 @@ impl ThreadsStore {
         };
         let category_id = self.category_id.clone();
         let channel_private = self.create_private;
+        let clan_id_parsed = clan_id.parse::<ClanId>().unwrap_or(ClanId(0));
```
huylexuan44: unwrap_or(ClanId(0)) silently drops every role mention; the same parse is repeated at line 1148.

crates/mezon-store/src/threads.rs
```diff
@@ -1119,6 +1170,67 @@ fn thread_create_fail_reason(err: &anyhow::Error) -> ThreadCreateFailReason {
...
+}
+
+fn create_thread_invitees(
+    mentioned: &[UserId],
+    parent_members: Option<&[UserId]>,
+    self_id: Option<UserId>,
+) -> Vec<UserId> {
+    let mut invitees = Vec::new();
+    for user_id in mentioned {
+        if Some(*user_id) == self_id || invitees.contains(user_id) {
```
huylexuan44: invitees.contains() never fires - mention_invite_candidates already deduped. Use a HashSet for the parent lookup too.

PR #373, opened 2026-08-19
Fix vietnamese for X11, Wayland with ibus, fcitx5 and bamboo, unikey

giachuvan:


crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+                ImEvent::ForwardKey {
+                    keyval,
+                    state,
+                    is_release,
+                } => {
+                    let modifiers = modifiers_from_state(xproto::KeyButMask::from(state as u16));
+                    let keysym = xkbc::Keysym::new(keyval);
+                    let keystroke = Keystroke {
+                        modifiers,
+                        key: xkbc::keysym_get_name(keysym),
```
huylexuan44: keysym_get_name yields Return/BackSpace/Escape/Left, but every GPUI binding uses the lowercase names from keystroke_from_xkb (Keysym::Return maps to enter). Any key the IME forwards back matches no binding, so Enter never sends and Escape never closes.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+
+    fn apply_im_events(&self, window: &X11WindowStatePtr, events: Vec<ImEvent>) {
+        for event in fold_im_events(events) {
+            match event {
+                ImEvent::Commit(text) => {
+                    self.0.borrow_mut().composing = false;
+                    window.handle_ime_commit(text);
+                    self.sync_surrounding_after_commit(window);
+                }
+                ImEvent::HidePreedit => {
```
huylexuan44, line 1381: handle_ime_unmark only clears marked_range, it does not remove the preedit text. parse_fcitx5_preedit maps an empty preedit to HidePreedit and never to Preedit with empty text, so the handle_ime_delete branch is unreachable for fcitx5 and the preedit stays in the buffer after Backspace or Escape.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+            && let Some(window) = window.as_ref()
+            && let Some(surrounding) = window.get_ime_surrounding()
+        {
+            im.set_surrounding(&surrounding);
+        }
+        self.0.borrow_mut().im = Some(im);
+        true
+    }
+
+    fn sync_surrounding_after_commit(&self, window: &X11WindowStatePtr) {
```
huylexuan44: This borrow_mut is held across get_ime_surrounding, which runs a full App::update and flush_effects. Anything re-entering X11Client there panics with already mutably borrowed. sync_dbus_im and apply_im_events call get_ime_surrounding before re-borrowing - do the same here.

crates/mezon-widgets/src/input.rs
```diff
@@ -1126,6 +1133,41 @@ impl EntityInputHandler for InputState {
     ) -> Option<usize> {
         Some(self.offset_to_utf16(self.index_for_mouse_position(point)))
     }
+
+    fn surrounding_text(
+        &mut self,
+        _window: &mut Window,
+        _cx: &mut Context<Self>,
+    ) -> Option<ImeSurroundingText> {
+        Some(ImeSurroundingText::from_selection(
```
huylexuan44, line 1169: This returns self.content even when is_masked() is true, so the login password is shipped to ibus/fcitx5 over D-Bus and to the Wayland compositor. The same file already blocks copy on masked content. Return None when masked.

crates/mezon-ui/src/chat/mention_input/text_field.rs
```diff
@@ -1184,6 +1190,41 @@ impl EntityInputHandler for MentionInputState {
         self.last_bounds?;
         Some(self.offset_to_utf16(self.index_for_mouse_position(point)))
     }
+
+    fn surrounding_text(
+        &mut self,
+        _window: &mut Window,
+        _cx: &mut Context<Self>,
+    ) -> Option<ImeSurroundingText> {
+        Some(ImeSurroundingText::from_selection(
```
huylexuan44, line 1194: Same masked-content leak as input.rs: return None when is_masked() so password fields never reach the IME.

crates/mezon-ui/src/components/primitives/textarea.rs
```diff
@@ -990,6 +996,41 @@ impl EntityInputHandler for TextArea {
         self.last_bounds?;
         Some(self.offset_to_utf16(self.index_for_mouse_position(point)))
     }
+
+    fn surrounding_text(
+        &mut self,
+        _window: &mut Window,
+        _cx: &mut Context<Self>,
+    ) -> Option<ImeSurroundingText> {
+        Some(ImeSurroundingText::from_selection(
```
huylexuan44, line 1028: Same masked-content leak: return None when is_masked().

crates/mezon-app/src/main.rs
```diff
@@ -51,30 +51,60 @@ impl tracing_tracy::Config for TracyConfig {
...
 fn configure_linux_session() {
     if mezon_store::record_wayland_session() {
         unsafe { std::env::set_var("MEZON_WAYLAND_SESSION", "1") };
     }
+    ensure_linux_utf8_ctype();
+    if prefers_native_wayland(
+        std::env::var("WAYLAND_DISPLAY").ok().as_deref(),
+        linux_session_override().as_deref(),
+    ) {
+        return;
```
huylexuan44, line 81: This early return skips the GDK_BACKEND=x11 block below, flipping every Linux user with WAYLAND_DISPLAY from XWayland to native Wayland. That changes renderer, decorations, scaling, scap capture, tray, clipboard and DnD - none of it IME. Split it out or keep it opt-in for one release.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+                    if !self.reconnect_dbus_im() {
+                        self.revive_xim_if_dead();
+                    }
+                }
+                false
+            }
+        }
+    }
+
+    fn selection_owns_editing_key(&self, window_id: u32, keyval: u32) -> bool {
```
huylexuan44, line 1206: Backspace/Delete bypass the IME here but nothing re-syncs surrounding text afterwards, so the engine keeps the pre-delete document and the next tone key rewrites the wrong character. navigation_skips_ime calls refresh_dbus_im_for_field for exactly this reason.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+                    key.event,
+                )
+            }
+            _ => return false,
+        };
+        if !self.0.borrow().dbus_im_focused && self.sync_dbus_im(None, true, true) {
+            self.0.borrow_mut().dbus_im_focused = true;
+            self.drain_dbus_im();
+        }
+        if shortcut_skips_ime(state, keyval) {
```
huylexuan44: This clears composing and hands the key to the app without resetting the engine or clearing the widget marked_range, so Ctrl+C during composition leaves a live preedit plus a stale marked range. select_all was patched for this class of bug - the general shortcut path was not.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -281,8 +385,12 @@ impl X11ClientStatePtr {
         let Some(client) = self.get_client() else {
             return;
         };
+        if client.sync_dbus_im(Some(bounds), true, true) {
+            client.drain_dbus_im();
+            return;
+        }
         let mut state = client.0.borrow_mut();
-        if state.composing || state.ximc.is_none() {
+        if state.ximc.is_none() {
```
huylexuan44: The state.composing guard was dropped here, so caret-bounds updates now push set_ic_values mid-composition on the XIM fallback path. XIM servers that treat that as a state change drop the in-flight preedit.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -130,6 +133,103 @@ enum RefreshState {
...
+        ImEvent::ForwardKey {
+            keyval: 32,
+            is_release: false,
+            ..
+        } => true,
+        _ => false,
+    }
+}
+
+fn maybe_append_fcitx_space(keyval: u32, events: &mut Vec<ImEvent>) {
```
huylexuan44, line 237: This injects a space whenever fcitx5 filters the space key and emits any commit/preedit. With CJK candidate selection (Commit of the candidate, then empty preedit) it appends a spurious trailing space to every committed word.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+                    return false;
+                }
+                let code = key.detail.into();
+                let keysym = client_state.xkb.key_get_one_sym(code);
+                if keysym.is_modifier_key() {
+                    return false;
+                }
+                (
+                    keysym.raw(),
+                    u32::from(key.detail),
```
huylexuan44, line 1105: IBus clients must pass hardware_keycode minus 8 (evdev), not the raw X11 keycode; fcitx5 on the same path expects the raw keycode, so one of the two backends gets a keycode shifted by 8.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+    fn dbus_reconnect_ready(&self) -> bool {
+        const COOLDOWN: Duration = Duration::from_secs(2);
+        let state = self.0.borrow();
+        match state.last_dbus_attempt {
+            Some(at) if at.elapsed() < COOLDOWN => false,
+            _ => true,
+        }
+    }
+
+    fn ensure_dbus_im(&self) -> bool {
```
huylexuan44, line 1248: This retries connect() from the key path every 2s forever: two fresh session-bus channels plus blocking NameHasOwner calls, machine-id and ibus bus files, all inside process_x11_events. Users without ibus/fcitx5 get a typing hitch every 2s with no backoff.

crates/vendor/gpui/src/platform.rs
```diff
@@ -1374,6 +1374,191 @@ impl PlatformInputHandler {
...
+        if text.len() <= Self::MAX_LEN {
+            return Self {
+                text,
+                cursor,
+                anchor,
+            };
+        }
+        let half = Self::MAX_LEN / 2;
+        let mut start = floor_char_boundary(&text, cursor.saturating_sub(half));
+        let mut end =
```
huylexuan44, line 1522: ceil_char_boundary rounds end up past start+MAX_LEN, and the deficit branch below floors start further back, so this can return 4001-4003 bytes. zwp_text_input_v3.set_surrounding_text caps at 4000. The while loop below is dead code since ceil_char_boundary already returns a boundary. Only multi-byte text hits this - the test uses ASCII.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1152 @@
...
+            .with_proxy(&self.dest, self.path.clone(), KEY_TIMEOUT)
+    }
+}
+
+impl Drop for X11ImContext {
+    fn drop(&mut self) {
+        for token in self.tokens.drain(..) {
+            self.conn.stop_receive(token);
+        }
+        let _ = self.call_void("DestroyIC");
```
huylexuan44: IBus has no DestroyIC on org.freedesktop.IBus.InputContext (it is Destroy on org.freedesktop.IBus.Service), the error is discarded, so every reconnect leaks an input context in ibus-daemon. This is also a blocking 80ms D-Bus call in Drop.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1152 @@
...
+                self.process_fcitx5_single(keyval, keycode, state, is_release, time)
+            }
+            FcitxKeyMode::Unknown => {
+                match self.process_fcitx5_batch(keyval, keycode, state, is_release, time) {
+                    Ok(handled) => {
+                        self.fcitx_key_mode.set(FcitxKeyMode::Batch);
+                        Ok(handled)
+                    }
+                    Err(_) => {
+                        self.fcitx_key_mode.set(FcitxKeyMode::Single);
```
huylexuan44, line 345: Any Err falls back to Single permanently, including a transient KEY_TIMEOUT on a cold engine. That loses the batch ordering this PR relies on, with no retry. Only fall back on UnknownMethod/UnknownInterface.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1152 @@
...
+            IbusSurroundingEnc::Unknown => {
+                match self.call_ibus_surrounding_text(text, cursor, anchor) {
+                    Ok(()) => {
+                        self.ibus_surrounding.set(IbusSurroundingEnc::IBusText);
+                        Ok(())
+                    }
+                    Err(_) => {
+                        let result = self.call_ibus_surrounding_plain(text, cursor, anchor);
+                        if result.is_ok() {
+                            self.ibus_surrounding.set(IbusSurroundingEnc::PlainVariant);
```
huylexuan44, line 443: Same permanent-downgrade-on-any-error issue: a timeout on the first call locks the wrong surrounding-text encoding for the session.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1152 @@
...
+            .method_call(
+                FCITX5_IC_IFACE,
+                "ProcessKeyEvent",
+                (keyval, keycode, state, is_release, time),
+            )
+            .map(|(handled,): (bool,)| handled)
+            .map_err(|error| error.to_string())
+    }
+
+    pub(crate) fn is_dead(&self) -> bool {
```
huylexuan44, line 400: fail_count is only bumped by process_key; set_surrounding, set_cursor_rect, focus_in/out and reset all discard their errors, so an engine that dies while the user is not typing never triggers reconnect.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1152 @@
...
+use dbus::blocking::{Connection, Proxy};
+use dbus::channel::{BusType, Channel, MatchingReceiver, Token};
+use dbus::message::MatchRule;
+use dbus::{Message, Path as DbusPath};
+use gpui::ImeSurroundingText;
+
+const CONNECT_TIMEOUT: Duration = Duration::from_millis(200);
+const KEY_TIMEOUT: Duration = Duration::from_millis(80);
+const SIGNAL_WAIT: Duration = Duration::from_millis(16);
+const COMMIT_WAIT: Duration = Duration::from_millis(40);
```
huylexuan44, line 18: Worst case per keystroke is 80ms KEY_TIMEOUT plus 16ms SIGNAL_WAIT plus 40ms COMMIT_WAIT on the X11 event loop. COMMIT_WAIT never completes when the engine sends no follow-up text (Backspace/Escape), so it is paid in full.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1152 @@
...
+fn parse_fcitx5_commit(message: &Message) -> Option<ImEvent> {
+    let (text,): (String,) = message.read1().ok()?;
+    Some(ImEvent::Commit(text))
+}
+
+fn parse_fcitx5_preedit(message: &Message) -> Option<ImEvent> {
+    let (chunks, caret): (Vec<(String, i32)>, i32) = message.read2().ok()?;
+    let text: String = chunks.into_iter().map(|(piece, _)| piece).collect();
+    if text.is_empty() {
+        return Some(ImEvent::HidePreedit);
```
huylexuan44, line 929: Empty preedit always becomes HidePreedit, never Preedit with empty text, which is why apply_im_events only unmarks instead of deleting. Emit the empty-preedit variant or make HidePreedit delete.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1152 @@
...
+    }
+}
+
+impl X11ImContext {
+    pub(crate) fn connect() -> Option<Self> {
+        for kind in connect_order(&preferred_backends()) {
+            match kind {
+                ImKind::Fcitx5 => match connect_fcitx5() {
+                    Ok(ctx) => return Some(ctx),
+                    Err(_) => {}
```
huylexuan44, line 125: This 1152-line subsystem emits no diagnostics at all while the PR also deletes the six existing [xim] eprintln lines. CLAUDE.md requires eprintln for vendor diagnostics since log is dropped. A user reporting broken Vietnamese input leaves no trace of which step failed.

crates/vendor/gpui_linux/src/linux/wayland/client.rs
```diff
@@ -1770,36 +1841,19 @@ impl Dispatch<zwp_text_input_v3::ZwpTextInputV3, ()> for WaylandClientStatePtr {
...
-                    }
                 } else {
-                    state.composing = false;
-                    drop(state);
-                    window.handle_ime(ImeInput::DeleteText);
+                    client.borrow_mut().composing = false;
+                }
+
+                sync_text_input(text_input, &window, ChangeCause::InputMethod);
+                if last_serial == serial {
```
huylexuan44: sync_text_input queues set_surrounding_text but commit only runs when last_serial equals serial, so the IME sees one-keystroke-stale surrounding text on alternate done events. The composing branch of update_ime_position also queues a set_cursor_rectangle it never commits.

crates/vendor/gpui_linux/src/linux/wayland/client.rs
```diff
@@ -409,19 +425,34 @@ impl WaylandClientStatePtr {
 
     pub fn update_ime_position(&self, bounds: Bounds<Pixels>) {
         let client = self.get_client();
-        let state = client.borrow_mut();
-        if state.composing || state.text_input.is_none() || state.pre_edit_text.is_some() {
+        let mut state = client.borrow_mut();
+        let Some(text_input) = state.text_input.take() else {
```
huylexuan44: text_input is taken out of state while sync_text_input re-enters GPUI via get_ime_surrounding. A nested disable_ime would see None, skip destroy, and then this restores the stale text_input - leaving the IME enabled on a field that no longer accepts input.

crates/mezon-canvas/src/editor.rs
```diff
@@ -1766,6 +1771,40 @@ impl EntityInputHandler for CanvasEditorState {
...
+            &self.selected_range,
+            self.marked_range.as_ref(),
+            self.selection_reversed,
+            before_len,
+            after_len,
+        );
+        if range.is_empty() {
+            return;
+        }
+        self.replace_text_in_range(range, "", cx);
```
huylexuan44, line 1830: delete_surrounding_text here calls the inherent replace_text_in_range, which never clears marked_range, so stale preedit byte offsets survive into a shrunk buffer. The other three widgets call the trait method, which ends with marked_range.take().

crates/mezon-widgets/src/text_edit.rs
```diff
@@ -3,61 +3,53 @@ use std::ops::Range;
...
+use unicode_segmentation::UnicodeSegmentation;
 
 pub const MAX_UNDO_HISTORY: usize = 256;
 
-#[derive(Clone, Copy, PartialEq, Eq)]
-enum CharKind {
-    Whitespace,
-    Word,
-    Punctuation,
+fn word_bound_segments(text: &str) -> impl Iterator<Item = (usize, usize)> + '_ {
```
huylexuan44: Switching to UAX#29 makes foo.bar a single word, so double-click and ctrl-arrow no longer split on the dot. Browsers and textarea - the React reference - do split. This also forced existing tests to be rewritten, and it is unrelated to the IME fix.

crates/mezon-widgets/src/text_edit.rs
```diff
@@ -93,32 +85,26 @@ pub fn home_target(text: &str, offset: usize) -> usize {
 
 pub fn word_range_at(text: &str, offset: usize) -> Range<usize> {
```
huylexuan44, line 94: This rescans split_word_bound_indices from byte 0 on every call instead of scanning back from the offset. word_range_at never breaks at all, so it is O(document) - shared with mezon-canvas editor, which holds long documents.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1046,373 @@ impl X11Client {
...
+                            is_held: false,
+                            prefer_character_input: false,
+                        }));
+                    }
+                }
+            }
+        }
+        if let Some(surrounding) = window.get_ime_surrounding() {
+            if let Some(im) = self.0.borrow_mut().im.as_mut() {
+                im.set_surrounding(&surrounding);
```
huylexuan44, line 1440: set_surrounding is issued twice per commit: once via sync_surrounding_after_commit inside the Commit arm and again in this trailing block. Both are blocking D-Bus calls.

crates/mezon-ui/src/chat/mention_input/text_field.rs
```diff
@@ -1058,11 +1060,15 @@ impl EntityInputHandler for MentionInputState {
...
             }
         }
-        let range = range_utf16
-            .as_ref()
-            .map(|range_utf16| self.range_from_utf16(range_utf16))
-            .or(self.marked_range.clone())
-            .unwrap_or(self.selected_range.clone());
+        let range = if let Some(range_utf16) = range_utf16.as_ref() {
+            self.range_from_utf16(range_utf16)
+        } else if self.selected_range.start != self.selected_range.end {
```
huylexuan44: Preferring a non-empty selected_range over marked_range means an IME that highlights a sub-range inside its own preedit gets that sub-range replaced instead of the whole preedit.

crates/vendor/gpui/src/platform.rs
```diff
@@ -1374,6 +1374,191 @@ impl PlatformInputHandler {
...
+    pub text: String,
+    /// Caret offset in UTF-8 bytes.
+    pub cursor: usize,
+    /// Selection anchor offset in UTF-8 bytes. Equal to `cursor` when there is no selection.
+    pub anchor: usize,
+}
+
+impl ImeSurroundingText {
+    /// Wayland text-input-v3 surrounding-text size limit.
+    pub const MAX_LEN: usize = 4000;
```
huylexuan44, line 1413: floor_char_boundary and ceil_char_boundary are duplicated verbatim in mezon-widgets/src/text_edit.rs. Also CLAUDE.md forbids doc comments in generated code - these /// blocks should go.

crates/mezon-widgets/src/input.rs
```diff
@@ -989,11 +992,15 @@ impl EntityInputHandler for InputState {
...
         cx: &mut Context<Self>,
     ) {
-        let range = range_utf16
-            .as_ref()
-            .map(|range_utf16| self.range_from_utf16(range_utf16))
-            .or(self.marked_range.clone())
-            .unwrap_or(self.selected_range.clone());
+        let range = if let Some(range_utf16) = range_utf16.as_ref() {
+            self.range_from_utf16(range_utf16)
+        } else if self.selected_range.start != self.selected_range.end {
```
huylexuan44: macOS CJK: multi-clause commit now replaces only the active clause because a non-empty selection beats marked_range; exempt selections inside marked_range (same in textarea.rs, mention text_field.rs, canvas editor.rs).

crates/mezon-app/src/main.rs
```diff
@@ -51,30 +51,60 @@ impl tracing_tracy::Config for TracyConfig {
...
+    std::env::var("MEZON_LINUX_SESSION").ok()
+}
+
 #[cfg(target_os = "linux")]
 fn configure_linux_session() {
     if mezon_store::record_wayland_session() {
         unsafe { std::env::set_var("MEZON_WAYLAND_SESSION", "1") };
     }
+    ensure_linux_utf8_ctype();
+    if prefers_native_wayland(
```
huylexuan44, line 77: Native-Wayland default breaks ibus Vietnamese on wlroots (no text-input-v3 bridge) and the new ibus D-Bus client is X11-only — regression for the exact setup this PR targets.

crates/mezon-app/src/main.rs
```diff
@@ -51,30 +51,60 @@ impl tracing_tracy::Config for TracyConfig {
...
+        std::env::var("WAYLAND_DISPLAY").ok().as_deref(),
+        linux_session_override().as_deref(),
+    ) {
+        return;
+    }
     if std::env::var_os("DISPLAY").is_none() {
```
huylexuan44, line 83: Early return on missing DISPLAY skips remove_var(WAYLAND_DISPLAY), so MEZON_LINUX_SESSION=x11 is silently ignored on Wayland-only sessions.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -130,6 +134,104 @@ enum RefreshState {
...
+        ImEvent::ForwardKey {
+            keyval: 32,
+            is_release: false,
+            ..
+        } => true,
+        _ => false,
+    }
+}
+
+fn maybe_append_fcitx_space(keyval: u32, events: &mut Vec<ImEvent>) {
```
huylexuan44, line 237: Injected Commit(" ") adds a stray space after fcitx5-pinyin candidate selection and wipes fcitx5-mozc's preedit mid-convert; the heuristic can't distinguish unikey from candidate-selection engines.

crates/mezon-ui/src/chat/mention_input/text_field.rs
```diff
@@ -1058,11 +1060,15 @@ impl EntityInputHandler for MentionInputState {
                 self.discard_ime_commit = None;
```
huylexuan44: discard_ime_commit is never consumed by the D-Bus/Wayland frontends (no commit echo after Reset), so the next identically-typed word is silently swallowed.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1049,357 @@ impl X11Client {
...
+                };
+                if kind == super::im_frontend::ImKind::Fcitx5 && !is_release && filtered {
+                    maybe_append_fcitx_space(keyval, &mut events);
+                }
+                if !events.is_empty() {
+                    self.apply_im_events(&window, events);
+                }
+                filtered
+            }
+            Err(_) => {
```
huylexuan44, line 1194: IM reply slower than KEY_TIMEOUT inserts the raw key locally, then the daemon's late Commit/UpdatePreedit still applies — characters double. Correlate or drop late events after a timeout.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1179 @@
...
+    }
+
+    pub(crate) fn take_events(&self) -> Vec<ImEvent> {
+        self.events
+            .lock()
+            .map(|mut events| std::mem::take(&mut *events))
+            .unwrap_or_default()
+    }
+
+    pub(crate) fn process_key(
```
huylexuan44, line 244: Synchronous D-Bus per keypress blocks the UI thread up to ~136ms (KEY_TIMEOUT + signal/commit waits), even for plain English with ibus-daemon running. Move IM I/O off the foreground thread.

crates/vendor/gpui/src/window.rs
```diff
@@ -2708,6 +2708,7 @@ impl Window {
             self.focus_listeners
                 .clone()
                 .retain(&(), |listener| listener(&event, self, cx));
+            self.invalidate_character_coordinates();
```
huylexuan44, line 2711: invalidate_character_coordinates runs update_ime_position with the input handler taken, so the new set_surrounding sync is a no-op on X11/Wayland and X11 drain_dbus_im can drop a queued commit.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1049,357 @@ impl X11Client {
...
+        let scale = state.scale_factor;
+        drop(state);
+        if take_focus && !composing {
+            im.focus_in();
+            im.process_io();
+        }
+        let window = window_id.and_then(|id| self.get_window(id));
+        if let Some(bounds) = bounds {
+            let scaled = bounds.scale(scale);
+            im.set_cursor_rect(
```
huylexuan44: SetCursorRect/SetCursorLocation expect root/screen coordinates; window-local bounds place the candidate popup away from the caret. Translate via get_root_position().

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1049,357 @@ impl X11Client {
...
+                if self.0.borrow().im.is_none() && !self.ensure_dbus_im() {
+                    return false;
+                }
+                let mut client_state = self.0.borrow_mut();
+                if client_state.im.is_none() {
+                    return false;
+                }
+                let code = key.detail.into();
+                let keysym = client_state.xkb.key_get_one_sym(code);
+                if keysym.is_modifier_key() {
```
huylexuan44: Modifier press/release never reaches ProcessKeyEvent, so fcitx5's Left-Shift VI/EN toggle and ibus-bamboo Shift switching don't work in-app. Forward modifier keys.

crates/vendor/gpui_linux/src/linux/wayland/client.rs
```diff
@@ -1770,36 +1840,19 @@ impl Dispatch<zwp_text_input_v3::ZwpTextInputV3, ()> for WaylandClientStatePtr {
...
-                    }
                 } else {
-                    state.composing = false;
-                    drop(state);
-                    window.handle_ime(ImeInput::DeleteText);
+                    client.borrow_mut().composing = false;
+                }
+
+                sync_text_input(text_input, &window, ChangeCause::InputMethod);
+                if last_serial == serial {
```
huylexuan44: done.serial equals the client's commit count, not the previous done serial; this gate skips the ack commit every other done and the surrounding sync later flushes under ChangeCause::Other, desyncing fcitx5-unikey.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -1107,6 +1606,10 @@ impl X11Client {
                 }
                 state.pre_edit_text.take();
```
huylexuan44, line 1653: FCITX5_CAPS declares ClientUnfocusCommit but FocusOut deletes the preedit and drops late commits — alt-tab destroys the in-progress word. Commit the marked text instead of deleting it.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1049,357 @@ impl X11Client {
...
+    }
+
+    fn apply_im_events(&self, window: &X11WindowStatePtr, events: Vec<ImEvent>) {
+        for event in fold_im_events(events) {
+            match event {
+                ImEvent::Commit(text) => {
+                    self.0.borrow_mut().composing = false;
+                    window.handle_ime_commit(text);
+                }
+                ImEvent::HidePreedit => {
```
huylexuan44, line 1381: HidePreedit maps to unmark (keeps text), so fcitx5 Escape-cancel commits the canceled composition; fold_im_events also drops HidePreedit whenever the batch has text.

crates/vendor/gpui_linux/src/linux/x11/xim_handler.rs
```diff
@@ -213,22 +267,148 @@ impl<C: Client<XEvent = xproto::KeyPressEvent>> ClientHandler<C> for XimHandler
...
+        _client: &mut C,
+        _input_method_id: u16,
+        _input_context_id: u16,
+    ) -> Result<(), ClientError> {
+        self.clear_preedit();
+        self.emit_preedit(String::new(), 0);
+        Ok(())
+    }
+
+    fn handle_reset_ic(
```
huylexuan44, line 335: Non-empty XIM_RESET_IC reply is turned into a commit after unmark already kept the text — click-away double-inserts the word. Discard the reply's preedit instead.

crates/vendor/gpui_linux/src/linux/x11/xim_handler.rs
```diff
@@ -103,13 +132,38 @@ impl XimHandler {
...
+    fn emit_preedit(&mut self, text: String, caret: i32) {
+        let caret = caret_utf16_range(&text, caret);
+        self.push_callback(XimCallbackEvent::XimPreeditEvent(self.window, text, caret));
+    }
+
+    fn clear_preedit(&mut self) {
+        self.preedit.clear();
+    }
+
+    pub fn should_skip_apply(&self, text: &str) -> bool {
```
huylexuan44: last_applied is also set on commit and never invalidated, so re-composing the same word renders blind and caret-only preedit updates are suppressed.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1179 @@
...
+        time: u32,
+    ) -> Result<bool, String> {
+        let result = match self.kind {
+            ImKind::Fcitx5 => self.process_fcitx5_key(keyval, keycode, state, is_release, time),
+            ImKind::IBus => self
+                .proxy()
+                .method_call(
+                    IBUS_IC_IFACE,
+                    "ProcessKeyEvent",
+                    (keyval, keycode, ibus_key_state(state, is_release)),
```
huylexuan44: IBus expects evdev keycodes: pass keycode - 8 in the ibus branch (reference ibusimcontext.c does this; the fcitx5 branch is correct as-is).

crates/mezon-ui/src/chat/mention_input/text_field.rs
```diff
@@ -1031,11 +1033,6 @@ impl EntityInputHandler for MentionInputState {
     }
 
     fn unmark_text(&mut self, _window: &mut Window, _cx: &mut Context<Self>) {
-        #[cfg(target_os = "linux")]
-        if let Some(marked) = self.marked_range.clone() {
-            let marked = self.clamp_range(marked);
-            self.discard_ime_commit = self.content.get(marked).map(str::to_string);
-        }
         self.marked_range = None;
```
huylexuan44, line 1045: Guard removal regresses D-Bus path: click while composing resets the engine and its commit echo re-inserts the preedit.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1049,354 @@ impl X11Client {
...
+            if !is_release {
+                self.0.borrow_mut().composing = false;
+                if let Some(window) = self.get_window(window_id) {
+                    window.handle_ime_unmark();
+                }
+            }
+            return false;
+        }
+        let composing = self.0.borrow().composing;
+        if is_release && !composing && !is_modifier {
```
huylexuan44, line 1156: Modifier releases carry their state bit, so shortcut_skips_ime eats them; IM sees Ctrl/Alt/Super press with no release.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1197 @@
...
+        if let Ok(mut events) = self.events.lock() {
+            events.extend(incoming);
+        }
+    }
+
+    pub(crate) fn take_events(&self) -> Vec<ImEvent> {
+        self.events
+            .lock()
+            .map(|mut events| {
+                if self.drop_late.get() {
```
huylexuan44: drop_late latches until the next keystroke, so async commits like a mouse candidate pick are discarded after one timeout.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -1107,15 +1603,20 @@ impl X11Client {
...
+                    im.focus_out();
+                }
                 if let Some((mut ximc, xim_handler)) = state.take_xim() {
                     if xim_handler.connected && xim_handler.ic_id != 0 {
                         let _ = ximc.unset_focus(xim_handler.im_id, xim_handler.ic_id);
                     }
                     state.restore_xim(ximc, xim_handler);
                 }
                 drop(state);
+                self.drain_dbus_im();
```
huylexuan44, line 1666: keyboard_focused_window is already None here, so this drain discards pending events, losing any unfocus commit.

crates/mezon-widgets/src/text_edit.rs
```diff
@@ -189,6 +189,105 @@ pub fn should_coalesce(last: Option<EditKind>, kind: EditKind) -> bool {
...
+    while i < text.len() {
+        if text.is_char_boundary(i) {
+            return i;
+        }
+        i += 1;
+    }
+    text.len()
+}
+
+pub fn ime_replace_range(selected: &Range<usize>, marked: Option<&Range<usize>>) -> Range<usize> {
```
huylexuan44, line 222: Non-empty selection outside marked wins here; upstream replaces marked, so an IM commit leaves the preedit text behind.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1049,354 @@ impl X11Client {
...
+                let code = key.detail.into();
+                let keysym = client_state.xkb.key_get_one_sym(code);
+                (
+                    keysym.raw(),
+                    u32::from(key.detail),
+                    u32::from(key.state),
+                    false,
+                    key.time,
+                    key.event,
+                    keysym.is_modifier_key(),
```
huylexuan44, line 1110: Every modifier edge now blocks on process_key (80ms timeout); a filtered Shift press also burns the 16ms signal wait.

crates/vendor/gpui_linux/src/linux/x11/window.rs
```diff
@@ -1269,6 +1311,51 @@ impl X11WindowStatePtr {
...
+        x: f32,
+        y: f32,
+        width: f32,
+        height: f32,
+    ) -> Option<(i32, i32, i32, i32)> {
+        let root = self.state.borrow().x_root_window;
+        let reply = get_reply(
+            || "X11 TranslateCoordinates for IME cursor failed.",
+            self.xcb
+                .translate_coordinates(self.x_window, root, x.round() as i16, y.round() as i16),
```
huylexuan44: Blocking TranslateCoordinates round trip runs per composed keystroke and inside the frame draw; a slow X server stalls frames.

giachuvan:
1. text_field.rs — Guard removal / click-while-composing
Agreed. Click-while-composing calls reset_ime() and unmark_text(). Without discarding the in-flight preedit, the engine’s commit echo is applied as a normal insert and the preedit comes back.
Fix: unmark_text again records discard_ime_commit from the marked range (Linux only). prepare_channel_switch already did this. A later composition is not swallowed: the next key goes through replace_and_mark with the first letter, which does not match the discarded syllable, so the flag is cleared.
2. client.rs — Modifier releases eaten by shortcut_skips_ime
Agreed. On X11 a Control/Alt/Super press often has no modifier bit yet, so it is sent to the IM. The release already has the modifier bit, so shortcut_skips_ime was true and the IM never saw the release.
Fix: if shortcut_skips_ime(state, keyval) && !is_modifier. Shortcuts still skip IME; modifier edges are forwarded so press/release stay paired.
3. im_frontend.rs — drop_late latch
Agreed. After a ProcessKeyEvent timeout we must drop leftover signals from that call. Leaving drop_late set until the next keystroke also dropped later async commits (e.g. mouse candidate pick).
Fix: drop_late is one-shot. The next take_events() clears the queue and resets the flag. Later commits are applied.
4. client.rs — FocusOut drain loses unfocus commit
Agreed. keyboard_focused_window was set to None before drain_dbus_im(). drain_dbus_im takes pending events then returns if there is no focused window, so ClientUnfocusCommit from FocusOut was discarded.
Fix: focus_out() first, drain while the window id is still set, then clear focus, reset_ime(), and unmark.
5. text_edit.rs — ime_replace_range prefers outside selection
Agreed. If a marked range exists, an IM commit must replace that preedit (same as GTK/Qt). Preferring a selection that sits outside the mark left the preedit text in place.
Fix: ime_replace_range always uses marked when present, otherwise the selection. Ctrl+A is unchanged: select_all already clears marked_range, so the next commit uses the full selection.
6. client.rs / im_frontend.rs — Modifier edges block on process_key
Agreed on the extra waits. We still send Shift to the IM (needed for VI/EN). We no longer add SIGNAL_WAIT (16ms) or COMMIT_WAIT (40ms) on modifier keys.
The 80ms KEY_TIMEOUT is a hung-IM ceiling, not a sleep. For modifiers we now use a 16ms ceiling and do not increment fail_count / set drop_late on timeout, so a slow Shift does not mark the IM dead or drop later commits.
7. window.rs — Blocking TranslateCoordinates per keystroke / frame
Agreed. A sync X round-trip on every composed key and inside invalidate_character_coordinates can stall the frame if the X server is slow.
Fix: cache the window’s root origin. IME caret = cached origin + local offset. Cache is invalidated on ConfigureNotify / set_bounds. One translate after a move/resize, not per keystroke.

crates/mezon-app/src/main.rs
huylexuan44: Verified fixed in bb54136.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Verified fixed in bb54136; 1cea1ec also caches the root origin so the round trip is off the key path.

crates/vendor/gpui_linux/src/linux/x11/xim_handler.rs
huylexuan44: Verified fixed in bb54136.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
huylexuan44: Verified fixed in bb54136.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Verified fixed in 1cea1ec: is_modifier now bypasses shortcut_skips_ime so press and release both reach the IM.

crates/mezon-widgets/src/text_edit.rs
huylexuan44: Verified fixed in 1cea1ec: marked range now always wins, matching upstream.

crates/vendor/gpui_linux/src/linux/x11/window.rs
huylexuan44: Verified fixed in 1cea1ec: root origin cached and invalidated on configure.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Verified fixed in 1cea1ec: drain runs before the focused window is cleared; a late commit degrades to unmark, no text loss.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
huylexuan44: Verified fixed in 1cea1ec: the latch clears after one drain instead of persisting to the next keystroke.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Mitigated in 1cea1ec: 16ms cap and no signal/commit waits for modifiers; still one blocking call per edge.

crates/mezon-ui/src/chat/mention_input/text_field.rs
huylexuan44: Re-armed in 1cea1ec with consume-on-echo. Note ibus reset never echoes, so a stale guard can swallow a later identical no-preedit commit; consider clearing it on the next key press.

huylexuan44: Verified fixed in bf777f36: on_key_down clears the guard; filtered keys never reach the widget so the echo is still caught. Residual race is ms-scale, not blocking.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Still open at bf777f36: both parsers map empty preedit to HidePreedit (unmark keeps text) and Preedit{empty}->handle_ime_delete has no producer. Repro: compose viet on fcitx5-unikey, press Esc - the canceled word stays as real text. Blocking.

crates/mezon-app/src/main.rs
huylexuan44: Still open at bf777f36: prefers_native_wayland flips every WAYLAND_DISPLAY user to native Wayland - renderer/decorations/scaling/capture/tray all change, and the new D-Bus IM client is X11-only so ibus Vietnamese on wlroots regresses. Keep XWayland default or gate native behind MEZON_LINUX_SESSION=wayland. Blocking.

crates/vendor/gpui_linux/src/linux/wayland/client.rs
huylexuan44: Still open at bf777f36: sync_text_input queues set_surrounding/set_cursor_rectangle unconditionally but commit() only runs when last_serial == serial, so the double-buffered requests sit unapplied and the IM sees stale surrounding on alternate dones. Track your own commit count and commit after every sync.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Still open at bf777f36: maybe_append_fcitx_space is unchanged - space-as-candidate-select (pinyin/mozc) still gets a spurious trailing space. If the supported scope is Vietnamese engines only, say so in a guard comment.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
huylexuan44: Still open at bf777f36: im_frontend.rs has zero eprintln and log:: is dropped per CLAUDE.md - a failed Vietnamese-typing report is undebuggable in the field. Add a MEZON_IM_DEBUG-gated eprintln at connect/reconnect/downgrade/timeout points.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Half-fixed at bf777f36: the shortcut path now unmarks the widget, but the engine is not reset, so it keeps composing the stale syllable and the next key commits old letters. Call reset_ime() here.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
huylexuan44: Partially fixed: modifiers now skip the waits, but Escape/preedit-clear batches still hit only_clears_preedit and burn the full 40ms COMMIT_WAIT for a commit that never comes.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1049,354 @@ impl X11Client {
...
+        if let Some(token) = token {
+            self.0.borrow().loop_handle.remove(token);
+        }
+        drop(im);
+    }
+
+    fn reconnect_dbus_im(&self) -> bool {
+        self.drop_dbus_im();
+        let Some(im) = X11ImContext::connect() else {
+            self.0.borrow_mut().dbus_unavailable = true;
```
huylexuan44, line 1283: dbus_unavailable latches forever: an IM daemon started after the app never gets picked up until restart. Retry on the next FocusIn instead of never.

huylexuan44:
Re-verified at bf777f36. All 7 points in your comment check out in code, and commit 4 (on_key_down clearing discard_ime_commit) closes the residual edge from point 1 - filtered keys never reach the widget, so the echo is still caught; the remaining race is ms-scale and not blocking.

Also verified fixed from earlier rounds: masked-password guards (input.rs + text_field.rs), keysym lowercase names, single set_surrounding per batch, no borrow held across get_ime_surrounding, ibus Destroy on the Service interface, downgrade-only-on-UnknownMethod, the 4000-byte surrounding cap, the D-Bus retry loop, invalidate_character_coordinates handler ordering, Backspace routed through the IM, and the word_range_at rewrite.

Still blocking merge (replies posted in the threads):
- Escape-cancel commits the canceled composition: empty preedit always maps to HidePreedit (unmark keeps text); Preedit-empty -> handle_ime_delete has no producer. Repro: compose a word on fcitx5-unikey, press Esc.
- Native-Wayland default flip in main.rs: changes renderer/decorations/scaling/capture/tray for every Wayland user, and regresses ibus Vietnamese on wlroots since the new IM client is X11-only.

Open, non-blocking: Wayland done-serial commit gate (stale surrounding), fcitx space heuristic vs candidate-selection engines, zero diagnostics in im_frontend, Ctrl+C-while-composing not resetting the engine, 40ms COMMIT_WAIT burn on Escape, dbus_unavailable latching forever (new inline comment).

Also worth a follow-up: only the chat composer arms the commit-echo guard - the other inputs (input.rs, textarea.rs) can still resurrect a preedit on click-away under fcitx5.

Manual X11 + Wayland test results (per the earlier checklist) are still needed before merge; CI cannot cover WM/IM interaction.

giachuvan:
1. Modifier edges still block on process_key

Agreed that a D-Bus call per Shift edge is inherent: Vietnamese VI/EN needs Shift on the IM. We no longer send Ctrl/Alt/Super/Hyper to the IM at all (ime_mode_modifier only keeps Shift, Caps, AltGr, Level5, Mode_switch). Those keys no longer pay KEY_TIMEOUT. Shift still has one short call (16ms ceiling, no signal/commit wait). Making that async would take D-Bus off the UI thread; we are not doing that in this pass.

2. dbus_unavailable latches forever

Agreed. FocusIn → enable_ime already cleared the flag, but if the window never lost focus the key path never retried, so an IM started after the app stayed invisible until restart.

Fix: FocusIn / enable_ime still retries immediately. The key path also retries after a 2s cooldown (dbus_retry_at) so a late ibus/fcitx5 start is picked up without alt-tab and without a connect attempt on every keystroke.

3. Escape-cancel commits the canceled composition — blocking

Agreed. Empty Fcitx5 preedit is mapped to HidePreedit → unmark (keep text). That is required for Space: Unikey often sends hide-only, and deleting would wipe the word. Escape is different: cancel with no commit must drop the preedit.

Fix: on Escape, a hide-only batch (no committed text) is rewritten to empty Preedit, which already calls handle_ime_delete. Space / hide-after-commit is unchanged. Also skip COMMIT_WAIT on Escape so cancel does not burn 40ms waiting for a commit that will not come.

4. Native-Wayland default

Not changing this. MEZON_LINUX_SESSION=x11 still forces X11/XWayland for the D-Bus IM client. Flipping the default changes renderer, decorations, scaling, capture, and tray for every Wayland user. That is a product call, not an IME correctness fix in this patch.

5. Ctrl+C while composing

Agreed. Shortcuts used to unmark only, so the engine kept the composition and the next key continued it.

Fix: shortcut skip now unmark (arms the commit-echo guard) then reset_ime().

6. Commit-echo guard only on the chat composer

Agreed. input.rs and textarea.rs now arm discard_ime_commit on unmark (Linux), swallow the echoed preedit, and clear the flag on the next real key (on_key_down), same as the mention field.

7. Still open / not in this patch

Wayland done-serial / stale surrounding — needs a separate text-input-v3 change.
Fcitx space heuristic — kept; deleting it regresses Vietnamese Space.
Diagnostics in im_frontend — not adding debug logs back.
Manual X11 + Wayland matrix — still required before merge; CI cannot cover WM/IM.

huylexuan44:
Re-checked at `6b15149b` after fix round 5.

Confirmed fixed in code:
- modifier press/release pairing and the reduced Shift path
- one-shot `drop_late`
- FocusOut drain ordering
- marked range precedence
- cached X11 root origin
- Escape cancel rewrite and Escape skipping `COMMIT_WAIT`
- Ctrl+C resetting the IM engine
- commit-echo guards in `input.rs` and `textarea.rs`

Still blocking merge:

1. **CI lint fails because of this PR.** `ImeSurroundingText::from_selection`, `cursor_chars`, and `anchor_chars` are public but have no docs (`platform.rs:1415,1490,1497`). Add docs or the repository-consistent missing-docs expectation.

2. **The native-Wayland default flip is unchanged.** `prefers_native_wayland` still sends every user with `WAYLAND_DISPLAY` to native Wayland unless they explicitly set `MEZON_LINUX_SESSION=x11`. This changes renderer/decorations/scaling/capture/tray outside the IME scope and regresses ibus Vietnamese on wlroots because the new D-Bus IM client is X11-only. If this is a product decision, it needs explicit product approval; otherwise retain the XWayland default or make native Wayland opt-in.

3. **The 2-second D-Bus retry reintroduces the key-path stall from the earlier review.** Both key press and release call `ensure_dbus_im`; after the cooldown, `reconnect_dbus_im` synchronously opens bus connections, calls `NameHasOwner`, and probes ibus files on the X11 event thread. Users without ibus/fcitx5 can therefore hitch again every ~2 seconds while typing. This is the same behavior called out in https://github.com/mezonai/mezon-desktop/pull/373#discussion_r3811851467. Retry on FocusIn, use exponential/backed-off scheduling outside the key path, or move connection work off the UI thread.

Still open/non-blocking from the prior round: Wayland done-serial/commit-count tracking, the fcitx Space heuristic for candidate-selection engines, and missing `im_frontend` diagnostics.

The Escape fix is correct by static inspection, but the X11 + Wayland manual matrix with ibus/fcitx5/bamboo/unikey is still required before merge. Current CI is Format ✅, Lint ❌, Test ❌, Cargo Deny ❌; the Test failure is in untouched `mezon-record`, and Cargo Deny is the existing `h2 0.4.15` advisory, but Lint is directly caused by this PR.

Not ready to merge yet.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1076,367 @@ impl X11Client {
...
+        self.0.borrow_mut().im = Some(im);
+        match result {
+            Ok(filtered) => {
+                let Some(window) = self.get_window(window_id) else {
+                    return filtered;
+                };
+                if kind == super::im_frontend::ImKind::Fcitx5 && !is_release && filtered {
+                    maybe_append_fcitx_space(keyval, &mut events);
+                }
+                if keyval == 0xff1b && !is_release {
```
huylexuan44: This rewrite only sees events returned synchronously, so it covers fcitx5 batch mode only: ibus and fcitx5-single deliver the preedit-clear as an async signal, Escape now skips SIGNAL_WAIT, so the late HidePreedit drains as unmark and the canceled word still stays on ibus. Latch escape-pending and rewrite hide-only batches in the next drain, or keep a short wait for Escape.

crates/vendor/gpui_linux/src/linux/x11/client.rs
huylexuan44: Accepting the heuristic as Vietnamese-scoped per your rationale - please add a code comment on maybe_append_fcitx_space stating the Unikey hide-only-Space case it exists for and the known CJK candidate-select tradeoff, so it survives future refactors.

huylexuan44:
Verified 6b15149 against your reply - all five fixes check out in code:

1. ime_mode_modifier keeps only Shift/Caps/AltGr/Level5/Mode_switch; Ctrl/Alt/Super skip the IM entirely. Daemon-grabbed toggles (Ctrl+Space) are unaffected since the engine grabs those itself - only bare Shift needs forwarding, and it still is.
2. dbus_retry_at: FocusIn retries immediately, key path at most once per 2s. Bounded reintroduction of the old retry - about two cheap NameHasOwner calls per 2s while typing with no IM running. Acceptable.
3. rewrite_escape_cancel + skipping the waits on Escape works for fcitx5 batch mode, where events come back in the method reply. It does not cover ibus / fcitx5-single, where the preedit-clear arrives as an async signal after the rewrite point - inline comment posted with the fix options. This is the one remaining correctness item on the Escape path.
4. Shortcut path now unmarks then reset_ime() - correct order, arms the echo guard.
5. swallow_discarded_ime_commit extracted to mezon-widgets with unit tests and wired into input.rs and textarea.rs - matches the mention field.

Deferred items (Wayland done-serial, diagnostics) noted as your call to defer; space heuristic accepted as Vietnamese-scoped with a request for a code comment. The native-Wayland default question stays open pending a maintainer decision - not re-arguing it here.

Once the ibus Escape gap is handled, the remaining gate is the manual X11 + Wayland matrix.

giachuvan:
Docs / CI lint

Agreed. ImeSurroundingText::from_selection, cursor_chars, and anchor_chars now have rustdoc so they satisfy missing_docs.

D-Bus retry on the key path

Agreed. The 2s cooldown still ran reconnect_dbus_im() from KeyPress/KeyRelease and could hitch users with no IM. dbus_unavailable latches on the key path again — no connect / NameHasOwner / ibus file probe while typing. Retry only from FocusIn → enable_ime (and other enable paths).

Escape cancel on IBus / fcitx5-single

Agreed. The rewrite only saw the synchronous process_key batch; IBus and fcitx5-single deliver hide as a later signal, and Escape skips SIGNAL_WAIT, so the late HidePreedit was applied as unmark.

Escape press now sets escape_cancel_pending. apply_im_events (key reply and drain_dbus_im) rewrites a later hide-only batch to empty Preedit → handle_ime_delete. A batch that already has text, the next non-Escape key, reset_ime, or FocusOut clears the latch so a later hide cannot delete a new composition. Fcitx5 batch is unchanged: hide in the method reply is rewritten in the same apply.

Fcitx Space heuristic

Kept for Vietnamese Unikey/Bamboo. Added a short comment at maybe_append_fcitx_space about the CJK trailing-space tradeoff.

Native Wayland default

Agreed this blocked merge and was a product default, not an IME-only tweak. We restored XWayland as the default.

prefers_native_wayland is true only when MEZON_LINUX_SESSION=wayland and WAYLAND_DISPLAY is set. Users with both DISPLAY and WAYLAND_DISPLAY get X11/XWayland and the D-Bus IM client with no extra env. Native Wayland is opt-in. If DISPLAY is unset we leave WAYLAND_DISPLAY (no X server to fall back to).

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1075,369 @@ impl X11Client {
...
+        true
+    }
+
+    fn apply_im_events(&self, window: &X11WindowStatePtr, mut events: Vec<ImEvent>) {
+        if self.0.borrow().escape_cancel_pending {
+            if events.iter().any(im_event_has_text) {
+                self.0.borrow_mut().escape_cancel_pending = false;
+            } else if !events.is_empty() {
+                rewrite_escape_cancel(&mut events);
+                self.0.borrow_mut().escape_cancel_pending = false;
```
huylexuan44: `escape_cancel_pending` is cleared for every non-empty batch without text, even when this batch contains no `HidePreedit`. `Connection::process` handles one D-Bus message per call, so an async engine can deliver `ForwardKey` or `DeleteSurrounding` first and `HidePreedit` on the next drain. The first event makes `rewrite_escape_cancel` a no-op but still clears the latch; the later hide then falls back to unmark and keeps the canceled composition. Make the rewrite return whether it actually converted a hide and only clear the latch then (or on a real text commit/new key/reset/FocusOut). Please add a split-drain test such as `ForwardKey` followed by `HidePreedit`.

giachuvan: Agreed. Connection::process can deliver ForwardKey/DeleteSurrounding on one drain and HidePreedit on the next. The previous code cleared escape_cancel_pending for any non-empty textless batch, so a no-op rewrite on the first drain dropped the latch and the later hide fell through to unmark.

rewrite_escape_cancel now returns whether it converted a hide. The latch clears only then, or when the batch already has real text (commit / non-empty preedit). Reset / FocusOut / the next non-Escape key still clear it as before. Added a split-drain unit test: ForwardKey keeps the latch, then HidePreedit converts to empty Preedit.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -901,6 +1088,365 @@ impl X11Client {
...
+            return false;
+        }
+        let mut client_state = self.0.borrow_mut();
+        let Some(mut im) = client_state.im.take() else {
+            return false;
+        };
+        drop(client_state);
+        let kind = im.kind();
+        let result = im.process_key(keyval, keycode, state, is_release, time);
+        let mut events = im.take_events();
```
huylexuan44, line 1178: [P1] Keep the timeout quarantine past this immediate empty drain. `process_key` clears the queue and sets `drop_late` on `Err`, but this unconditional `take_events()` runs immediately afterward; the queue is normally empty, so `take_events` resets `drop_late` before a late `Commit`/`Preedit` signal arrives. The caller then inserts the raw key, and the fd watch applies the late signal, reproducing duplicate characters. Please add a regression sequence `timeout -> empty take -> late Commit` and either correlate/reset the context or retain quarantine until stale events are actually discarded.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1225 @@
...
+    let cursor = cursor.min(text.len());
+    if !text.is_char_boundary(cursor) {
+        return (0, 0);
+    }
+    let caret_chars = text[..cursor].chars().count() as i32;
+    let start_chars = caret_chars.saturating_add(offset).max(0) as usize;
+    let end_chars = start_chars.saturating_add(nchars as usize);
+    let start_bytes = char_index_to_byte(text, start_chars);
+    let end_bytes = char_index_to_byte(text, end_chars);
+    let before = cursor.saturating_sub(start_bytes);
```
huylexuan44: [P1] Preserve the exact `DeleteSurroundingText(offset, size)` range instead of expanding it to the caret. For `text="abcd"`, cursor at 4, `offset=-4`, `nchars=1`, the protocol requests deletion of only `a`; this helper returns `(before=4, after=0)`, so `surrounding_delete_range` removes all of `abcd`. Fcitx defines the range start as `cursor + offset`, which need not touch the caret. The current before/after API cannot represent a disjoint range, so pass an exact relative/document range (or safely reject unsupported non-adjacent requests) and add tests on both sides of the caret.

huylexuan44 (commented):
Re-review after 60fef71: the split-drain Escape fix is correct and its regression test covers the reported ordering. The remaining findings below are independent; I do not recommend merging the current Wayland/X11 IME scope until the P1 items are fixed.

crates/vendor/gpui_linux/src/linux/wayland/client.rs
huylexuan44: [P1] Track the number of client `commit()` requests, not the previous `done` serial. text-input-v3 says `done.serial` equals the count of commits already issued by this object. After `enable_ime()` sends commit #1, a matching `done { serial: 1 }` sees `last_serial == 0` here and incorrectly skips the commit, leaving the newly queued surrounding text/cursor rectangle unapplied. The composing branch of `update_ime_position` also queues a cursor rectangle without committing it. Keep an explicit client commit counter and commit pending state only when `serial` matches that count.

crates/vendor/gpui_linux/src/linux/x11/im_frontend.rs
```diff
@@ -0,0 +1,1225 @@
...
+
+fn parse_fcitx5_preedit(message: &Message) -> Option<ImEvent> {
+    let (chunks, caret): (Vec<(String, i32)>, i32) = message.read2().ok()?;
+    let text: String = chunks.into_iter().map(|(piece, _)| piece).collect();
+    if text.is_empty() {
+        return Some(ImEvent::HidePreedit);
+    }
+    Some(ImEvent::Preedit {
+        text,
+        caret_chars: caret,
```
huylexuan44: [P2] Fcitx5 preedit cursors are UTF-8 byte offsets, but this stores the value as `caret_chars` and later counts Unicode scalars in `caret_utf16_range`. Fcitx forwards `Text::cursor()` directly in `UpdateFormattedPreedit`, and that API is documented as byte-based. For `"hóa"`, byte cursor 3 (after `"hó"`) is interpreted as 3 characters and moves the caret to the end. Convert the Fcitx byte boundary to UTF-16 here, or carry the backend-specific encoding separately; add a multibyte mid-preedit test.

crates/mezon-canvas/src/editor.rs
```diff
@@ -1712,11 +1714,11 @@ impl EntityInputHandler for CanvasEditorState {
...
     ) {
-        let range = range_utf16
-            .as_ref()
-            .map(|r| self.range_from_utf16(r))
-            .or(self.marked_range.clone())
-            .unwrap_or(self.selected_range.clone());
+        let range = if let Some(range_utf16) = range_utf16.as_ref() {
+            self.range_from_utf16(range_utf16)
+        } else {
+            ime_replace_range(&self.selected_range, self.marked_range.as_ref())
```
huylexuan44, line 1739: [P2] Now that this path accepts D-Bus IME commits, add the same Reset-echo protection used by Input, TextArea, and MentionInput. The X11 click/shortcut path calls `Reset` and then unmarks; some D-Bus engines emit the old preedit as a late `Commit`. Canvas `unmark_text` only clears `marked_range`, so preedit `"hoa"` remains and the echoed `"hoa"` is inserted here at the caret, producing `"hoahoa"`. Store/swallow the discarded commit for Canvas and cover `preedit -> unmark/reset -> identical commit`.

crates/vendor/gpui_linux/src/linux/x11/xim_handler.rs
```diff
@@ -103,13 +132,42 @@ impl XimHandler {
...
```
huylexuan44: [P2] Include the caret in this duplicate-suppression key. `handle_preedit_caret` emits the same text with a new caret, but `should_skip_apply` compares only the text, so every caret-only XIM update returns early in `xim_handle_preedit`. Store `(text, caret)` or suppress only a fully identical preedit event; add a test where the string stays constant and the caret moves.

giachuvan:
1. drop_late cleared by empty take_events — P1

Agreed. On timeout we set drop_late and clear the queue; the immediate empty take_events() was resetting the flag before a late Commit/Preedit arrived, so the fd watch applied it after the raw key was inserted.

Fix: quarantine stays set while the queue is empty. Clearing happens only when a late event is actually discarded, or at the start of the next process_key (drain + clear first). Added a regression unit test: timeout → empty take keeps the latch → late Commit is discarded.

2. DeleteSurroundingText expanded to the caret — P1

Agreed. For text="abcd", cursor=4, offset=-4, nchars=1, Fcitx asks to delete only a. The old helper returned (before=4, after=0) and wiped abcd.

GPUI only exposes adjacent before/after deletes. Fix: map only ranges that touch the caret (before-only, after-only, or spanning). Disjoint ranges return (0, 0) and are ignored. Adjacent VN cases (offset=-1) unchanged. Tests cover both sides of the caret and disjoint rejection.

3. Wayland done.serial vs last done serial — P1

Agreed. text-input-v3 done.serial is the count of client commit() calls. After enable_ime commit #1, done { serial: 1 } saw last_serial == 0 and skipped the follow-up commit, leaving surrounding/cursor from sync_text_input unapplied. Composing update_ime_position also set the cursor rectangle without committing.

Fix: track text_input_commit_count. Commit pending state only when serial == commit_count. Composing cursor updates now commit too.

4. Fcitx5 preedit cursor is bytes — P2

Agreed. Fcitx Text::cursor() is byte-based; we stored it as caret_chars and counted Unicode scalars. For "hóa", byte 3 (after "hó") became char 3 and jumped to the end.

Fix: convert Fcitx byte caret → character count at parse time (signal + batch). IBus stays character-based. Multibyte mid-preedit unit test added.

5. Canvas missing Reset-echo guard — P2

Agreed. Canvas unmark_text only cleared marked_range, so a late identical Commit after Reset produced "hoahoa".

Fix: same discard_ime_commit / swallow_discarded_ime_commit path as Input/TextArea/MentionInput, plus on_key_down clearing. Test: preedit → unmark → identical commit is swallowed.

6. XIM should_skip_apply ignores caret — P2

Agreed. handle_preedit_caret re-emits the same text with a new caret; text-only comparison skipped those updates.

Fix: store (text, caret) and skip only fully identical preedit events. Unit test: same string, moved caret is not skipped.

PR #374, opened 2026-08-19
Fix bug open file when send

giachuvan:


crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
                         .collect::<Vec<_>>()
                 })
                 .await;
-            this.update_in(cx, |this, window, cx| this.add_pending(pending, window, cx))
-                .ok();
+            this.update_in(cx, |this, window, cx| {
+                if this.import_generation != generation {
```
huylexuan44: Second drop bumps import_generation, so the first batch files are silently discarded here. Use an in-flight counter.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
...
+                    return;
+                }
+                this.importing_attachments = false;
```
huylexuan44: importing_attachments is cleared by whichever import finishes first, so Enter can still submit while an older one runs.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
...
+                this.add_pending(pending, window, cx);
+                this.claim_composer_focus_after_drop(window, cx);
+                if this.submit_after_import {
+                    this.submit_after_import = false;
+                    if !this.popup_open() {
+                        cx.emit(MentionInputEvent::Submit);
```
huylexuan44: bind_channel never clears submit_after_import, so this can send into a different channel. Capture draft_channel on Enter.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
...
                 .await;
-            this.update_in(cx, |this, window, cx| this.add_pending(pending, window, cx))
-                .ok();
+            this.update_in(cx, |this, window, cx| {
+                if this.import_generation != generation {
+                    return;
+                }
+                this.importing_attachments = false;
+                this.add_pending(pending, window, cx);
+                this.claim_composer_focus_after_drop(window, cx);
```
huylexuan44: Re-claiming focus after the async import steals focus and raises the window later. Activate only at drop time.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
...
                 })
                 .await;
-            this.update_in(cx, |this, window, cx| this.add_pending(pending, window, cx))
-                .ok();
+            this.update_in(cx, |this, window, cx| {
+                if this.import_generation != generation {
+                    return;
+                }
+                this.importing_attachments = false;
+                this.add_pending(pending, window, cx);
```
huylexuan44: add_pending may open the upload-limit modal and focus it; line 1078 steals that focus back and the submit still fires.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
...
+                this.claim_composer_focus_after_drop(window, cx);
+                if this.submit_after_import {
+                    this.submit_after_import = false;
```
huylexuan44: submit_after_import is reset before the popup check, so a queued Enter is silently swallowed. Move the reset inside.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1030,6 +1056,10 @@ impl MentionInput {
         if paths.is_empty() {
             return;
         }
+        self.import_generation = self.import_generation.wrapping_add(1);
+        let generation = self.import_generation;
+        self.importing_attachments = true;
+        self.claim_composer_focus_after_drop(window, cx);
```
huylexuan44: Claims focus unconditionally, but area.rs gates it on input_visible; drops in read-only channels now steal focus.

crates/mezon-ui/src/chat/area.rs
```diff
@@ -642,6 +642,7 @@ impl ChatArea {
                             let dropped: Vec<PathBuf> = paths.paths().to_vec();
                             drop_input.update(cx, |input, cx| {
                                 if input_visible {
+                                    window.activate_window();
```
huylexuan44: Redundant now that add_dropped_paths activates the window itself. Better: return early when send_denied.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +714,26 @@ impl MentionInput {
...
+        self.focus_input(window, cx);
+    }
+
+    fn claim_composer_focus_after_drop(&self, window: &mut Window, cx: &mut Context<Self>) {
+        self.claim_composer_focus(window, cx);
+        let this = cx.entity();
+        #[cfg(target_os = "linux")]
+        {
+            let this = this.clone();
+            window.on_next_frame(move |window, cx| {
```
huylexuan44: on_next_frame only queues a callback, it does not request a frame, so this may never run.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1384,9 +1427,14 @@ impl MentionInput {
     fn on_enter(&mut self, window: &mut Window, cx: &mut Context<Self>) {
         if self.popup_open() {
             self.accept(self.selected, window, cx);
-        } else {
-            cx.emit(MentionInputEvent::Submit);
+            return;
+        }
+        if self.importing_attachments {
```
huylexuan44: open_file_picker and on_paste_images have the same race but no guard, so Enter after picking files still submits early.

crates/mezon-ui/src/chat/mention_input/mod.rs
huylexuan44: import_generation discards the earlier batch: a second drop or paste silently loses the first drop attachments. Imports are additive, not replacements - use a refcount and always call add_pending.

crates/mezon-ui/src/chat/mention_input/mod.rs
huylexuan44: bind_channel/apply_draft never reset these fields, so a queued Enter auto-sends into whichever channel is active when the import lands. Capture draft_channel at import start and clear all three on switch.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
...
-            this.update_in(cx, |this, window, cx| this.add_pending(pending, window, cx))
-                .ok();
+            this.update_in(cx, |this, window, cx| {
+                if this.import_generation != generation {
+                    return;
+                }
+                this.importing_attachments = false;
+                this.add_pending(pending, window, cx);
+                this.claim_composer_focus_after_drop(window, cx);
+                if this.submit_after_import {
```
huylexuan44: Submit fires even when add_pending rejected the batch (upload limit) or candidates was empty, sending the bare text while the limit modal is open. Gate the submit on the batch actually landing.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1030,6 +1056,10 @@ impl MentionInput {
         if paths.is_empty() {
             return;
         }
+        self.import_generation = self.import_generation.wrapping_add(1);
```
huylexuan44: Only add_dropped_paths is guarded. open_file_picker, on_paste_images and convert_text_to_file still build attachments without arming the flag, so the file-picker case in the PR title keeps the original race.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1030,6 +1056,10 @@ impl MentionInput {
...
+        let generation = self.import_generation;
+        self.importing_attachments = true;
```
huylexuan44: Setting this does not cx.notify() and nothing renders a pending placeholder, so Enter looks completely dead for the whole import (video probe can take seconds).

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1039,8 +1069,21 @@ impl MentionInput {
...
+            this.update_in(cx, |this, window, cx| {
+                if this.import_generation != generation {
+                    return;
+                }
+                this.importing_attachments = false;
+                this.add_pending(pending, window, cx);
+                this.claim_composer_focus_after_drop(window, cx);
+                if this.submit_after_import {
+                    this.submit_after_import = false;
+                    if !this.popup_open() {
```
huylexuan44: submit_after_import is already cleared here, so a popup that reopened during the import swallows the queued Enter with no feedback. Keep the flag or accept the suggestion instead.

crates/mezon-ui/src/chat/mention_input/mod.rs
huylexuan44: Re-claiming focus on completion calls activate_window after an unbounded delay and steals OS foreground from whatever app the user switched to. Only claim focus at drop time.

crates/mezon-ui/src/chat/area.rs
huylexuan44: add_dropped_paths already calls activate_window plus focus_input, so this pair is redundant. Remove it or push the input_visible guard down.

crates/mezon-ui/src/chat/area.rs
```diff
@@ -642,6 +642,7 @@ impl ChatArea {
...
                                     input.focus_input(window, cx);
                                 }
                                 input.add_dropped_paths(dropped, window, cx)
```
huylexuan44, line 647: add_dropped_paths runs even when input_visible is false, and it now activates the window and focuses a composer the user cannot see. Pass the guard into it.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +714,26 @@ impl MentionInput {
         window.focus(&handle, cx);
     }
 
+    fn claim_composer_focus(&self, window: &mut Window, cx: &mut App) {
+        window.activate_window();
+        self.focus_input(window, cx);
+    }
+
+    fn claim_composer_focus_after_drop(&self, window: &mut Window, cx: &mut Context<Self>) {
```
huylexuan44: This fires activate_window 2-3 times per call (immediate, on_next_frame on Linux, defer) and is called twice per drop. One immediate claim plus one window.defer is the whole mechanism.

crates/mezon-ui/src/chat/mention_input/mod.rs
huylexuan44: on_next_frame never runs without a scheduled repaint - window.focus early-returns when the handle is already focused and add_pending skips notify on an empty batch - and it pins a strong entity meanwhile. The defer below already covers this.

crates/mezon-ui/src/chat/mention_input/mod.rs
huylexuan44: capture.rs composer_submit still returns ok:true with a sent snapshot while the Enter is only queued here, so the MCP harness believes the message went out.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1384,9 +1427,14 @@ impl MentionInput {
     fn on_enter(&mut self, window: &mut Window, cx: &mut Context<Self>) {
         if self.popup_open() {
             self.accept(self.selected, window, cx);
-        } else {
-            cx.emit(MentionInputEvent::Submit);
+            return;
+        }
+        if self.importing_attachments {
+            self.submit_after_import = true;
+            self.claim_composer_focus(window, cx);
```
huylexuan44: activate_window on every Enter is pointless: the window must already be key to have delivered the key event. Setting the flag is enough.

huylexuan44:
## Root cause analysis: why this is Linux-only, and where the real fix belongs

### Do macOS and Windows have this bug? No — and the reason is at the OS level

No GPUI backend activates or focuses the window when it accepts a drop. I checked all four:

| Platform | Drop handler | Activates/focuses? |
|---|---|---|
| macOS | `perform_drag_operation` (`gpui_macos/src/window.rs:2935`) | No |
| Windows | `IDropTarget::Drop` (`gpui_windows/src/window.rs:1240`) | No |
| X11 | `XdndDrop` (`gpui_linux/src/linux/x11/client.rs:1010`) | No |
| Wayland | `wl_data_device::Event::Drop` (`gpui_linux/src/linux/wayland/client.rs:2432`) | No |

macOS and Windows do not reproduce because the OS does it for us (AppKit makes the window key on drop; Windows OLE drag-drop plus `IDropTargetHelper` gives the target window foreground activation). X11/Wayland have no equivalent — XDND does not transfer keyboard focus at all. So this is genuinely Linux-only and the fix should be Linux-scoped.

### The actual X11 defect — and why the retry was needed

`gpui_linux/src/linux/x11/window.rs:1519`:

```rust
let data = [1, xproto::Time::CURRENT_TIME.into(), 0, 0, 0];
//              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ = 0
```

`_NET_ACTIVE_WINDOW` is sent with **timestamp 0 (CurrentTime)**. EWMH requires the timestamp of the client's last user interaction; with 0, mutter/KWin/Xfwm apply focus-stealing prevention and typically only set the urgency hint (taskbar flash) instead of actually focusing.

That is why a single `activate_window()` did not stick, why commit 1 needed `on_next_frame`, and why commit 2 still keeps a `window.defer` retry. **Retrying a request the WM rejected is a lottery** — it will behave differently across GNOME/KDE/XFCE.

The timestamp we need is already available and currently ignored: the `XdndDrop` client message carries it in `data.l[2]`, which `x11/client.rs:937` already destructures as `arg2`.

### Recommended fix (Option 1) — fix it in the platform layer

In the `XdndDrop` branch of `x11/client.rs:1010`, request activation with the real drop timestamp before dispatching `Submit`:

```rust
} else if event.type_ == state.atoms.XdndDrop {
    let drop_time = arg2;                      // XdndDrop data.l[2]
    xdnd_send_finished(&state.xcb_connection, &state.atoms, event.window, state.xdnd_state.other_window);
    xdnd_request_activation(&state.xcb_connection, &state.atoms, event.window, state.x_root_window, drop_time);
    let position = state.xdnd_state.position;
    drop(state);
    window.handle_input(PlatformInput::FileDrop(FileDropEvent::Submit { position }));
    self.0.borrow_mut().xdnd_state = Xdnd::default();
}
```

`xdnd_request_activation` sends `_NET_ACTIVE_WINDOW` with `[1, drop_time, 0, 0, 0]`, falling back to `set_input_focus(..., window, drop_time)` where the WM does not support it.

Why this is better:

- **Deterministic.** One spec-compliant request that WMs accept. No `window.defer`, no retry.
- Fixes **every** drop target in the app (composer, avatar upload, settings), not just the chat composer.
- Removes `activate_window()` from UI code entirely — no double platform call, no risk of stealing focus on macOS/Windows.
- This is a vendor edit, but the repo already carries `mezon vendor edit` markers in exactly these two files (`x11/window.rs` and `wayland/window.rs`, both in `activate`), so the re-apply cost on a snapshot bump is already being paid.

The app side then reduces to:

```rust
fn claim_composer_focus_after_drop(&self, window: &mut Window, cx: &mut Context<Self>) {
    self.focus_input(window, cx);   // focusing the FocusHandle is app-level and correct everywhere
}
```

### Option 2 — stopgap if you do not want to touch vendor yet

```rust
fn claim_composer_focus_after_drop(&self, window: &mut Window, cx: &mut Context<Self>) {
    #[cfg(target_os = "linux")]
    window.activate_window();
    self.focus_input(window, cx);
}
```

One call, no defer, `#[cfg]`-gated so macOS/Windows do not pay a spurious `makeKeyAndOrderFront:` plus foreground-executor spawn. But be clear that on X11 this can still be rejected by the WM because of the zero timestamp — it is mitigation, not a fix.

### Warning: Wayland is probably still broken

The vendored code says so itself (`wayland/window.rs:1297`):

> "Even though the activation is likely going to be rejected, KWin and Mutter can use the app_id to visually indicate we're requesting attention."

Self-activation without a token from the drag source is refused by KWin/Mutter. So this PR most likely does **not** fix Wayland. The correct path is `xdg_activation_v1` with a token whose serial comes from the drop event's input serial. If that is out of scope for this snapshot, the team should at least know Wayland is still affected and test it explicitly.

### Please split this PR

It currently bundles two unrelated bugs:

- **Bug A** (the reported one): no focus after an XDND drop on Linux, so Enter goes to another app. The real fix is a handful of lines.
- **Bug B**: Enter pressed before the async import finishes sends the message without the attachment. That is `imports_in_flight` / `submit_after_import` / `QueuedImportSubmit` — roughly **130 of the 163 added lines**.

Bug B is real and cross-platform and worth fixing, but it is the source of every behavioural risk flagged inline below (Enter silently swallowed when nothing lands, attachments lost on channel switch, no importing indicator). Suggest shipping Bug A first (small, Linux-scoped, easy to review) and taking Bug B in its own PR with tests — there are no tests in this PR, and the queue -> landed -> submit state machine is pure logic that is easy to cover.

### Performance

For the record: perf is clean. None of the new state is read in `render`, so there is no per-frame cost and no new allocation on a hot path. The composer sits outside the `.cached()` wrappers in `area.rs` (669 wraps the timeline, 680 the typing row), so the extra `cx.notify()` calls cannot cascade into cached subtrees. The only measurable items are two redundant composer re-renders per import and two `activate_window()` platform calls per drop.

One real consequence worth knowing: `build_pending` runs `mezon_video::probe_video` (poster-frame decode) sequentially for the whole batch in a single background task. That used to only delay the attachment chips; now it gates the Enter key for the entire batch, with no UI indication.

### How to verify

This only reproduces on Linux, so it needs the Linux GUI container.

1. `echo $XDG_SESSION_TYPE` — test **X11 and Wayland separately**; different code paths, different failure modes.
2. Drag a file from the file manager into the composer and press Enter without clicking anything.
3. Measure directly with `xdotool getactivewindow getwindowname` right after the drop — that tells you whether activation actually landed, instead of inferring it from "did the file open in a browser".
4. Regression check on macOS/Windows: dropping from a background app must not yank the window to the foreground.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +719,104 @@ impl MentionInput {
         window.focus(&handle, cx);
     }
 
+    fn claim_composer_focus(&self, window: &mut Window, cx: &mut App) {
+        window.activate_window();
```
huylexuan44: X11 _NET_ACTIVE_WINDOW is sent with timestamp 0 (x11/window.rs:1519); WMs reject that, which is why one call never stuck.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +719,104 @@ impl MentionInput {
...
+        self.focus_input(window, cx);
+    }
+
+    fn claim_composer_focus_after_drop(&self, window: &mut Window, cx: &mut Context<Self>) {
+        self.claim_composer_focus(window, cx);
+        let this = cx.entity();
+        window.defer(cx, move |window, cx| {
```
huylexuan44: Retrying a WM-rejected activation is a lottery. Fix the X11 timestamp, drop this defer, and gate the call to linux.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +719,104 @@ impl MentionInput {
         window.focus(&handle, cx);
     }
 
+    fn claim_composer_focus(&self, window: &mut Window, cx: &mut App) {
```
huylexuan44: Wayland self-activation is rejected without a token from the drag source, so this likely does not fix Wayland at all.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -464,6 +469,8 @@ pub struct MentionInput {
     preview_cache: Entity<LruImageCache>,
     settings: Entity<Settings>,
     pending_attachments: Vec<PendingAttachment>,
+    imports_in_flight: usize,
```
huylexuan44: Enter-before-import-finishes is a separate cross-platform bug from the Linux focus fix. Split it into its own PR.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +719,104 @@ impl MentionInput {
...
+    fn flush_submit_after_import(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        let Some(queued) = self.submit_after_import.take() else {
+            return;
+        };
+        if self.popup_open() {
+            self.accept(self.selected, window, cx);
+            return;
+        }
+        if queued.channel == self.draft_channel && queued.landed {
+            cx.emit(MentionInputEvent::Submit);
```
huylexuan44: Queued Enter is dropped when nothing lands (build_pending rejects every file, no toast) so the user just gets silence.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +719,104 @@ impl MentionInput {
...
+    }
+
+    fn finish_import(
+        &mut self,
+        pending: Vec<PendingAttachment>,
+        import_channel: Option<ChannelId>,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        let landed = if self.draft_channel == import_channel {
```
huylexuan44: Attachments are discarded entirely when the channel changed. Route them into the leaving channel saved draft instead.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +719,104 @@ impl MentionInput {
...
+        self.claim_composer_focus(window, cx);
+        let this = cx.entity();
+        window.defer(cx, move |window, cx| {
+            this.update(cx, |this, cx| this.claim_composer_focus(window, cx));
+        });
+    }
+
+    fn begin_import(&mut self, cx: &mut Context<Self>) -> Option<ChannelId> {
+        self.imports_in_flight = self.imports_in_flight.saturating_add(1);
+        cx.notify();
```
huylexuan44: Nothing renders imports_in_flight, so this notify is a no-op re-render and the user sees no importing state.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -708,6 +719,104 @@ impl MentionInput {
...
+        if self.imports_in_flight == 0 {
+            self.flush_submit_after_import(window, cx);
+        }
+    }
+
+    fn flush_submit_after_import(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        let Some(queued) = self.submit_after_import.take() else {
+            return;
+        };
+        if self.popup_open() {
```
huylexuan44: on_enter never queues while the popup is open, so this accepts a mention the user never pressed Enter for.

crates/mezon-ui/src/app/capture.rs
```diff
@@ -632,10 +632,12 @@ pub fn composer_drop_paths(cx: &mut App, paths: Vec<String>) -> anyhow::Result<V
...
     let before = composer_snapshot(cx)?;
-    with_composer(cx, |composer, window, cx| {
-        composer.probe_enter(window, cx);
-    })?;
-    Ok(json!({ "ok": true, "sent": before }))
+    let submitted = with_composer(cx, |composer, window, cx| composer.probe_enter(window, cx))?;
+    if submitted {
+        Ok(json!({ "ok": true, "sent": before }))
+    } else {
+        Ok(json!({ "ok": true, "queued": true, "sent": before }))
```
huylexuan44: probe_enter also returns false when the popup was open, so this reports queued when nothing was queued.

giachuvan:
What I changed:
This is a Linux-only focus bug. No GPUI backend activates the window on drop; macOS/Windows still work because the OS gives the drop target keyboard focus. XDND does not transfer focus, and _NET_ACTIVE_WINDOW with timestamp 0 (CurrentTime) is rejected by mutter/KWin/Xfwm (focus-stealing prevention). That is why a single activate_window() never stuck and why a window.defer retry was a lottery.
The real X11 timestamp is already on the XdndDrop client message (data.l[2], already destructured as arg2). The fix belongs in the platform layer, not the composer.
X11 (gpui_linux x11/client.rs): on XdndDrop, request activation with the drop timestamp (arg2, falling back to the XdndPosition timestamp). _NET_ACTIVE_WINDOW is sent as [1, drop_time, 0, 0, 0], plus set_input_focus(..., drop_time). One spec-compliant request; no retry.
Wayland (wayland/client.rs + window.rs): best-effort xdg_activation_v1 using the DataDevice (drag Enter) serial. Self-activation without a token from the drag source can still be rejected by KWin/Mutter — please test X11 and Wayland separately (echo $XDG_SESSION_TYPE).
App/UI: dropped activate_window(), on_next_frame / window.defer, and all Linux-ungated retries. After a drop we only focus_input (app-level FocusHandle), which is correct on every platform. No spurious makeKeyAndOrderFront: / foreground activation on macOS/Windows.
This also covers every drop target in the app (composer, avatar upload, settings), not just the chat composer.
PR split (comments 4–9)
We agree this PR bundled two bugs. Bug B is out of this change:
Enter-before-import-finishes (imports_in_flight / submit_after_import / QueuedImportSubmit)
queued Enter dropped when build_pending rejects every file (silence, no toast)
attachments discarded on channel switch
cx.notify() on imports_in_flight with nothing rendered
queued Enter while a mention popup is open
probe_enter reporting queued when the popup was open and nothing was queued
Composer / capture.rs / area.rs are back to the previous behavior: no queue state machine. Bug B is real and cross-platform; we’ll take it in a follow-up with tests for the queue → landed → submit logic.
How to verify
Linux GUI only:
echo $XDG_SESSION_TYPE — test X11 and Wayland separately.
Drag a file from the file manager into the composer and press Enter without clicking anything.
Optionally measure activation with xdotool getactivewindow getwindowname right after the drop.
Regression: on macOS/Windows, dropping from a background app must not yank the window to the foreground.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -2444,6 +2461,47 @@ fn xdnd_get_supported_atom(
...
+// _NET_ACTIVE_WINDOW with the drop timestamp (XdndDrop data.l[2]), not
+// CurrentTime — WMs reject timestamp 0 as focus-stealing.
+fn xdnd_request_activation(
+    xcb_connection: &XCBConnection,
+    atoms: &XcbAtoms,
+    window: xproto::Window,
+    root: xproto::Window,
+    timestamp: u32,
+) {
+    let time = if timestamp == 0 {
```
huylexuan44: Time::CURRENT_TIME is Self(0) in x11rb, so this branch returns the exact 0 the fix exists to avoid. Handle 0 for real (skip the request) or drop the branch.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -2444,6 +2461,47 @@ fn xdnd_get_supported_atom(
...
+            false,
+            root,
+            EventMask::SUBSTRUCTURE_REDIRECT | EventMask::SUBSTRUCTURE_NOTIFY,
+            message,
+        ),
+    )
+    .log_err();
+    check_reply(
+        || "Failed to set input focus for XDnD",
+        xcb_connection.set_input_focus(xproto::InputFocus::POINTER_ROOT, window, time),
```
huylexuan44: SetInputFocus bypasses the WM focus-stealing policy the _NET_ACTIVE_WINDOW above just deferred to; gate it on _NET_ACTIVE_WINDOW missing from _NET_SUPPORTED, and prefer revert_to PARENT over POINTER_ROOT.

crates/vendor/gpui_linux/src/linux/wayland/window.rs
```diff
@@ -1046,6 +1046,23 @@ impl WaylandWindowStatePtr {
         }
     }
 
+    pub fn request_dnd_activation(&self) {
```
huylexuan44, line 1052: Missing the mezon vendor edit marker (also on the call site at wayland/client.rs:2447) — required so this survives the next snapshot bump.

huylexuan44:
Re-reviewed a8d23c82. The platform-layer approach is right and both backends are wired end-to-end (verified the Wayland token path: Done -> PendingActivation::Window -> activation.activate). Merge status:

- Not mergeable yet: 3 inline comments remain (dead timestamp==0 branch, SetInputFocus bypassing the WM policy, missing vendor-edit markers on the Wayland side).
- Lint failure here is a stale branch (canvas errors fixed on develop by b784a6f8) - but note develop's tip now fails Lint again on integrations_tab.rs:661 (useless .clone(), from #375), so rebasing alone won't go green until that lands.
- Cargo Deny (orphaned RUSTSEC-2024-0415 ignore in deny.toml) and Test (mezon-record recorder.rs:387 'finish: Incomplete') fail identically on develop - pre-existing, not this PR.
- Still needed before merge: manual results on X11 and Wayland (xdotool getactivewindow right after the drop); CI cannot cover WM interaction, and SetInputFocus should be checked for focus flicker on mutter/KWin/Xfwm.

giachuvan:
Addressed all three. Time::CURRENT_TIME is 0 in x11rb, so a zero timestamp now skips activation instead of sending CurrentTime. SetInputFocus runs only when _NET_ACTIVE_WINDOW is absent from _NET_SUPPORTED (failed read assumes the WM supports it), with revert_to PARENT. mezon vendor edit markers are on request_dnd_activation and the Wayland Drop call site.

huylexuan44:
Re-reviewed `eef9bdc`. The three previous inline comments are addressed: X11 now uses the real drop timestamp, skips zero, gates `SetInputFocus` on `_NET_ACTIVE_WINDOW` support, and the Wayland vendor-edit markers are present.

One additional Wayland concern remains: `request_dnd_activation` writes `PendingActivation::Window` into the single global `pending_activation: Option<PendingActivation>`, while the `xdg_activation_token_v1::Done` handler is shared by all token objects and simply `take()`s whichever action is stored at that moment. If a URI, path, or another window activation overlaps the DND request, the first `Done` can execute the wrong action and the second sees `None`. Please associate the pending action with the token object/request id, serialize activation requests, or establish why overlap is impossible.

Before approval, please post manual results for both native X11 and native Wayland. `xdotool` is valid only for X11; on Wayland please verify actual keyboard focus or the `xdg_toplevel::Activated` state. Also please link the follow-up issue for Enter-before-import: the queue state machine was removed from this PR, so those product bugs remain intentionally out of scope.

PR #375, opened 2026-08-19
Channel Integration Webhook

giachuvan:


crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+                theme,
+            ))
+            .child(Self::render_divider(theme))
+            .when(loading && rows.is_empty(), |el| {
+                el.child(
+                    div()
+                        .text_sm()
+                        .text_color(theme.text_muted)
+                        .py_4()
+                        .child(mezon_i18n::t(locale, "common.loading")),
```
huylexuan44: common.loading does not exist in any locale file, so this renders the literal "common.loading". Use root.loading.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+        self.edit_input_sub = None;
+        self.channel_menu_open = false;
+        self.avatar_uploading = false;
+    }
+
+    fn cleanup_stale_edit_state(&mut self, cx: &App) {
+        let Some(expanded_id) = self.expanded_id.as_deref() else {
+            return;
+        };
+        if self.webhook_by_id(expanded_id, cx).is_none() {
```
huylexuan44, line 280: Expanded editor never resyncs on realtime change: save bar self-triggers and Save reverts the other user edit.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+                                    .size(px(15.0))
+                                    .text_color(theme.tokens.text_theme_primary),
+                            )
+                            .into_any_element()
+                    }),
+            )
+    }
+
+    fn render_channel_picker(&self, theme: &Theme, cx: &mut Context<Self>) -> impl IntoElement {
+        let mut options = self.channel_options(cx);
```
huylexuan44: channel_options() is rebuilt every render. Cache it in a field and refresh on the channel_list subscription.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+            cx.subscribe(&webhook_store, |this, _, event, cx| {
+                if matches!(
+                    event,
+                    WebhookEvent::ChannelWebhooksChanged { clan_id } if *clan_id == this.clan_id
+                ) {
+                    this.cleanup_stale_edit_state(cx);
+                    cx.notify();
+                }
+            }),
+            cx.subscribe(&members, |_, _, _, cx| cx.notify()),
```
huylexuan44: ClanMembersStore emits for every clan; filter on clan_id like the webhook subscription above.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+        loading: bool,
+        locale: &str,
+        theme: &Theme,
+        cx: &mut Context<Self>,
+    ) -> impl IntoElement {
+        let avatar_cache = shared_avatar_cache(cx);
+        let entity = cx.entity();
+        let mut cards = Vec::new();
+        for row in rows {
+            let created = format_i18n_full_date_from_seconds(row.create_time_seconds, locale);
```
huylexuan44: format_i18n_full_date_from_seconds and creator_name run per row per render. Precompute outside render.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+                        }),
+                ),
+        )
+        .when(expanded, |card| {
+            card.child(tab.render_edit_panel(&id, locale, theme, avatar_cache, cx))
+        })
+}
+
+fn webhook_count_label(count: usize, locale: &str) -> String {
+    let key = if count <= 1 {
```
huylexuan44, line 1403: React uses webhook_other for count >= 1, so 1 renders as "1 webhooks". This diverges at count == 1.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+                            img(crate::util::assets::EMPTY_WEBHOOK)
+                                .w(px(272.0))
+                                .h(px(145.0))
+                                .flex_none(),
+                        )
+                        .child(
+                            div()
+                                .font_weight(FontWeight::MEDIUM)
+                                .text_color(theme.tokens.text_theme_primary)
+                                .child(mezon_i18n::t(locale, "integrations.noWebhooks")),
```
huylexuan44, line 1270: integrations.noWebhooks was only updated in en and vi; the other 9 locales still say "click here to create one".

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+            id: webhook.id.clone(),
+            name: webhook.webhook_name.clone(),
+            avatar: webhook.avatar.clone(),
+            creator_id: webhook.creator_id,
+            create_time_seconds: webhook.create_time_seconds,
+        }
+    }
+}
+
+pub struct IntegrationsTab {
```
huylexuan44, line 57: Duplicates clan/settings/channel_webhook_tab.rs. React reuses one component via isClanSetting; extract shared parts.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+            return;
+        };
+        if self.webhook_by_id(expanded_id, cx).is_none() {
+            self.discard_edit_state();
+        }
+    }
+
+    fn toggle_expand(&mut self, id: String, window: &mut Window, cx: &mut Context<Self>) {
+        if self.expanded_id.as_ref() == Some(&id) {
+            self.discard_edit_state();
```
huylexuan44: Silently discards unsaved edits while the save bar is showing. Confirm or auto-reset instead.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -0,0 +1,1404 @@
...
+            .text_color(theme.text_primary)
+            .when(creating, |el| el.opacity(0.6))
+            .when(!creating, |el| {
+                el.cursor_pointer()
+                    .hover(|s| s.bg(theme.brand_hover))
+                    .on_click(cx.listener(|this, _, _, cx| {
+                        this.create_webhook(cx);
+                    }))
+            })
+            .when(creating, |el| {
```
huylexuan44, line 632: Spinner and the label both render while creating, with no gap between them.

crates/mezon-store/src/webhook.rs
```diff
@@ -132,6 +135,41 @@ impl WebhookStore {
...
+            cx.emit(WebhookEvent::ChannelWebhooksChanged { clan_id });
+            cx.notify();
+        }
+    }
+
+    fn refresh_after_lag(&mut self, cx: &mut Context<Self>) {
+        self.invalidate();
+        let clan_ids: Vec<ClanId> = self.channel_cache.iter().map(|(id, _)| *id).collect();
+        for clan_id in clan_ids {
+            self.fetch_channel_webhooks(clan_id, cx);
```
huylexuan44, line 169: invalidate() marks the clan cache stale but only channel webhooks are refetched here.

giachuvan:
1. Fixed. Loading copy now uses root.loading (the key exists in all locales). common.loading was a missing key and would have rendered as a literal.
2. Fixed. On realtime ChannelWebhooksChanged, if the expanded editor is still clean (draft matches the last synced baseline), we resync name/avatar/channel from the store so the save bar does not appear on its own and Save cannot overwrite another user’s edit. If the user already has local edits, we keep the draft.
3. Fixed. channel_options is cached on IntegrationsTab and rebuilt from the channel_list observe callback, not on every render.
4. Fixed. The ClanMembersStore subscription now filters on event.clan_id() == this.clan_id, same pattern as the webhook subscription.
5. Fixed. Date formatting and creator name are precomputed into WebhookRow.created_label when rows are rebuilt (webhook/members/locale changes), not inside the per-row render loop.
6. Intentionally not matching React here. React uses webhook_other for count >= 1, so 1 becomes “1 webhooks”. That is a React pluralization bug. Desktop keeps webhookCount for 0 and 1 (1 webhook) and webhook_other from 2. Tests lock this.
7. Already aligned before this round. All 11 locales use the empty-state string (EN: “You have no webhooks!”), not the old “click here to create one” CTA. Added a test so no locale regresses that CTA.
8. Out of scope for this PR. Channel settings IntegrationsTab is the Electron-matching implementation. Clan settings ChannelWebhookTab still diverges (dashed create box, painted URL, TabBar landing). Shared extraction / clan UI alignment will be a follow-up PR on clan settings, not a rewrite of this channel tab.
9. Fixed. Collapsing the row, switching to another webhook, or going back to the Integrations landing while dirty now shows a confirm: “Bỏ thay đổi?” (VI) / “Discard changes?” (EN). Cancel keeps the draft; confirm discards.
10. Fixed. The creating state still shows spinner + label, with gap_1() between them.
11. Fixed. refresh_after_lag still invalidate()s both caches, and now refetches clan webhooks for every clan in clan_cache as well as channel webhooks.

PR #377, opened 2026-08-19
Fix bug empty list channels when rejoin after leave

giachuvan:


crates/mezon-store/src/channel.rs
```diff
@@ -1008,6 +1012,7 @@ impl ChannelList {
     }
 
     pub fn load_for_clan(&mut self, clan_id: ClanId, cx: &mut Context<Self>) {
+        self.unforget_clan(clan_id);
```
huylexuan44: layout.rs:1115 loads on cache miss right after leave and clears this guard — gate unforget on the clan being in ClanList.

crates/mezon-store/src/channel.rs
```diff
@@ -748,6 +748,10 @@ impl ChannelList {
         }
     }
 
+    fn unforget_clan(&mut self, clan_id: ClanId) {
+        self.forgotten_clans.remove(&clan_id);
+    }
+
     pub fn forget_clan(&mut self, clan_id: ClanId, cx: &mut Context<Self>) {
```
huylexuan44, line 779: forget_clan strips user_channels but user_channels_loaded stays true, so a rejoined clan is missing from the palette.

giachuvan:
1. unforget_clan / load_for_clan
Agreed.
After leave, layout.rs still hits load_for_clan on a cache miss. That used to call unforget_clan unconditionally and drop the forgotten guard, so a leftover route could refetch and restore a clan the user just left.
unforget_clan now only clears the guard if the clan is still in ClanList (real rejoin / still a member). load_for_clan, load_structure_for_clan, and refresh_clan return early if the clan stays forgotten.
A leftover load_for_clan after leave no longer unforgets. Rejoin first puts the clan back on ClanList, then load_for_clan is allowed to fetch again.
2. user_channels_loaded after forget_clan
Agreed.
forget_clan stripped that clan’s rows from user_channels but left user_channels_loaded = true. ensure_user_channels_loaded then skipped list_channel_by_user_id, so the command palette never picked the clan back up after rejoin.
forget_clan now sets user_channels_loaded = false so the next palette / hashtag open refetches the user-channel list.

crates/mezon-store/src/channel.rs
```diff
@@ -1008,6 +1016,10 @@ impl ChannelList {
     }
 
     pub fn load_for_clan(&mut self, clan_id: ClanId, cx: &mut Context<Self>) {
+        self.unforget_clan(clan_id, cx);
+        if self.forgotten_clans.contains(&clan_id) {
```
huylexuan44, line 1042: Regression: invite accept calls select_clan before ClanList reload lands, so this returns and nothing retries — rejoined clan stays empty until you switch clans. update_clans emits no ActiveClanChanged when active is unchanged.

crates/mezon-store/src/channel.rs
```diff
@@ -748,8 +748,16 @@ impl ChannelList {
...
+    fn unforget_clan(&mut self, clan_id: ClanId, cx: &App) {
+        if ClanList::global(cx).read(cx).clan(clan_id).is_none() {
+            return;
+        }
+        self.forgotten_clans.remove(&clan_id);
+    }
+
     pub fn forget_clan(&mut self, clan_id: ClanId, cx: &mut Context<Self>) {
         self.forgotten_clans.insert(clan_id);
+        self.user_channels_loaded = false;
```
huylexuan44: Reset at leave is too early: opening the palette between leave and rejoin sets user_channels_loaded back to true, so the rejoined clan is still missing. Reset it on rejoin instead.

crates/mezon-store/src/channel.rs
```diff
@@ -748,8 +748,16 @@ impl ChannelList {
...
 
+    fn unforget_clan(&mut self, clan_id: ClanId, cx: &App) {
+        if ClanList::global(cx).read(cx).clan(clan_id).is_none() {
+            return;
+        }
+        self.forgotten_clans.remove(&clan_id);
+    }
+
     pub fn forget_clan(&mut self, clan_id: ClanId, cx: &mut Context<Self>) {
         self.forgotten_clans.insert(clan_id);
```
huylexuan44, line 780: user_channels_loading is not cleared here, so an in-flight list_channel_by_user_id resolves after leave and restores the left clan channels, then sets loaded back to true.

crates/mezon-store/src/channel.rs
```diff
@@ -748,8 +748,16 @@ impl ChannelList {
         }
     }
 
+    fn unforget_clan(&mut self, clan_id: ClanId, cx: &App) {
+        if ClanList::global(cx).read(cx).clan(clan_id).is_none() {
+            return;
+        }
+        self.forgotten_clans.remove(&clan_id);
```
huylexuan44, line 759: Suggest emitting ClanEvent::Joined from update_clans and handling it here: clear forgotten_clans, reset user_channels_loaded, reload the active clan. Fixes both gaps at once.

giachuvan:
eave clan emptied the channel list after rejoin (until app restart). Desktop leave already called RemoveClanUsers correctly — it does not delete the clan.
Cause: On leave we forget the clan (drop channel cache + set a guard so an in-flight structure fetch cannot paint the clan you just left). That guard was only cleared on logout/kill, so rejoin fetched channels then discarded them.
Fix, tied to UX:
Leave — forget the clan. Do not reset user_channels_loaded (opening the command palette between leave and rejoin would refetch and mark it loaded again, so the rejoined clan stayed missing). Do bump a generation and clear user_channels_loading so a list_channel_by_user_id that resolves after leave cannot restore the left clan.
Rejoin / accept invite — select_clan runs before ClanList reload, so we must not unforget yet (ClanList does not contain the clan). When the list lands, update_clans emits ClanEvent::Joined. That handler unforgets, resets user_channels_loaded, and reloads the active clan. Switching away and back is not required.
This matches the review notes: unforget on Joined (not on a premature load_for_clan), reset user-channel loaded state on rejoin, and drop in-flight user-channel fetches on leave.

PR #378, opened 2026-08-19
Fix bug render public thread not join in sidebar

giachuvan:


crates/mezon-store/src/channel.rs
```diff
@@ -2978,6 +2978,9 @@ impl ChannelList {
                 if self.is_locally_archived(channel_id) {
                     return;
                 }
+                if e.parent_id != 0 {
```
huylexuan44, line 2981: React also gates this on channel_private == 0; private top-level channels still land in every member sidebar.

crates/mezon-store/src/channel.rs
```diff
@@ -2978,6 +2978,9 @@ impl ChannelList {
...
+                    return;
```
huylexuan44, line 2982: parent_id on line 2998 can now only be None — simplify it, the insert_channel parent branch is unreachable here.

crates/mezon-store/src/channel.rs
```diff
@@ -6911,6 +6928,45 @@ mod tests {
...
+                channels.handle_event(
+                    &channel_created_event(42, 1, crate::threads::CHANNEL_TYPE_THREAD as i32),
+                    cx,
+                );
+
+                assert!(
+                    !channels.channel_in_clan(ClanId(1), ChannelId(42)),
+                    "ChannelCreated must not insert a thread the viewer has not joined"
+                );
+                assert!(channels.user_channel(ChannelId(42)).is_none());
```
huylexuan44: Passes on develop too — this arm never wrote user_channels, so only the assert above guards the fix.

giachuvan:
1. Gate on channel_private == 0
We are not adding that check.
React only calls createChannelSocket for top-level public channels. That does not mean desktop must do the same.
The server does not fan out ChannelCreated for a private top-level channel to the clan. It only sends it on the creator’s notification stream. Other members never receive that event, so a private channel cannot land on every member’s sidebar from this path. They only see it after UserChannelAdded (invite / join).
Gating channel_private == 0 here would only stop the creator from inserting their own new private channel on ChannelCreated. That copies a React detail; it is not a desktop bug.
2. parent_id after the early return
Agreed — after if e.parent_id != 0 { return }, parent_id is always none.
Updated the constructed Channel to parent_id: None. The insert_channel “place under parent” branch is unreachable on this path.
3. user_channel assert
Agreed — that assert does not guard the fix.
ChannelCreated never wrote user_channels, so user_channel(42).is_none() already passed on develop. The regression check is !channel_in_clan(ClanId(1), ChannelId(42)) (old code inserted the thread into the clan sidebar cache). Removed the unused assert.

PR #381, opened 2026-08-20
Reimplement send flower effect

giachuvan:


huylexuan44 (commented):
Review notes for the flower-effect reimplementation.

crates/mezon-ui/src/chat/flower_celebration.rs
```diff
@@ -0,0 +1,1531 @@
...
+            p.vy += p.fall_gravity;
+            p.vy *= 0.986;
+        }
+        p.flutter += p.flutter_speed;
+        let flutter_x = p.flutter.sin()
+            * match p.shape {
+                ParticleShape::Ribbon | ParticleShape::Streamer => 3.2,
+                _ => 1.2,
+            };
+        p.x += p.vx + flutter_x;
```
huylexuan44: [P2] Particle physics is advanced once per rendered frame here instead of by elapsed time. On a 120 Hz display, position, rotation, flutter, and life decay progress roughly twice as fast as at 60 Hz; at low FPS the four-second store timer can remove the overlay before the particles reach the same state. Please pass the frame delta into this function and scale these updates by a normalized timestep (including exponential scaling for drag).

crates/mezon-store/src/voice.rs
```diff
@@ -1219,14 +1217,24 @@ impl VoiceStore {
...
             .unwrap_or_default();
-        let label = SharedString::from(
-            mezon_i18n::t(&locale, "channelVoice.giveFlowerGiven")
-                .replace("{{giver}}", &giver_name)
-                .replace("{{receiver}}", &receiver_name),
-        );
+        let local_is_receiver = self.local_user_id().as_deref() == Some(receiver_id.as_str());
+        let label = SharedString::from(if local_is_receiver {
+            format!(
+                "{} {}",
```
huylexuan44: [P2] Concatenating translated fragments with names in this fixed order breaks several locales. For example, Japanese and Korean need the giver before the `received` fragment, while the German fragment already ends with `erhalten`; several `someoneReceived` translations also say `to you` before another receiver is appended. Please use complete localized templates with `{{giver}}` / `{{receiver}}` placeholders, as `channelVoice.giveFlowerGiven` already does.

crates/mezon-ui/src/chat/flower_celebration.rs
```diff
@@ -0,0 +1,1531 @@
...
+    let pad_x = px(12.);
+    let pad_y = px(8.);
+    let mut run = window.text_style().to_run(label.len());
+    run.color = white().opacity(slide);
+    run.font.weight = FontWeight::SEMIBOLD;
+    let line =
+        window
+            .text_system()
+            .shape_line(label.clone(), font_size, std::slice::from_ref(&run), None);
+    let pill_w = line.width() + pad_x * 2.;
```
huylexuan44: [P2] `pill_w` is unbounded because this is a single shaped line. A long display name or translation can make it wider than the voice pane, producing a negative `x` and clipped text. The previous UI capped the caption at 85% width; please cap the pill to the available bounds and wrap or truncate the label.

PR #390, opened 2026-08-21
Fix jump inbox items from topic

giachuvan:


crates/mezon-store/src/messages.rs
```diff
@@ -3320,6 +3337,107 @@ impl MessagesStore {
...
+        &mut self,
+        topic_id: i64,
+        message_id: MessageId,
+        cx: &mut Context<Self>,
+    ) {
+        if message_id.is_optimistic() {
+            return;
+        }
+        let topic_key = ChannelId(topic_id);
+        if !topic_jump_needs_around_fetch(
```
huylexuan44: This early return leaves active_topic_parent unset. set_active_topic() clears it when the panel is opened, so reopening a retained/cached topic and hitting this branch makes load_more_topic() and load_more_topic_bottom() return false. Please resolve/store the parent channel before the cache-hit check.

crates/mezon-store/src/messages.rs
```diff
@@ -3320,6 +3337,107 @@ impl MessagesStore {
...
+                        message_id = anchor,
+                        "request_topic_jump: target not in AROUND window"
+                    );
+                    this.set_channel(topic_key, window);
+                    cx.emit(MessagesEvent::TopicUpdated { topic_id });
+                    cx.notify();
+                    return;
+                }
+                recompute_message_grouping(&mut window);
+                this.set_channel(topic_key, window);
```
huylexuan44: set_channel() seeds last_message_by_channel from window.last() when no tail is known. For a fresh AROUND fetch this is only the end of the partial window, not necessarily the real topic tail, so topic_has_more_bottom() becomes false and replies newer than the window cannot be loaded. Please preserve the real tail from topic detail or track the newer edge separately.

crates/mezon-store/src/topics.rs
```diff
@@ -647,6 +662,66 @@ impl TopicsStore {
...
+            };
+            let Ok(origin_id) = detail.message_id.parse::<i64>() else {
+                return;
+            };
+            if origin_id <= 0 {
+                return;
+            }
+            let origin_id = MessageId(origin_id);
+            let _ = this.update(cx, |this, cx| {
+                this.pending_inbox_jump = Some(PendingInboxTopicJump {
```
huylexuan44: pending_inbox_jump is only cleared by a matching JumpTo or reset. If navigation changes before the parent jump lands, this request can remain indefinitely; a later matching event can open an old topic, and overlapping detail requests can let an older response overwrite the latest click. Please add a generation/route guard and clear or expire stale requests.

crates/mezon-store/src/messages.rs
```diff
@@ -1986,10 +1994,28 @@ impl MessagesStore {
...
+        for (bucket_id, channel) in self.cache.iter() {
+            if !channel.messages.contains_id(message_id) {
+                continue;
+            }
+            if let Some(msg) = channel.messages.get_by_id(message_id)
+                && let Some(topic) = msg.topic_id.filter(|id| id.get() != 0)
+            {
+                return Some(topic.get());
+            }
+            if self.active_channel_id != Some(*bucket_id) {
```
huylexuan44: cache holds plain channels too — this returns a channel id as topic_id and the inbox jump dies. Verify bucket is a topic.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -1674,17 +1674,7 @@ impl ChannelMessages {
                     }
                 }
                 MessagesEvent::JumpTo { message_id } => {
```
huylexuan44: Topic jumps reuse JumpTo, so this arms pending_jump with a reply never in this viewport. Guard like is_topic_jump_target.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -4575,13 +4623,14 @@ impl Render for ChannelMessages {
...
 
-        if let Some(target) = self.pending_jump.take()
+        if let Some(target) = self.pending_jump
             && let Some(pos) = store
                 .read(cx)
                 .viewport_messages()
                 .iter()
                 .position(|m| m.id == target)
         {
+            self.pending_jump = None;
```
huylexuan44: Dropping take() leaves an unservable jump armed forever and rescans the viewport every frame.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -233,8 +236,14 @@ impl InboxPopoverPanel {
     }
 
     fn sync_from_store(&mut self, cx: &App) {
```
huylexuan44: pin_top on any first-item change scrolls to top, so an incoming notification yanks the list away from the reader.

crates/mezon-client/src/inbox.rs
```diff
@@ -574,14 +587,16 @@ fn parse_notification_content(bytes: &[u8]) -> Option<InboxMessagePreview> {
         }
         return parse_message_preview_json(bytes);
     }
+    if let Ok(fcm) = api::DirectFcmProto::decode(bytes)
```
huylexuan44: New message_id gate drops the whole preview when FCM lacks field 19; content used to render. Confirm the server sets it.

crates/mezon-store/src/inbox.rs
```diff
@@ -409,35 +491,19 @@ impl InboxStore {
...
-            bucket.items.insert(0, notification);
-            if bucket.items.len() > REALTIME_BUCKET_CAP {
-                bucket.items.truncate(REALTIME_BUCKET_CAP);
-            }
-            changed = true;
-        }
-        if changed {
-            self.emit_updated(cx);
-            cx.notify();
+            self.prepend_local(
```
huylexuan44: prepend_local also fires Updated when it only reorders a dup, and the fetched_at gate is gone. Confirm both intended.

crates/mezon-store/src/inbox.rs
```diff
@@ -305,13 +366,35 @@ impl InboxStore {
...
+    }
+
     fn apply_fetch_result(
         &mut self,
         category: InboxCategory,
         generation: u64,
         result: Result<Vec<InboxNotification>, anyhow::Error>,
         cx: &mut Context<Self>,
     ) {
+        let remembered = self.topic_by_message.clone();
```
huylexuan44: topic_by_message is cloned on every page fetch and only cleared on reset, so it grows for the whole session.

crates/mezon-store/src/messages.rs
```diff
@@ -3309,11 +3448,115 @@ impl MessagesStore {
...
+                    let start = target
+                        .saturating_sub(half)
+                        .min(window.len() - MAX_MESSAGES_PER_CHANNEL);
+                    window = window[start..start + MAX_MESSAGES_PER_CHANNEL].to_vec();
                 }
+                let found = window.iter().any(|m| m.id == message_id);
+                if !found {
+                    tracing::warn!(
+                        message_id = anchor,
+                        "request_topic_jump: target not in AROUND window"
```
huylexuan44, line 3704: Unlike jump_to_message, this clobbers the topic buffer with a window missing the target. Leave the buffer alone.

crates/mezon-store/src/messages.rs
```diff
@@ -3038,19 +3064,102 @@ impl MessagesStore {
...
+            .topic_id
+            .map(|topic| topic.get())
+            .filter(|id| *id != 0)
+            .or_else(|| {
+                self.active_topic_id
+                    .filter(|topic| *topic == storage_id)
+                    .map(|topic| topic.get())
+            })
+            .or_else(|| {
+                (storage_id != channel_id && storage_id.get() != 0).then_some(storage_id.get())
```
huylexuan44: Redundant: reaction_storage_channel only returns the topic or the active channel, so this repeats the previous or_else.

crates/mezon-store/src/messages.rs
```diff
@@ -62,6 +63,10 @@ const DIRECTION_AFTER: i32 = 1;
 /// `Direction_Mode.AROUND_TIMESTAMP` — fetch a window centered on a message
 /// (used by jump-to-message when the target is not loaded).
 const DIRECTION_AROUND: i32 = 2;
+
+fn topic_jump_needs_around_fetch(found_in_cache: bool) -> bool {
```
huylexuan44: This is just a negation and reads as a double negative at the call site. Inline it and drop the tautological test.

crates/mezon-store/src/messages.rs
```diff
@@ -3038,19 +3064,102 @@ impl MessagesStore {
...
+                .filter(|msg| msg.id == message_id)
+                .cloned()
+        })
+    }
+
+    async fn save_message_2_inbox(
+        api: &Arc<AppApi>,
+        request: mezon_proto::api::Message2InboxRequest,
+    ) -> bool {
+        let Err(error) = api.create_message_2_inbox(request.clone()).await else {
```
huylexuan44, line 3238: INVALID_ARGUMENT is not avatar-specific. Note which server behaviour this retry works around.

crates/mezon-store/src/voice.rs
```diff
@@ -2096,16 +2096,15 @@ impl VoiceStore {
         }
         let channel_id = channel_id.to_string();
         let clan_id = clan_id.to_string();
-        let room_name = self.room_name.clone();
         let api = self.api.clone();
         cx.spawn(async move |this, cx| {
             let result = match action {
                 ModerationAction::Mute => {
-                    api.mute_participant_mezon_meet(&channel_id, &clan_id, &room_name, &identity)
+                    api.mute_participant_mezon_meet(&channel_id, &clan_id, &identity)
```
huylexuan44: identity must now parse as i64, so a non-numeric participant fails locally; the room_name guard above is dead.

giachuvan:
1. request_topic_jump cache-hit early return (messages.rs:3478)
Valid. set_active_topic() clears active_topic_parent, and the cache-hit branch returned before restoring it. load_more_topic() / load_more_topic_bottom() require active_topic_parent and would silently no-op (e.g. after topic creation via set_active_topic without a fresh fetch_topic_messages).
Fix: resolve and store the parent channel before the cache-hit return, same as the AROUND-fetch path.

2. topic_id_for_message treating any cached bucket as a topic (messages.rs:2009)
Valid. The fallback return Some(bucket_id.get()) could return a plain channel id from an unrelated LRU cache entry, breaking inbox → topic jumps.
Fix: only treat a bucket as a topic when active_topic_id matches or TopicsStore::topic_by_id confirms it.

3 & 4. JumpTo on channel view + stuck pending_jump (channel_messages.rs:1676, 4633)
Valid. Topic replies emit JumpTo, but the channel list (not the topic box) subscribed unconditionally via begin_highlight, arming pending_jump for a message never in that viewport. Because pending_jump was only cleared on a successful scroll, it stayed set and rescanned every frame.
Fix: guard JumpTo with is_topic_jump_target (topic box already had this); clear pending_jump when the target belongs to the topic panel.

5. Inbox scroll jump on new notification (panel.rs:236)
Valid. sync_from_store passed pin_top=true whenever the first item id changed, so a prepended notification scrolled the list to the top while the user was reading lower items.
Fix: use sync_list_state(false) on store-driven updates; explicit scroll-to-top remains on tab switch only.

9. Failed AROUND fetch clobbering topic buffer (messages.rs:3544)
Valid. Unlike jump_to_message, a missing anchor still called replace_channel with a partial window, wiping good cached data.
Fix: match jump_to_message — log and return without mutating the bucket.

11. topic_jump_needs_around_fetch helper (messages.rs:67)
Valid style nit. Removed the tautological wrapper and inlined the cache check at the call site.

12. INVALID_ARGUMENT retry logging (messages.rs:3095)
Partially valid. InvalidArgument is not avatar-specific in general, but for CreateMessage2Inbox the server only returns it from ValidateImageURLSecure on the avatar field (mezon-api/server/api_pin_message.go:26–28). Attachment URL failures use the same code but are a separate path.
Fix: clarified the warn log to name ValidateImageURLSecure explicitly.

Not changed — reviewer overstated or behavior is intentional
6. FCM message_id > 0 gate (inbox.rs:590)
Disagree as a bug; gate is intentional.

CreateMessage2Inbox always sets DirectFcmProto.message_id server-side (api_pin_message.go:67).
The gate avoids mis-decoding non-FCM protobuf payloads as DirectFcmProto.
If FCM decode fails the gate, inbox_notification_from_api still has a string/JSON fallback (lines 858–871).
Legacy notifications with FCM bytes but message_id == 0 could lose the structured preview — worth monitoring, not a regression from this PR. No code change unless we want a legacy fallback.
7. prepend_local emitting Updated on dup reorder + fetched_at invalidation (inbox.rs:501)
Both intentional.

Reordering a pending duplicate to the top must notify the panel so server/reconcile state is visible.
refresh_category() after add_to_inbox deliberately clears fetched_at / server_loaded so the next open reconciles with the server — this fixed “only first add-to-inbox item shows until restart.”
8. topic_by_message.clone() on every fetch (inbox.rs:397)
Acknowledged, low severity. The map only grows for messages we've actually resolved a topic for; it's cleared on inbox reset. A full-session unbounded growth is theoretically possible but bounded by user activity. Clone cost per fetch is small. Can add a cap later if needed; not blocking.

10. Redundant topic_id fallback in add_to_inbox (messages.rs:3157)
Mostly redundant, not wrong. The third or_else duplicates cases already covered when active_topic_id == storage_id. Harmless defensive fallback for edge cases where reaction_storage_channel returns the topic bucket id. Optional cleanup, no functional bug.

13. Voice moderation identity / room_name guard (voice.rs:2103)
Reviewer partially wrong.

room_name.is_empty() is not dead — it guards “are we connected to a meet room?” before calling the API. The API no longer takes room_name, but the connection guard is still valid.
LiveKit participant identity is the numeric user id in this codebase (see identity.parse::<i64>() elsewhere in voice.rs:3103 and the comment at line 354). The proto change from username/room_name → user_id matches that.
Non-numeric identities would have failed the old username-based API too in practice.

crates/mezon-store/src/inbox.rs
```diff
@@ -320,20 +465,28 @@ impl InboxStore {
...
+                    let existing_ids: HashSet<String> =
                         bucket.items.iter().map(|n| n.id.clone()).collect();
-                    bucket
+                    let existing_message_ids: HashSet<String> = bucket
                         .items
-                        .extend(items.into_iter().filter(|n| !existing.contains(&n.id)));
+                        .iter()
+                        .filter_map(|n| n.effective_message_id())
+                        .collect();
+                    bucket.items.extend(items.into_iter().filter(|n| {
```
huylexuan44: Dedupe on message_id drops cross-channel mentions: message ids are only unique per channel. Key on notification id.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -1674,17 +1674,10 @@ impl ChannelMessages {
...
-                    this._highlight_timer = Some(cx.spawn(async move |this, cx| {
-                        cx.background_executor()
-                            .timer(Duration::from_millis(1500))
-                            .await;
-                        let _ = this.update(cx, |this, cx| {
-                            this.highlight_id = None;
-                            cx.notify();
-                        });
-                    }));
+                    if this.is_topic_jump_target(*message_id, cx) {
```
huylexuan44, line 1684: Bare message-id match swallows parent-channel jumps while the topic panel scrolls elsewhere. Compare topic_id too.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -3490,6 +3558,34 @@ impl ChannelMessages {
         self.sync_channel_seen_when_focused(app_focused, cx);
     }
 
+    fn sync_topic_seen(&mut self, cx: &mut Context<Self>) {
```
huylexuan44, line 3852: Missing the has_more_bottom guard sync_channel_seen_when_focused has: end of a partial AROUND window marks all read.

crates/mezon-store/src/inbox.rs
```diff
@@ -267,28 +305,89 @@ impl InboxStore {
...
-        {
-            return;
+        let incoming_message_id = notification.effective_message_id();
+        if let Some(pos) = bucket.items.iter().position(|existing| {
+            existing.id == notification.id
+                || incoming_message_id.as_deref().is_some_and(|message_id| {
+                    existing.effective_message_id().as_deref() == Some(message_id)
+                })
+        }) {
+            let mut existing = bucket.items.remove(pos);
```
huylexuan44: Duplicate keeps the pending- stub and drops the server row, so delete never fires. Keep the server notification.

crates/mezon-client/src/transport.rs
```diff
@@ -5211,6 +5218,7 @@ impl MezonTransport {
             references: Vec::new(),
             reactions: Vec::new(),
             entity_mentions: Vec::new(),
+            topic_id: 0,
```
huylexuan44: topic_id is in scope but hard-coded to 0, so a just-sent topic reply is the only message with topic_id None.

crates/mezon-store/src/messages.rs
```diff
@@ -1393,6 +1399,48 @@ impl MessagesStore {
...
+        let pending = PendingLastSeen {
+            clan_id,
+            channel_id: topic_id,
+            message_id,
+            create_time,
+            mode: STREAM_MODE_THREAD,
+            badge_count: 0,
+        };
+        self.set_last_read_message(topic_id, message_id);
+        self.pending_last_seen = Some(pending);
```
huylexuan44: Overwrites the shared pending_last_seen slot and cancels its timer, so the channel last-seen write is lost.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -740,6 +760,12 @@ fn schedule_inbox_jump(
     channel_id: ChannelId,
     message_id: MessageId,
 ) {
+    if let Route::Channel { clan_id, .. } = &route {
```
huylexuan44: apply_read runs before navigate, so seed_last_read_from_channel copies a faked newest id and the divider never draws.

crates/mezon-store/src/messages.rs
```diff
@@ -3597,7 +3890,7 @@ impl MessagesStore {
                 channel.messages.push_grouped(msg);
             }
         } else {
-            self.set_channel(topic_key, vec![msg]);
+            self.replace_channel(topic_key, vec![msg], false);
```
huylexuan44: has_more hard-coded false: a bucket seeded by one realtime reply can never page older replies. Use has_more_from_oldest.

crates/mezon-store/src/badge.rs
```diff
@@ -545,33 +545,44 @@ impl BadgeService {
...
+            .filter(|ts| *ts > 0)
+            .map(i64::from)
+            .unwrap_or_else(|| i64::from(notification.create_time_seconds));
+        tracing::debug!(
+            message_id = message_id.get(),
+            msg_time,
+            content_len = notif.content.len(),
+            "badge: parsed notification content"
+        );
+        let badge_channel = notification
```
huylexuan44: badge_channel uses effective_topic_id but the guard at 626 still reads notif.topic_id, so the clan badge never clears.

crates/mezon-store/src/messages.rs
```diff
@@ -3309,11 +3498,115 @@ impl MessagesStore {
...
+                if !found {
+                    tracing::warn!(
+                        message_id = anchor,
+                        "request_topic_jump: target not in AROUND window"
+                    );
+                    cx.notify();
+                    return;
                 }
+                recompute_message_grouping(&mut window);
+                let has_more = window.len() >= TOPIC_FULL_PAGE_LEN;
```
huylexuan44: AROUND windows are centered, so this length test kills load-older near a topic tail. Use has_more_from_oldest.

crates/mezon-store/src/messages.rs
```diff
@@ -3309,11 +3498,115 @@ impl MessagesStore {
...
+                if window.len() > MAX_MESSAGES_PER_CHANNEL {
+                    let target = window.iter().position(|m| m.id == message_id).unwrap_or(0);
+                    let half = MAX_MESSAGES_PER_CHANNEL / 2;
+                    let start = target
+                        .saturating_sub(half)
+                        .min(window.len() - MAX_MESSAGES_PER_CHANNEL);
+                    window = window[start..start + MAX_MESSAGES_PER_CHANNEL].to_vec();
+                }
+                let found = window.iter().any(|m| m.id == message_id);
+                if !found {
```
huylexuan44, line 3701: The not-found and error paths return without clearing pending_topic_jump, so the failing request re-fires forever.

crates/mezon-store/src/topics.rs
```diff
@@ -647,6 +676,103 @@ impl TopicsStore {
...
+        };
+        let now = Instant::now();
+        let Some(pending) = self.pending_inbox_jump.as_ref() else {
+            return;
+        };
+        if pending.is_expired(now) {
+            self.pending_inbox_jump = None;
+            return;
+        }
+        if !pending.matches(*message_id, now) {
```
huylexuan44: matches() ignores channel_id and topic_id and expiry is only checked lazily, so an unrelated JumpTo can hijack this.

crates/mezon-store/src/messages.rs
```diff
@@ -3038,19 +3112,104 @@ impl MessagesStore {
...
+        let Some(msg) = self.inbox_source_message(message_id, cx) else {
+            tracing::warn!(
+                message_id = message_id.get(),
+                storage_id = storage_id.get(),
+                channel_id = channel_id.get(),
+                "add_to_inbox: message not in cache"
+            );
             return;
         };
+        let topic_id = msg
```
huylexuan44: msg.topic_id is set on topic origin messages too, so marking one saves a row that opens the panel instead of jumping.

crates/mezon-store/src/inbox.rs
```diff
@@ -267,28 +305,89 @@ impl InboxStore {
...
+            return;
+        };
+        self.topic_by_message.insert(message_id, topic_id);
+    }
+
+    pub fn remembered_topic_id(&self, message_id: &str) -> Option<&str> {
+        self.topic_by_message.get(message_id).map(String::as_str)
+    }
+
+    fn apply_remembered_topic_ids(
```
huylexuan44, line 355: Keyed by message_id, which is only unique per channel, and replayed onto every page: stamps topics on unrelated rows.

crates/mezon-store/src/inbox.rs
```diff
@@ -472,6 +621,55 @@ impl InboxStore {
...
+            | MessageCode::UpdateEphemeralMsg
+            | MessageCode::DeleteEphemeralMsg
+            | MessageCode::CreatePin
+            | MessageCode::Typing
+            | MessageCode::Indicator
+            | MessageCode::Welcome
+    )
+}
+
+fn channel_message_is_mention_for_inbox(message: &ChannelMessage, cx: &App) -> bool {
```
huylexuan44: Full mention decode for every message in every clan, with none of the mute or membership filtering the server applies.

crates/mezon-store/src/inbox.rs
```diff
@@ -190,13 +207,25 @@ impl InboxStore {
         if bucket.loading {
             return false;
         }
+        if !bucket.server_loaded
+            && (!bucket.items.is_empty()
```
huylexuan44: Regression: realtime prepend now creates the bucket, then this blocks the first server fetch for the whole session. Drop it; merge_server_page already handles local items.

crates/mezon-store/src/inbox.rs
```diff
@@ -511,17 +751,133 @@ mod tests {
...
             ..CategoryBucket::default()
         };
-        assert!(InboxStore::should_fetch_initial(Some(&local_only)));
+        assert!(!InboxStore::should_fetch_initial(Some(&local_only)));
         assert!(InboxStore::should_fetch_initial(None));
     }
 
     #[test]
-    fn skip_initial_fetch_when_server_page_is_fresh() {
+    fn skip_initial_fetch_when_socket_items_exist_before_server_page() {
```
huylexuan44: This test pins the regression above; it should expect a fetch again.

crates/mezon-store/src/inbox.rs
```diff
@@ -109,6 +123,9 @@ impl InboxStore {
             dispatch.on(RealtimeKind::Notifications, &entity, |this, event, cx| {
                 this.handle_event(event, cx);
             });
+            dispatch.on(RealtimeKind::ChannelMessage, &entity, |this, event, cx| {
```
huylexuan44: BadgeService and TopicBadgeStore already call note_mention for this message, so each mention decodes and emits Updated 3 times. Keep one path.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -102,23 +102,34 @@ impl InboxPopoverPanel {
...
-            });
-        }
-
         let _inbox_sub = cx.subscribe(&inbox_store, |this, _, event, cx| {
             let InboxEvent::Updated { .. } = event;
             if this.tab == InboxTab::Topics {
                 return;
             }
-            this.sync_from_store(cx);
+            this.sync_from_store(cx, true);
```
huylexuan44: pin_top on every Updated resets scroll to top; with 3 Updated per mention the reader gets yanked. Pin only on open.

crates/mezon-store/src/messages.rs
```diff
@@ -9292,6 +9613,43 @@ mod tests {
...
+            let clan = ClanId(1);
+            let parent = ChannelId(10);
+            let topic = ChannelId(77);
+
+            store.update(cx, |store, cx| {
+                store.activate(clan, parent, true, false, 1, 2, cx);
+                store.set_active_topic(Some(topic.get()), cx);
+                store.append_topic_message(
+                    topic.get(),
+                    api_page(&[5]).messages.into_iter().next().unwrap(),
```
huylexuan44, line 9685: append_topic_message takes 6 args since d00d372c; add Vec::new(), true so lib tests compile (CI Lint and Test are red).

crates/mezon-store/src/inbox.rs
```diff
@@ -267,28 +321,114 @@ impl InboxStore {
...
+            {
+                existing
+            } else {
+                notification
+            };
+            if merged.topic_id.is_none() {
+                merged.topic_id = existing_topic_id.or(incoming_topic_id);
+            }
+            if let Some(preview) = merged.message.as_mut() {
+                if preview.topic_id.is_none() {
```
huylexuan44: clippy collapsible_if fails Lint: fold this check into the let-chain above.

crates/mezon-ui/src/chat/channel_header.rs
```diff
@@ -1018,6 +1018,9 @@ impl ChatHeader {
             &mezon_store::NotificationSettingStore::global(cx),
             |_, _, cx| cx.notify(),
         );
+        let _clan_observe = cx.observe(&mezon_store::ClanList::global(cx), |_, _, cx| cx.notify());
```
huylexuan44: Unconditional notify on every ClanList change bypasses the inbox_badge gate below; gate it like ChatLayout::inbox_display_changed.

crates/mezon-store/src/messages.rs
```diff
@@ -2032,10 +2089,36 @@ impl MessagesStore {
         true
     }
 
-    /// Jump to a message (cf. React `jumpToMessage`, used by reply previews).
-    /// If the target is already in the buffer, emit [`MessagesEvent::JumpTo`] so
-    /// the UI scrolls to it. Otherwise fetch a window centered on it
-    /// (`AROUND_TIMESTAMP`), replace the buffer, and emit `Reset` then `JumpTo`.
+    pub fn topic_id_for_message(&self, message_id: MessageId, cx: &App) -> Option<i64> {
```
huylexuan44: topic_id_for_message has no callers. Delete it.

crates/mezon-store/src/inbox.rs
```diff
@@ -305,13 +445,64 @@ impl InboxStore {
...
+
     fn apply_fetch_result(
         &mut self,
         category: InboxCategory,
         generation: u64,
         result: Result<Vec<InboxNotification>, anyhow::Error>,
+        is_first_page: bool,
         cx: &mut Context<Self>,
     ) {
+        let remembered = self.topic_by_message.clone();
```
huylexuan44: Still cloned on every page and never pruned; the resolved thread on this changed nothing.

crates/mezon-store/src/messages.rs
```diff
@@ -9233,6 +9549,11 @@ mod tests {
         assert!(plan.is_empty());
     }
 
+    #[test]
+    fn topic_jump_skips_around_fetch_when_reply_is_cached() {
+        assert_eq!(DIRECTION_AROUND, 2);
```
huylexuan44: This asserts a constant, not the behaviour in the name. Delete it or test request_topic_jump on a cached reply.

crates/mezon-store/src/topic_badges.rs
```diff
@@ -452,4 +492,20 @@ mod tests {
             .sum();
         assert_eq!(total, 5);
     }
+
+    #[test]
+    fn reset_topic_removes_entry() {
```
huylexuan44: This tests HashMap::remove, not TopicBadgeStore. Call clear_topic instead.

crates/mezon-store/src/inbox.rs
```diff
@@ -511,17 +751,133 @@ mod tests {
...
+            reset_generation: 0,
+            _conn_watch: Task::ready(()),
+        };
+        let mention = sample_notification("7");
+        assert!(store.should_skip_realtime(&mention));
+        assert!(store.should_prepend_realtime(&mention));
+    }
+
+    #[test]
+    fn schedule_refresh_queues_when_already_loading() {
```
huylexuan44: Sets a bool and asserts it; schedule_refresh_category is never called. Test the store method.

crates/mezon-client/src/inbox.rs
```diff
@@ -795,6 +836,24 @@ pub fn inbox_notification_from_marked_message(
     }
 }
 
+pub fn inbox_notification_from_channel_mention(message: &api::ChannelMessage) -> InboxNotification {
+    let preview = preview_from_channel_message(message.clone());
+    InboxNotification {
+        id: pending_inbox_notification_id(message.message_id),
```
huylexuan44: Message ids are per-channel, so pending-{message_id} collides across channels. Add channel_id to the key.

crates/mezon-store/src/inbox.rs
```diff
@@ -267,28 +305,114 @@ impl InboxStore {
...
-        }
-        if let Some(message_id) = notification.effective_message_id()
-            && bucket.items.iter().any(|existing| {
-                existing.effective_message_id().as_deref() == Some(message_id.as_str())
-            })
-        {
-            return;
+        let incoming_message_id = notification.effective_message_id();
+        if let Some(pos) = bucket.items.iter().position(|existing| {
+            existing.id == notification.id
```
huylexuan44: Not channel-scoped, unlike the message-id branch below - a pending id reused in two channels merges and drops one.

crates/mezon-store/src/inbox.rs
```diff
@@ -305,48 +429,127 @@ impl InboxStore {
         });
     }
 
+    fn merge_server_page(
+        local: Vec<InboxNotification>,
+        fetched: Vec<InboxNotification>,
+    ) -> Vec<InboxNotification> {
+        let fetched_ids: HashSet<String> = fetched.iter().map(|n| n.id.clone()).collect();
+        let fetched_message_ids: HashSet<String> = fetched
```
huylexuan44: No channel key here, so a server item for msg N in channel B drops a pending item for msg N in channel A.

crates/mezon-store/src/inbox.rs
```diff
@@ -267,28 +305,114 @@ impl InboxStore {
...
+            item.topic_id = Some(topic_id.clone());
+            if let Some(preview) = item.message.as_mut() {
+                preview.topic_id = Some(topic_id.clone());
+            }
+        }
     }
 
     fn drop_pending_duplicates(items: &mut Vec<InboxNotification>, incoming: &[InboxNotification]) {
-        let incoming_message_ids: std::collections::HashSet<String> = incoming
+        let incoming_message_ids: HashSet<String> = incoming
```
huylexuan44: Same missing channel scoping as merge_server_page: key incoming_message_ids by (channel_id, message_id).

crates/mezon-store/src/messages.rs
```diff
@@ -1435,8 +1441,48 @@ impl MessagesStore {
...
+            message_id,
+        ) {
+            return;
+        }
+        let pending = PendingLastSeen {
+            clan_id,
+            channel_id: topic_id,
+            message_id,
+            create_time,
+            mode: STREAM_MODE_THREAD,
```
huylexuan44: mode is hard-coded to STREAM_MODE_THREAD; topics are not threads and replies use the parent channel mode. Pass self.mode.

crates/mezon-store/src/messages.rs
```diff
@@ -1435,8 +1441,48 @@ impl MessagesStore {
...
+            badge_count: 0,
```
huylexuan44: badge_count 0 clears nothing, so the topic unread comes back after refetch. Compute a live count like note_viewport_seen.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -2044,16 +2060,74 @@ impl ChannelMessages {
...
+                .timer(Duration::from_millis(1500))
+                .await;
+            let _ = this.update(cx, |this, cx| {
+                this.highlight_id = None;
+                cx.notify();
+            });
+        }));
+    }
+
+    fn is_topic_jump_target(&self, message_id: MessageId, cx: &App) -> bool {
```
huylexuan44, line 2062: Returns false when is_topic_box, but on_topic_store_event only runs on the topic box, so its JumpTo branch is dead code.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -2215,6 +2289,13 @@ impl ChannelMessages {
     }
 
     fn on_topic_store_event(&mut self, event: &MessagesEvent, cx: &mut Context<Self>) {
+        if let MessagesEvent::JumpTo { message_id } = event {
```
huylexuan44: This branch can never fire: is_topic_jump_target bails on is_topic_box. Drop it or invert the guard.

crates/mezon-store/src/messages.rs
```diff
@@ -6886,8 +7164,15 @@ impl MessagesStore {
     }
 
     fn set_channel(&mut self, channel_id: ChannelId, messages: Vec<Message>) {
-        let active = self.active_channel_id;
         let has_more = has_more_from_oldest(&messages);
+        self.replace_channel(channel_id, messages, has_more);
+    }
+
+    fn replace_channel(&mut self, channel_id: ChannelId, messages: Vec<Message>, has_more: bool) {
+        let protect = self
```
huylexuan44: Protecting active_topic_id drops the old guarantee that active_channel_id survives eviction. Protect both.

crates/mezon-store/src/topics.rs
```diff
@@ -647,6 +677,106 @@ impl TopicsStore {
...
+        if let Some(origin_id) = self.known_topic_origin(topic_id) {
+            self.begin_inbox_topic_jump_to_origin(channel_id, topic_id, origin_id, reply_id, cx);
+            return;
+        }
+        let api = self.api.clone();
+        cx.spawn(async move |this, cx| {
+            let detail = match api.get_topic_detail(&topic_id.to_string()).await {
+                Ok(topic) => topic,
+                Err(e) => {
+                    tracing::error!("get_topic_detail failed for inbox jump: {e}");
```
huylexuan44, line 745: Returning here drops the jump entirely - no fallback to the plain message jump, so the click silently does nothing.

crates/mezon-store/src/inbox.rs
```diff
@@ -467,11 +656,27 @@ impl InboxStore {
...
         for category in categories {
-            self.fetch_page(GLOBAL_INBOX_BUCKET_CLAN_ID, category, None, cx);
+            if Self::should_fetch_initial(self.bucket(category)) {
+                self.fetch_page(GLOBAL_INBOX_BUCKET_CLAN_ID, category, None, cx);
+            }
         }
     }
 }
 
+pub(crate) fn skip_inbox_mention_code(code: i32) -> bool {
```
huylexuan44, line 637: Narrower duplicate of is_user_timeline() - misses Ephemeral, AuditLog, CreateThread. Use is_user_timeline here.

crates/mezon-store/src/topic_badges.rs
```diff
@@ -164,18 +196,22 @@ impl TopicBadgeStore {
             Some(&message_id),
             &dedupe_key,
         ) {
-            vec![clan_id]
+            (
+                vec![clan_id],
+                vec![inbox_notification_from_channel_mention(m)],
```
huylexuan44: No skip_inbox_mention_code filter here, so ephemeral and system codes reach the inbox from this path.

crates/mezon-store/src/badge.rs
```diff
@@ -545,33 +558,47 @@ impl BadgeService {
...
-        tracing::debug!(
-            message_id = message_id_raw,
-            content_time,
-            content_len = notif.content.len(),
-            content_first_byte = notif.content.first().copied(),
-            "badge: parsed notification content"
-        );
-        let message_id = MessageId(message_id_raw);
-        if message_id.is_zero() {
+        let Ok(notification) = inbox_notification_from_api(notif.clone()) else {
```
huylexuan44: InboxStore and TopicBadgeStore already ingest this notification; three decodes per mention. Let InboxStore own it.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -102,23 +102,34 @@ impl InboxPopoverPanel {
...
         let _inbox_sub = cx.subscribe(&inbox_store, |this, _, event, cx| {
             let InboxEvent::Updated { .. } = event;
             if this.tab == InboxTab::Topics {
                 return;
             }
-            this.sync_from_store(cx);
+            this.sync_from_store(cx, false);
+            cx.notify();
+        });
+        let _inbox_obs = cx.observe(&inbox_store, |this, _, cx| {
```
huylexuan44: Store emits Updated and notifies together, so subscribe+observe rebuild every row twice per change. Keep one.

crates/mezon-ui/src/chat/inbox/panel.rs
```diff
@@ -188,25 +201,23 @@ impl InboxPopoverPanel {
...
-            self.scroll_list_to_top();
-            return;
-        }
-        if old_count == count {
-            return;
-        }
-        if old_count == 0 {
+        let first_id = self.cached_items.first().map(|row| row.id().to_string());
+        if pin_top || old_count == 0 || self.list_first_id != first_id {
             self.list_state.reset(count);
```
huylexuan44, line 197: reset + scroll_to_top on any first-id change yanks the reader to the top when a mention arrives while scrolled down.

crates/mezon-ui/src/chat/layout.rs
```diff
@@ -469,6 +469,18 @@ impl ChatLayout {
             }
         })
         .detach();
+        cx.observe(&InboxStore::global(cx), |this, _, cx| {
```
huylexuan44: inbox_display_changed only reads ClanList, so these two observers can never see a change. Drop them.

crates/mezon-ui/src/chat/channel_header.rs
```diff
@@ -1018,6 +1018,18 @@ impl ChatHeader {
...
+            let inbox_badge = this
+                .clan_id
+                .as_deref()
+                .is_some_and(|id| clan_has_inbox_badge(id, cx));
+            if this.inbox_badge != inbox_badge {
+                this.inbox_badge = inbox_badge;
+                cx.notify();
+            }
+        });
+        let _inbox_observe =
```
huylexuan44: Unconditional notify on every inbox store change; gate it like _clan_observe above.

crates/mezon-store/src/inbox.rs
```diff
@@ -193,10 +203,13 @@ impl InboxStore {
         if !bucket.server_loaded {
             return true;
         }
+        if bucket.items.is_empty() {
+            return true;
+        }
         if bucket.fetched_at.is_some_and(|t| t.elapsed() < CACHE_TTL) {
```
huylexuan44: Dead branch: this return and the fall-through are both false, so CACHE_TTL no longer gates anything.

crates/mezon-store/src/inbox.rs
```diff
@@ -211,6 +224,24 @@ impl InboxStore {
...
+        bucket.fetched_at = None;
+        bucket.server_loaded = false;
+    }
+
+    pub fn refresh_category(&mut self, category: InboxCategory, cx: &mut Context<Self>) {
+        Self::invalidate(self.bucket_mut(category));
+        self.fetch_page(GLOBAL_INBOX_BUCKET_CLAN_ID, category, None, cx);
+    }
+
+    pub fn schedule_refresh_category(&mut self, category: InboxCategory, cx: &mut Context<Self>) {
```
huylexuan44: Only called from tests; the real site (add_to_inbox) uses refresh_category, which fetch_page drops while loading.

crates/mezon-store/src/inbox.rs
```diff
@@ -267,28 +305,114 @@ impl InboxStore {
...
+        let Some(topic_id) = notification.effective_topic_id() else {
+            return;
+        };
+        self.topic_by_message.insert(
+            topic_remember_key(&notification.channel_id, &message_id),
+            topic_id,
+        );
+    }
+
+    pub fn remembered_topic_id(&self, channel_id: &str, message_id: &str) -> Option<&str> {
```
huylexuan44: remembered_topic_id has no callers.

crates/mezon-store/src/inbox.rs
```diff
@@ -193,10 +203,13 @@ impl InboxStore {
         if !bucket.server_loaded {
             return true;
         }
+        if bucket.items.is_empty() {
```
huylexuan44: Above the TTL check, an empty bucket now refetches on every popover open; CACHE_TTL no longer gates it.

crates/mezon-store/src/messages.rs
```diff
@@ -1535,6 +1580,18 @@ impl MessagesStore {
         }));
     }
 
+    fn set_pending_last_seen(&mut self, pending: PendingLastSeen, cx: &mut Context<Self>) {
```
huylexuan44, line 1600: Flushing on channel change bypasses the debounce every time topic and channel alternate; each alternation writes.

crates/mezon-store/src/messages.rs
```diff
@@ -3287,10 +3444,23 @@ impl MessagesStore {
...
 
+    pub fn pending_topic_jump(&self) -> Option<MessageId> {
+        self.pending_topic_jump
+    }
+
+    pub fn queue_topic_jump(&mut self, message_id: MessageId) {
+        self.pending_topic_jump = Some(message_id);
+    }
+
+    pub fn clear_pending_topic_jump(&mut self) {
```
huylexuan44, line 3518: Only cleared from render via apply_topic_jump; if the panel never renders, the id sticks and blocks later highlights.

crates/mezon-store/src/topics.rs
```diff
@@ -647,6 +677,106 @@ impl TopicsStore {
...
+            return;
+        };
+        if pending.is_expired(now) {
+            self.pending_inbox_jump = None;
+            return;
+        }
+        let Some(active_channel_id) = MessagesStore::global(cx).read(cx).active_channel_id() else {
+            return;
+        };
+        if !pending.matches(*message_id, active_channel_id, now) {
```
huylexuan44: Matching on (origin_id, channel_id) for 20s hijacks an ordinary reply jump to the same message. Add a request token.

crates/mezon-store/src/topics.rs
```diff
@@ -310,6 +334,9 @@ impl TopicsStore {
     fn new(api: Arc<AppApi>, cx: &mut Context<Self>) -> Self {
         let conn_watch = Self::spawn_connection_watch(api.clone(), cx);
         Self::register_realtime(cx);
+        let messages_sub = cx.subscribe(&MessagesStore::global(cx), |this, _, event, cx| {
```
huylexuan44: MessagesStore::global panics if TopicsStore is built first; use try_global to avoid a new init-order dependency.

crates/mezon-store/src/messages.rs
```diff
@@ -3213,30 +3357,43 @@ impl MessagesStore {
...
             ..Default::default()
         };
+        tracing::info!(
+            message_id = message_id.get(),
+            channel_id = channel_id.get(),
+            storage_id = storage_id.get(),
+            topic_id,
+            "add_to_inbox"
+        );
+        let notification = inbox_notification_from_marked_message_local(&marked);
```
huylexuan44: Row now appears before the API confirms and is deleted on failure; also adds a list_notifications round trip per mark.

crates/mezon-client/src/inbox.rs
```diff
@@ -574,14 +589,16 @@ fn parse_notification_content(bytes: &[u8]) -> Option<InboxMessagePreview> {
         }
         return parse_message_preview_json(bytes);
     }
+    if let Ok(fcm) = api::DirectFcmProto::decode(bytes)
```
huylexuan44: Gating on message_id > 0 drops the preview for FCM payloads without one; the old branch was unconditional.

crates/mezon-store/src/messages.rs
```diff
@@ -3368,11 +3538,117 @@ impl MessagesStore {
...
+                        message_id = anchor,
+                        "request_topic_jump: target not in AROUND window"
+                    );
+                    this.clear_pending_topic_jump();
+                    cx.notify();
+                    return;
                 }
+                recompute_message_grouping(&mut window);
+                let has_more = has_more_from_oldest(&window);
+                this.replace_channel(topic_key, window, has_more);
```
huylexuan44: last_message_by_channel is left untouched here, so an AROUND window loaded first makes topic_has_more_bottom false.

crates/mezon-store/src/badge.rs
```diff
@@ -545,33 +558,47 @@ impl BadgeService {
...
-            content_time
-        } else {
-            i64::from(notif.create_time_seconds)
         };
-        let badge_channel = if notif.topic_id != 0 {
-            ChannelId(notif.topic_id)
-        } else {
-            channel_id
+        InboxStore::global(cx).update(cx, |inbox, cx| {
+            inbox.note_mention(notification.clone(), cx);
```
huylexuan44: Full inbox parse plus two Notification clones on the badge path, replacing the old lightweight parse.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -1789,6 +1799,12 @@ impl ChannelMessages {
                 this.mark_scroll_activity(cx);
 
                 if this.is_topic_box {
+                    if at_bottom {
+                        this.sync_topic_seen(cx);
```
huylexuan44, line 1772: Runs on every scroll event; should_write_last_seen returns true on equality, so each one respawns the debounce task.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -1528,6 +1529,22 @@ impl ChannelMessages {
             cx.notify();
         });
 
+        let topic_badge_sub = cx.subscribe(&TopicBadgeStore::global(cx), |this, _, event, cx| {
```
huylexuan44, line 1508: No change gate, unlike topics_event_sub below - every badge update repaints the list. Reuse its fingerprint.

crates/mezon-ui/src/chat/message/channel_messages.rs
```diff
@@ -4564,6 +4687,7 @@ impl ChannelMessages {
         self.sync_topic_header(cx);
         self.maybe_paginate_topic(cx);
         let header_shown = self.header_shown;
+        self.apply_topic_jump(header_shown, cx);
```
huylexuan44: Runs every frame: reads the store, scans topic_row_ids, and calls MessagesStore::update from inside render.

crates/mezon-store/src/inbox.rs
```diff
@@ -305,48 +429,127 @@ impl InboxStore {
...
+                        } else {
+                            Self::drop_pending_duplicates(&mut bucket.items, &items);
+                            let existing_ids: HashSet<String> =
+                                bucket.items.iter().map(|n| n.id.clone()).collect();
+                            bucket.items.extend(
+                                items.into_iter().filter(|n| !existing_ids.contains(&n.id)),
+                            );
+                            Self::sort_items(&mut bucket.items);
+                        }
+                        Self::apply_remembered_topic_ids(&mut bucket.items, &remembered);
```
huylexuan44: Rescans the whole bucket per page; effective_topic_id can hit serde_json on raw_content for each of up to 200 items.

crates/mezon-store/src/inbox.rs
```diff
@@ -305,48 +429,127 @@ impl InboxStore {
...
+            Ok(remembered) => {
+                self.topic_by_message = remembered;
+                self.prune_topic_by_message();
                 self.emit_updated(cx);
                 cx.notify();
             }
         }
     }
 
+    fn prune_topic_by_message(&mut self) {
```
huylexuan44: Rescans every bucket after each page, allocating two Strings per item. Prune only the keys this page touched.

PR #409, opened 2026-08-25
Show full list channels for integration

giachuvan:


crates/mezon-ui/src/clan/settings/channel_webhook_tab.rs
```diff
@@ -53,7 +53,10 @@ impl ChannelWebhookTab {
         settings: Entity<Settings>,
         cx: &mut Context<Self>,
     ) -> Self {
-        channel_list.update(cx, |store, cx| store.load_for_clan(clan_id, cx));
+        channel_list.update(cx, |store, cx| {
+            store.load_for_clan(clan_id, cx);
+            store.ensure_user_channels_loaded(cx);
```
huylexuan44, line 58: This async load re-sorts channel_options later, but selected_channel_index at L76 is a raw index, never remapped.

crates/mezon-store/src/channel.rs
```diff
@@ -3344,6 +3355,59 @@ impl ChannelList {
...
+        out.sort_by(|a, b| {
+            a.name
+                .to_lowercase()
+                .cmp(&b.name.to_lowercase())
+                .then_with(|| a.id.get().cmp(&b.id.get()))
+        });
+        out
+    }
+
+    fn category_name_for_channel(&self, clan_id: ClanId, channel_id: ChannelId) -> String {
```
huylexuan44: O(users x channels) rescan per call, rebuilt on every ChannelList notify. Build a ChannelId->category map once.

crates/mezon-store/src/channel.rs
```diff
@@ -3344,6 +3355,59 @@ impl ChannelList {
         self.cache.get(&clan_id).map_or(&[], Vec::as_slice)
     }
 
+    pub fn webhook_target_channels_for_clan(&self, clan_id: ClanId) -> Vec<WebhookTargetChannel> {
+        let mut seen = HashSet::new();
+        let mut out = Vec::new();
+
+        for channel in self.user_channels() {
```
huylexuan44, line 3363: user_channels() skips only the server archived flag, so locally deleted/archived channels still show. Filter both.

crates/mezon-client/src/transport.rs
```diff
@@ -4397,7 +4398,7 @@ impl MezonTransport {
         let api_name = "ListChannelDescs";
         let body = api::ListChannelDescsRequest {
             clan_id,
-            limit: 500,
+            limit: CHANNEL_DESC_FETCH_LIMIT,
```
huylexuan44, line 4401: Raising this affects every list_channel_descs caller, and >1000 channels still truncate silently. Page or log it.

crates/mezon-store/src/channel.rs
```diff
@@ -3344,6 +3355,59 @@ impl ChannelList {
...
+                }
+                out.push(WebhookTargetChannel {
+                    id: channel.id,
+                    name: channel.name.clone(),
+                    category_name: category.name.clone(),
+                });
+            }
+        }
+
+        out.sort_by(|a, b| {
```
huylexuan44: React sorts selected-first and keeps store order, not alphabetical. Confirm this deviation is intended.

crates/mezon-ui/src/chat/channel_settings/integrations_tab.rs
```diff
@@ -88,7 +88,10 @@ impl IntegrationsTab {
         ClanMembersStore::global(cx).update(cx, |store, cx| {
             store.ensure_loaded(clan_id, cx);
         });
-        ChannelList::global(cx).update(cx, |store, cx| store.load_for_clan(clan_id, cx));
+        ChannelList::global(cx).update(cx, |store, cx| {
+            store.load_for_clan(clan_id, cx);
+            store.ensure_user_channels_loaded(cx);
```
huylexuan44, line 93: This can be the first ListChannelByUserId; merge_user_channels_from_api_descs clears user_channels and resets badges.

crates/mezon-store/src/channel.rs
```diff
@@ -3344,6 +3355,59 @@ impl ChannelList {
...
+            a.name
+                .to_lowercase()
+                .cmp(&b.name.to_lowercase())
+                .then_with(|| a.id.get().cmp(&b.id.get()))
+        });
+        out
+    }
+
+    fn category_name_for_channel(&self, clan_id: ClanId, channel_id: ChannelId) -> String {
+        for category in self.categories_for_clan(clan_id) {
```
huylexuan44, line 3407: Index 0 is always the synthetic favCate named favoriteChannel, so starred channels render FAVORITECHANNEL. Skip it.

crates/mezon-store/src/channel.rs
```diff
@@ -3344,6 +3355,59 @@ impl ChannelList {
...
+            if category
+                .channels
+                .iter()
+                .any(|channel| channel.id == channel_id)
+            {
+                return category.name.clone();
+            }
+        }
+        String::new()
```
huylexuan44: ListChannelByUserId returns no category, so the channels this PR recovers are exactly the ones with a blank badge.

crates/mezon-store/src/channel.rs
```diff
@@ -5442,6 +5506,90 @@ mod tests {
...
+            channel_avatar: String::new(),
+            topic: String::new(),
+            age_restricted: 0,
+            e2ee: 0,
+            app_id: 0,
+        }
+    }
+
+    #[gpui::test]
+    fn webhook_target_channels_excludes_threads_and_voice(cx: &mut gpui::TestAppContext) {
```
huylexuan44, line 5548: apply_clan_structure back-fills user_channels, so loop 1 satisfies both tests; deleting the structure loop stays green.

crates/mezon-ui/src/clan/settings/channel_webhook_tab.rs
huylexuan44: Also no clamp: an out-of-range Some(i) after the list shrinks makes create_webhook a permanent silent no-op.

crates/mezon-client/src/transport.rs
huylexuan44: Worse: api_channel.go:1076 keys the cache on (clan_id, user_id) without limit, so a warm cache still serves 500.

crates/mezon-store/src/channel.rs
huylexuan44: to_lowercase() is code-point order, so Vietnamese names (U+0111, U+01B0) sort after every ASCII name.

giachuvan:
Overview
This PR fixes incomplete webhook channel dropdowns on large clans (~800+ channels/threads) by sourcing picker data from ListChannelByUserId (user_channels) with a structure-cache fallback, while also raising ListChannelDescs limit to 1000 for sidebar cold-path improvement.

Below is our response to each review comment.

1. channel_webhook_tab.rs — stale selected_channel_index after async reload
Comment: Re-sorting channel_options leaves selected_channel_index as a raw index; out-of-range index makes create_webhook a silent no-op.

Status: Fixed

Replaced index-based selection with selected_channel_id: Option<ChannelId>. On every ChannelList notification we rebuild options and call sync_selected_channel():

Keeps selection if the channel still exists
Resets to the first option if the previous selection disappeared
Clears selection when the list is empty
This removes index invalidation entirely.

2. channel.rs — O(users × channels) category lookup per channel
Comment: category_name_for_channel rescans categories on every call; rebuilt on every ChannelList notify.

Status: Fixed

webhook_target_channels_for_clan now builds a single HashMap<ChannelId, String> per call via channel_category_names_for_clan, then does O(1) lookups. No per-channel category scan.

3. channel.rs — user_channels() includes locally archived/deleted channels
Comment: user_channels() only filters server archived flag; locally archived/deleted channels may still appear.

Status: Fixed

webhook_target_channels_for_clan now skips channels where is_locally_archived(id) or is_locally_deleted(id) is true — matching how apply_clan_structure filters the sidebar cache.

Test added: webhook_target_channels_excludes_locally_archived_user_channels — local archive + ListChannelByUserId refetch still excludes the channel.

4. transport.rs — raising limit to 1000; cache/pagination concerns
Comment: Affects all list_channel_descs callers; >1000 still truncates; server cache keyed by (clan_id, user_id) without limit may still serve 500.

Status: Partially addressed — intentional scope

Limit 1000 kept — aligns with server MAX_USER_CHANNEL; improves sidebar/structure cold fetch (previously 500).
Webhook picker completeness does not depend on this — the primary fix uses ListChannelByUserId (no limit, no structure-cache cap).
Warm server cache serving stale 500 — server-side behavior; fixing cache key invalidation or adding pagination is out of scope for this PR. Tracked as a follow-up for sidebar consistency across the app.
No force-refresh on settings open — would add redundant API load without beating user_channels for the picker use case.
5. channel.rs — alphabetical sort vs React; Vietnamese collation
Comment: React keeps store order (selected-first in UI only); to_lowercase() breaks Vietnamese sort order.

Status: Fixed (sort removed) + aligned with React where applicable

Removed alphabetical sort_by(to_lowercase) from webhook_target_channels_for_clan. Order is now: user_channels iteration order, then structure-cache fallback order — closer to React Redux entity order.
integrations_tab (channel settings edit dropdown): already pins selected channel first at render time only (render_channel_picker), matching React WebhookItemChannelDropdown.
channel_webhook_tab (clan settings create filter): no React equivalent exists; we added the same render-time selected-first sort for long lists (native UX, not a React port).
6. integrations_tab.rs — ensure_user_channels_loaded may clear badges on first fetch
Comment: First ListChannelByUserId call clears user_channels via merge_user_channels_from_api_descs.

Status: Accepted — existing pattern, no change

ensure_user_channels_loaded only fetches when !user_channels_loaded (same as forward modal / command palette).
After fetch, sync_user_channels_from_all_loaded_caches re-merges badge/unread fields from loaded structure caches.
This is pre-existing store behavior, not introduced by the webhook fix. No additional fetch from the webhook tab alone.
7. channel.rs — favorCate → label "FAVORITECHANNEL"
Comment: Category index 0 is synthetic favorCate / "favoriteChannel"; starred channels get wrong badge.

Status: Fixed

Structure loop skips category.id == FAVOR_CATE_ID.
Category name map skips FAVOR_CATE_ID and uses or_insert so the real category wins.
Test added: webhook_target_channels_skips_favorite_category_label.

8. channel.rs — blank category badge for ListChannelByUserId-only channels
Comment: Channels recovered only via user_channels have no category from API.

Status: Accepted limitation + UI polish

ListChannelByUserId does not return category metadata — expected.
These channels are still selectable and valid webhook targets.
Rare edge case (channel in user list but absent from truncated structure cache).
UI fix: category badge is hidden when category_name is empty (no misleading blank label).
No extra API call solely for category labels.

9. channel.rs — tests don't isolate structure vs user_channels paths
Comment: apply_clan_structure back-fills user_channels, so both test loops pass; removing structure loop would still pass.

Status: Fixed

webhook_target_channels_merges_user_channels_missing_from_structure now calls merge_user_channels_from_api_descs without apply_clan_structure — validates the user_channels-only path.
Added tests for favorCate label and local-archive filtering (see #3, #7).

PR #410, opened 2026-08-25
Fix drag attachments into topic

giachuvan:


crates/mezon-ui/src/chat/create_topic_panel.rs
```diff
@@ -252,7 +254,40 @@ impl Render for TopicPanel {
             .child(self.input_bar.clone())
             .child(self.typing.clone());
 
+        let drop_title: SharedString =
+            mezon_i18n::t(&locale, "common.dropFilesToUploadToTopic").into();
+        let mention_input = self.mention_input.clone();
+        let drop_overlay = div()
```
huylexuan44, line 260: Overlay and drop handler duplicate area.rs:580-626 and 681-694 - extract a shared helper to keep both in sync.

crates/mezon-ui/src/chat/create_topic_panel.rs
```diff
@@ -252,7 +254,40 @@ impl Render for TopicPanel {
...
+            .absolute()
+            .inset_0()
+            .invisible()
+            .group_drag_over::<ExternalPaths>("topic-drop-zone", |style| style.visible())
+            .flex()
+            .items_center()
+            .justify_center()
+            .bg(rgba(0x5865f233))
```
huylexuan44: React uses blue-500 (#3b82f6) for tint and border here, not blurple #5865f2 - see TopicDiscussionBox.tsx:317. Alpha 0x33 is correct.

crates/mezon-ui/src/chat/create_topic_panel.rs
```diff
@@ -252,7 +254,40 @@ impl Render for TopicPanel {
...
+            .border_2()
+            .border_dashed()
+            .border_color(rgb(0x5865f2))
+            .rounded(px(8.))
+            .child(
+                div()
+                    .px(px(24.))
+                    .py(px(12.))
+                    .rounded(px(8.))
+                    .bg(rgb(0x5865f2))
```
huylexuan44: Inner badge is missing shadow-lg from the React reference.

crates/mezon-ui/src/chat/create_topic_panel.rs
```diff
@@ -252,7 +254,40 @@ impl Render for TopicPanel {
...
+                            .text_lg()
+                            .font_weight(FontWeight::SEMIBOLD)
+                            .text_color(rgb(0xffffff))
+                            .child(drop_title),
+                    ),
+            );
+
         v_flex()
+            .relative()
+            .group("topic-drop-zone")
```
huylexuan44, line 291: Group on the panel root makes the overlay cover the header; React scopes it below TopicHeader (topicDiscussion/index.tsx:26).

giachuvan:
Comment 1 — Extract shared helper with area.rs (overlay + drop handler)
Not fixed — kept as-is (option A).

The reviewer said this duplicates area.rs:580-626 and 681-694, but the two overlays are actually different per the React reference:

Channel (area.rs): dark dimmed background, large blurple #5865f2 box, with icon + channel title + upload instructions.
Topic (TopicDiscussionBox.tsx): light blue-500 overlay, dashed border, single-line badge dropFilesToUploadToTopic.
Only the on_drop → add_dropped_paths path (~5 lines) is shared. Extracting a common overlay would force two different UIs into one helper, which doesn’t match the 1:1 port goal. We’re not porting the channel pattern to topic just because the reviewer suggested DRY.

Comment 2 — Color #5865f2 instead of #3b82f6
Fixed.

React TopicDiscussionBox.tsx:317-319 uses blue-500 (#3b82f6) for tint, border, and inner badge. #5865f2 is the channel drop zone blurple (area.rs) and was mistakenly reused for topic. Updated to #3b82f6; alpha 0x33 (~20%) kept as-is to match bg-opacity-20.

Comment 3 — Inner badge missing shadow-lg
Fixed.

React has shadow-lg on the inner badge (TopicDiscussionBox.tsx:319). Added .shadow_lg() in GPUI, consistent with other components in the codebase.

Comment 4 — Group on panel root causes overlay to cover the header
Fixed.

React clearly separates drop zone scope:

TopicHeader is outside the drop zone (topicDiscussion/index.tsx:26-30)
Only TopicDiscussionBox has onDragEnter / onDrop and the absolute inset-0 overlay
Desktop TopicPanel combines header + content in one component. Previously, group("topic-drop-zone") and the overlay were on the panel root, so they covered the header too. Moved group, on_drop, and the overlay to a child container below the header (drop_body), covering only timeline + composer — matching React.

PR #411, opened 2026-08-25
Fix/edit event

duocphungvanNCC:
> `Description:`

**env development:** window
- edit event
- confirm and cancel event
- share link event
- invite people into event
- fix the title bar in clan guide, member & channel list page is misaligned.
- adjust icon clan guide.


crates/mezon-store/src/events.rs
```diff
@@ -237,7 +258,10 @@ impl EventsStore {
                 } else {
                     ClanId(event.clan_id)
                 };
-                let item = ClanEventItem::from_api(event);
+                let mut item = ClanEventItem::from_api(event);
```
huylexuan44: **[P1] Preserve the requested repeat type here.** The current backend CreateEvent response does not populate RepeatType, while the realtime event does. Since realtime is emitted before this response returns, this assignment can overwrite a correctly hydrated recurring event with repeat_type = 0. The edit form then maps 0 to “does not repeat” and can silently persist that change. Capture draft.repeat_type before moving the request and use it when constructing this item, or refetch before caching; please add a recurring-event regression test.

crates/mezon-ui/src/chat/media_channel/media_image.rs
```diff
@@ -21,6 +21,24 @@ pub fn render_media_image(
...
 
+pub fn render_media_image_rounded_top(
+    bg: Rgba,
+    muted: Rgba,
+    url: SharedString,
+    width: Pixels,
+    height: Pixels,
+    radius: Pixels,
+) -> impl IntoElement {
+    img(url)
```
huylexuan44, line 32: **[P2] Attach a bounded image cache inside the modal.** Shell renders modals through a deferred boundary, which does not restore the ancestor image_cache stack. This bare dynamic img therefore falls through to the non-evicting GPUI global asset cache, retaining unique event-cover renditions for the app lifetime. Pass an LruImageCache into this helper or attach one at the modal root.

crates/mezon-ui/src/chat/clan_events_page.rs
```diff
@@ -434,6 +700,402 @@ impl ClanEventsModal {
...
+                        .ok();
+                },
+            );
+        }
+
+        let weak_copy = weak.clone();
+        menu = menu.item(
+            mezon_i18n::t(&locale, "eventCreator.actions.copyEventLink"),
+            move |_window, cx| {
+                cx.write_to_clipboard(ClipboardItem::new_string(event_link.clone()));
```
huylexuan44: **[P3] Avoid copying an empty link.** Address-only events may legitimately have neither channel_voice_id nor channel_id, so event_link() returns an empty String, but this action still writes it to the clipboard and reports success. Hide or disable Copy Event Link when event_link is empty, or define a real event URL for this case.

crates/mezon-ui/src/chat/clan_events_page.rs
```diff
@@ -324,32 +409,132 @@ impl ClanEventsModal {
...
+                                                        store.set_user_interested(
+                                                            this.clan_id,
+                                                            event_id,
+                                                            user_id,
+                                                            !user_is_interested,
+                                                            cx,
+                                                        )
+                                                    },
+                                                );
+                                                this.event_action_task =
```
huylexuan44: **[P2] Do not replace one global task for every interest action.** Assigning this field drops and immediately cancels the previous GPUI Task. Clicking another event can therefore cancel the prior RPC and its rollback after the optimistic user_ids change has already been applied. Track pending tasks per event, detach with explicit lifecycle handling, or disable only the affected event until completion.

crates/mezon-ui/src/chat/create_event_modal.rs
```diff
@@ -965,6 +1074,17 @@ impl CreateEventModal {
...
         let theme = cx.theme();
+        let editing_private = self
+            .original_event
+            .as_ref()
+            .is_some_and(|event| event.is_private);
+        let editing_non_private = self
+            .original_event
+            .as_ref()
+            .is_some_and(|event| !event.is_private);
+        let can_select_audience = self.original_event.as_ref().map_or(true, |event| {
```
huylexuan44: **[P1] This is the second required-lint failure.** Clippy reports unnecessary-map-or here under -D warnings. This can be expressed with self.original_event.as_ref().is_none_or(...).

crates/mezon-ui/src/chat/create_event_modal.rs
```diff
@@ -603,14 +684,23 @@ impl CreateEventModal {
...
+                        .is_some_and(|event| start == event.start_time_seconds))
                 && end > start
         })
     }
 
     fn time_error(&self, cx: &App) -> Option<String> {
         let (start, end) = self.timestamps(cx)?;
-        if start <= chrono::Utc::now().timestamp().max(0) as u32 {
+        if start <= chrono::Utc::now().timestamp().max(0) as u32
+            && !self
```
huylexuan44: **[P1] This fails the required Clippy gate.** With -D warnings, Clippy reports nonminimal-bool here. Rewrite the Option check using is_none_or so the PR can pass lint.

PR #424, opened 2026-08-27
Fix render quote content in list threads

giachuvan:


crates/mezon-ui/src/chat/threads_popover.rs
```diff
@@ -890,7 +895,7 @@ fn resolve_thread_preview(thread: &ThreadSummary, cx: &App) -> ThreadCardPreview
     };
 
     ThreadCardPreview {
-        content: thread.last_message_content.clone(),
+        content: display_text_from_message_content(&thread.last_message_content),
```
huylexuan44: [P2] Both values reaching this view are already decoded display text: thread summaries use parse_message_text, and cached Message.content comes from content_tokens.t. Parsing them again as wire JSON makes legitimate messages such as {"foo":"bar"} render blank, or changes {"t":"other"} to other. Please collapse whitespace directly and test with decoded text rather than a raw wire payload.

crates/mezon-ui/src/chat/threads_popover.rs
```diff
@@ -899,6 +904,32 @@ fn resolve_thread_preview(thread: &ThreadSummary, cx: &App) -> ThreadCardPreview
     }
 }
 
+fn thread_list_preview_text(content: &str) -> String {
+    let text = display_text_from_message_content(content);
+    let normalized = text
+        .replace("\\n", " ")
+        .replace("\\r", " ")
+        .replace("\\t", " ");
```
huylexuan44: [P2] These replacements corrupt valid user text containing literal \\n, \\r, or \\t sequences, such as code snippets and Windows paths like C:\\temp. Actual newline, carriage-return, and tab characters are already handled by char::is_whitespace() below, so the literal replacements should be removed.

crates/mezon-ui/src/chat/threads_popover.rs
```diff
@@ -688,7 +689,7 @@ fn thread_card(
             mezon_i18n::t(locale, "message.attachments.attachment")
         )
     } else {
-        preview.content.clone()
+        thread_list_preview_text(&preview.content)
```
huylexuan44: [P2] This normalization now runs inside thread_card for every visible row on each re-render. The helper performs JSON parsing, several chained replace allocations, and another output allocation; the ThreadSummary path also parses twice. Since transport already decodes the content, please precompute or cache the normalized preview when store data changes and keep the row render path allocation-light.

PR #425, opened 2026-08-27
Feat: action channel list page admin

duocphungvanNCC:
> `Description:`

**env development:** window
- right click open option admin manage
- alow archive channel
- navigate into edit channel
- need confirm delete channel

image

### DIFF
--- a/crates/mezon-store/src/channel_settings.rs
+++ b/crates/mezon-store/src/channel_settings.rs
@@ -116,11 +116,7 @@ impl ChannelSettingsStore {
         channel_id: ChannelId,
         cx: &mut Context<Self>,
     ) {
-        for pending in self.pending_restored_rows.values_mut() {
-            pending.remove(&channel_id);
-        }
-        self.pending_restored_rows
-            .retain(|_, pending| !pending.is_empty());
+        clear_pending_restore(&mut self.pending_restored_rows, clan_id, channel_id);
         self.rows.remove(&(clan_id, channel_id));
         let changed = remove_matching_rows(&mut self.rows, clan_id, channel_id);
         self.notify_changed(clan_id, changed, cx);
@@ -218,7 +214,7 @@ impl ChannelSettingsStore {
                     Ok(mut rows) => {
                         merge_pending_restores(&mut rows, &mut this.pending_restored_rows, key);
                         if let Some(channel_list) = crate::channel::ChannelList::try_global(cx) {
-                            let active_ids = rows.iter().map(|row| row.id).collect::<Vec<_>>();
+                            let active_ids = active_row_ids(&rows);
                             channel_list.update(cx, |channel_list, cx| {
                                 channel_list.reconcile_active_channels(clan_id, &active_ids, cx)
                             });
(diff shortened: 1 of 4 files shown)


crates/mezon-store/src/channel_settings.rs
```diff
@@ -332,16 +394,28 @@ impl ChannelSettingsStore {
...
-                );
+                let parent_id = ChannelId(event.parent_id);
+                let channel_id = ChannelId(event.channel_id);
+                let changed = patch_matching_rows(&mut self.rows, clan_id, channel_id, |row| {
+                    row.label = event.channel_label.clone();
+                    row.private = event.channel_private;
+                    row.channel_type = event.channel_type;
+                    row.active = event.active;
+                });
+                if missing_active_update_is_restore(event.active, &changed) {
```
huylexuan44: [P1] Scope this restore path to loaded settings keys. ChannelSettingsStore is global, so changed.is_empty() is true for ordinary ChannelUpdated events whenever this admin page or parent was never loaded. Active channel edits carry active = 1, so this path synthesizes a restore and upsert_pending_restore unconditionally calls the admin-only ListChannelSetting RPC. Normal users then generate permission errors and retain pending rows. Prefer the explicit ChannelArchive(active = 1) restore signal, and only patch or reload keys already loaded or loading.

crates/mezon-store/src/channel_settings.rs
```diff
@@ -165,7 +215,14 @@ impl ChannelSettingsStore {
                     return;
                 }
                 match result {
-                    Ok(rows) => {
+                    Ok(mut rows) => {
+                        merge_pending_restores(&mut rows, &mut this.pending_restored_rows, key);
+                        if let Some(channel_list) = crate::channel::ChannelList::try_global(cx) {
+                            let active_ids = rows.iter().map(|row| row.id).collect::<Vec<_>>();
```
huylexuan44: [P1] Filter this list by row.active != 0 before calling reconcile_active_channels. setting_item_visible intentionally retains inactive thread rows, so the current code passes archived thread IDs as active and removes their ChannelList archived tombstones. Expanding or reloading a parent can therefore make an active = 0 server thread look locally restored and trigger refresh or detail work for it.

crates/mezon-store/src/channel_settings.rs
```diff
@@ -86,13 +90,42 @@ impl ChannelSettingsStore {
...
+            .find(|row| row.id == channel_id)
+    }
+
+    pub fn remove_channel_locally(
+        &mut self,
+        clan_id: ClanId,
+        channel_id: ChannelId,
+        cx: &mut Context<Self>,
+    ) {
+        for pending in self.pending_restored_rows.values_mut() {
```
huylexuan44: [P2] Reuse this pending-restore cleanup in the RealtimeEvent::ChannelDeleted branch too. If a restore event is pending and the channel is deleted before the confirming reload completes, that branch removes only visible rows; merge_pending_restores can then inject the deleted synthetic row again on later fetches.

crates/mezon-ui/src/chat/clan_channels_page.rs
```diff
@@ -862,12 +894,145 @@ impl Render for ClanChannelsPage {
...
+    let has_permission = |permission| {
+        permissions
+            .as_ref()
+            .is_some_and(|permissions| permissions.read(cx).check(clan_id, None, permission, cx))
+    };
+    let has_owner = has_permission(PERMISSION_CLAN_OWNER);
+    let has_admin = has_permission(PERMISSION_ADMINISTRATOR);
+    let has_manage_clan = has_permission(PERMISSION_MANAGE_CLAN);
+    let has_manage_channel = has_permission(PERMISSION_MANAGE_CHANNEL);
+    let can_archive = !archive_menu_hidden(channel_type, is_welcome)
```
huylexuan44: [P2] Include state.row.active != 0 in can_archive (or hide inactive threads). The settings fetch deliberately keeps active = 0 threads, so they currently still get an Archive Thread action; archiving them again is a no-op and the next settings refresh brings the row back.

huylexuan44 (commented):
Found four correctness issues in the realtime/settings reconciliation.

PR #435, opened 2026-08-28
Fix join stream channel refresh expired jwt

giachuvan:


crates/mezon-store/src/stream.rs
```diff
@@ -23,6 +23,18 @@ const STREAM_FETCH_LIMIT: i32 = 100;
...
+fn stream_uses_jwt(session: &mezon_client::Session) -> bool {
+    session.session_id.is_empty()
+}
+
+fn stream_jwt_expired(session: &mezon_client::Session) -> bool {
+    if !stream_uses_jwt(session) {
+        return false;
+    }
+    let now = mezon_client::server_now_secs();
+    session.token.is_empty() || session.expires_at == 0 || now >= session.expires_at
```
huylexuan44: expires_at == 0 is "unknown" in jwt_is_fresh and token_expired. Here it means expired, refresh no-ops, and join fails forever.

crates/mezon-store/src/stream.rs
```diff
@@ -23,6 +23,18 @@ const STREAM_FETCH_LIMIT: i32 = 100;
...
 
+fn stream_uses_jwt(session: &mezon_client::Session) -> bool {
+    session.session_id.is_empty()
+}
+
+fn stream_jwt_expired(session: &mezon_client::Session) -> bool {
+    if !stream_uses_jwt(session) {
+        return false;
+    }
+    let now = mezon_client::server_now_secs();
```
huylexuan44, line 41: No clock-skew margin. connection.rs uses JWT_SKEW = 60s; a token with 2s left still dies during the WS connect.

crates/mezon-store/src/stream.rs
```diff
@@ -754,6 +858,48 @@ impl StreamStore {
...
+        let auth_state = login.read(cx).auth_state();
+        auth_state.update(cx, |state, cx| {
+            let session = match state {
+                AuthState::Authenticated(session) | AuthState::Connecting(session) => session,
+                _ => return None,
+            };
+            if session.user_id != expected_user_id {
+                return None;
+            }
+            session.apply_refresh(
```
huylexuan44: Applies blindly. spawn_token_watch (connection.rs:537) compares first; a token that landed during the await gets clobbered.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
+                    "stream join refreshing expired jwt credential"
+                );
+                match api.renew_fallback_token().await {
+                    Ok(renewed) => {
+                        if let Some(applied) =
+                            apply_stream_session_refresh(&renewed, &session.user_id, cx)
+                        {
+                            session = applied;
+                        } else {
+                            session.apply_refresh(
```
huylexuan44: This branch spends the refresh token then writes the replacement only to the task-local Session, so it is lost app-wide.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
         self.error_message = None;
```
huylexuan44, line 550: Both early returns above (512-523) skip disconnect_session(), leaving the previous session and its audio running.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
+                        .expires_at
+                        .saturating_sub(mezon_client::server_now_secs()),
+                    "stream join refreshing expired jwt credential"
+                );
+                match api.renew_fallback_token().await {
+                    Ok(renewed) => {
+                        if let Some(applied) =
+                            apply_stream_session_refresh(&renewed, &session.user_id, cx)
+                        {
+                            session = applied;
```
huylexuan44: applied may carry a session_id, so the gate below passes without revalidating the JWT and ws_credential() sends the sid.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
+                clan_id = clan_id.get(),
+                cred_kind,
+                refreshed,
+                jwt_valid_for = session
+                    .expires_at
+                    .saturating_sub(mezon_client::server_now_secs()),
+                "stream store join requested"
+            );
+
+            let started = this.update(cx, |this, cx| {
```
huylexuan44, line 658: JOIN_TIMEOUT is armed before the refresh, so the refresh spends the connect budget. Arm the timer here instead.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
-            output_device_id,
-            self.volume,
-            self.muted,
-        );
-        let events = stream_session.events().clone();
-        self.session = Some(stream_session);
+        let api = self.api.clone();
+        let ws_base_url = config.stream_ws_url.clone();
+        let mut session = session.clone();
+        let volume = self.volume;
```
huylexuan44: volume/muted snapshotted pre-await and sync_audio_playback() no-ops while session is None. Read them inside the update at 618.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
+            let mut refreshed = false;
+            if stream_jwt_expired(&session) {
+                tracing::info!(
+                    channel_id = channel_id.get(),
+                    jwt_valid_for = session
+                        .expires_at
+                        .saturating_sub(mezon_client::server_now_secs()),
+                    "stream join refreshing expired jwt credential"
+                );
+                match api.renew_fallback_token().await {
```
huylexuan44: Non-idempotent RPC inside the cancel-on-drop _session_task; disconnect_session() during this await strands a spent refresh token.

crates/mezon-store/src/stream.rs
```diff
@@ -754,6 +858,48 @@ impl StreamStore {
...
+        return None;
+    };
+    let persisted = session.clone();
+    cx.background_executor()
+        .spawn(async move {
+            if let Err(e) = crate::login::LoginStore::persist_session(&persisted) {
+                tracing::warn!("Failed to persist stream-refreshed session: {e}");
+            }
+        })
+        .detach();
```
huylexuan44: connection.rs:563 and :738 await this write on purpose; detaching leaves a rotated-away refresh token if the app quits here.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
+                        this.phase,
+                        StreamPhase::Joining {
+                            channel_id: joining_channel,
+                            ..
+                        } if joining_channel == channel_id
+                    ) {
+                        return;
+                    }
+                    this.phase = StreamPhase::Error("Stream connection failed".into());
+                    this.error_message = Some("Stream connection failed".into());
```
huylexuan44: error_message is store-global and rendered without a channel check, so this banner shows on whatever stream channel is open.

crates/mezon-store/src/stream.rs
```diff
@@ -754,6 +858,48 @@ impl StreamStore {
     }
 }
 
+fn apply_stream_session_refresh(
```
huylexuan44: spawn_token_watch (connection.rs:520) already adopts and persists these RenewedTokens. Drop this helper and use renewed.token.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
-        let events = stream_session.events().clone();
-        self.session = Some(stream_session);
+        let api = self.api.clone();
+        let ws_base_url = config.stream_ws_url.clone();
+        let mut session = session.clone();
+        let volume = self.volume;
+        let muted = self.muted;
+        let frame_store = self.frame_store.clone();
 
         self._session_task = Some(cx.spawn(async move |this, cx| {
```
huylexuan44: StreamSession::start never blocks; spawning unconditionally costs a main-loop turn on every join and forces both phase guards.

crates/mezon-store/src/stream.rs
```diff
@@ -520,32 +532,124 @@ impl StreamStore {
...
+                    if !matches!(
+                        this.phase,
+                        StreamPhase::Joining {
+                            channel_id: joining_channel,
+                            ..
+                        } if joining_channel == channel_id
+                    ) {
+                        return;
+                    }
+                    this.phase = StreamPhase::Error("Stream connection failed".into());
```
huylexuan44: This path logs nothing and reuses the string mezon-stream emits for a real WS failure. Log the cause and use a distinct message.

crates/mezon-store/src/stream.rs
```diff
@@ -838,4 +984,37 @@ mod tests {
         assert_eq!(filtered.len(), 1);
         assert_eq!(filtered[0].user_id, UserId(10));
     }
+
+    #[test]
```
huylexuan44, line 1058: init_store never calls LoginStore::init, so apply_stream_session_refresh is unreachable in tests. Pass Entity<AuthState> instead.

PR #440, opened 2026-08-30
Fix(gpui): sync IME surrounding in-frame to avoid reentrant App borro…

giachuvan:
…w on focus

### DIFF
--- a/crates/mezon-ui/src/chat/mention_input/mod.rs
+++ b/crates/mezon-ui/src/chat/mention_input/mod.rs
@@ -2055,8 +2055,15 @@ impl MentionInput {
     }
 
     fn open_popup(&mut self, tab: SubPanel, window: &mut Window, cx: &mut Context<Self>) {
+        self.input.update(cx, |input, cx| {
+            input.drop_uncommitted_preedit(cx);
+        });
+        window.reset_ime();
         let locale = self.locale(cx);
         let popup = cx.new(|cx| GifStickerEmojiPopup::new(tab, locale, window, cx));
+        popup.update(cx, |popup, cx| {
+            popup.focus_search(window, cx);
+        });
         let pick_sub = cx.subscribe_in(
             &popup,
             window,
@@ -2071,6 +2078,7 @@ impl MentionInput {
                     cx,
                 ),
                 GifStickerEmojiEvent::Sticker { url, filename } => {
+                    this.abandon_search_ime(window, cx);
                     this.close_popup();
                     cx.emit(MentionInputEvent::SendSticker {
                         url: url.clone(),
@@ -2079,6 +2087,7 @@ impl MentionInput {
                     cx.notify();
                 }
                 GifStickerEmojiEvent::Gif { url, width, height } => {
+                    this.abandon_search_ime(window, cx);
                     this.close_popup();
                     cx.emit(MentionInputEvent::SendGif {
                         url: url.clone(),
@@ -2088,6 +2097,7 @@ impl MentionInput {
                     cx.notify();
                 }
                 GifStickerEmojiEvent::Sound { url, filename } => {
+                    this.abandon_search_ime(window, cx);
                     this.close_popup();
                     cx.emit(MentionInputEvent::SendSound {
                         url: url.clone(),
@@ -2214,12 +2224,21 @@ impl MentionInput {
         Ok(())
     }
 
+    fn abandon_search_ime(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        if let Some(popup) = &self.popup {
+            popup.update(cx, |popup, cx| popup.drop_search_preedit(window, cx));
+        } else {
+            window.reset_ime();
+        }
+    }
+
     fn insert_emoji(
         &mut self,
         emoji: EmojiSuggestRaw,
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
+        self.abandon_search_ime(window, cx);
         let content_len = self.input.read(cx).value().len();
         let at = self.input.read(cx).cursor().min(content_len);
         let display = emoji.shortname;
--- a/crates/mezon-ui/src/chat/mention_input/mod.rs
+++ b/crates/mezon-ui/src/chat/mention_input/mod.rs
@@ -774,7 +774,7 @@ impl MentionInput {
         self.committed = committed_from_compose_tokens(&text, tokens);
         self.pending_attachments = attachments;
         self.reset_popup();
-        self.close_popup();
+        self.close_popup(window, cx);
         self.clear_suggestions(cx);
         self.clear_ephemeral(cx);
         self.clear_ogp_preview(cx);
@@ -832,7 +832,7 @@ impl MentionInput {
             self.convert_text_to_file(text, window, cx);
             self.committed.clear();
             self.reset_popup();
-            self.close_popup();
+            self.close_popup(window, cx);
             self.clear_ogp_preview(cx);
             self.input.update(cx, |input, cx| {
                 input.set_mention_spans(Vec::new(), cx);
@@ -845,7 +845,7 @@ impl MentionInput {
         let ogp = self.take_outgoing_ogp();
         self.committed.clear();
         self.reset_popup();
-        self.close_popup();
+        self.close_popup(window, cx);
         self.input.update(cx, |input, cx| {
             input.set_mention_spans(Vec::new(), cx);
             input.clear_after_send(swallow, window, cx);
@@ -2042,7 +2042,7 @@ impl MentionInput {
     fn toggle_tab(&mut self, tab: SubPanel, window: &mut Window, cx: &mut Context<Self>) {
         if let Some(popup) = &self.popup {
             if popup.read(cx).active_tab() == tab {
-                self.close_popup();
+                self.close_popup(window, cx);
             } else {
                 popup.update(cx, |popup, cx| popup.set_tab(tab, window, cx));
             }
@@ -2079,7 +2079,7 @@ impl MentionInput {
                 ),
                 GifStickerEmojiEvent::Sticker { url, filename } => {
                     this.abandon_search_ime(window, cx);
-                    this.close_popup();
+                    this.close_popup(window, cx);
                     cx.emit(MentionInputEvent::SendSticker {
                         url: url.clone(),
                         filename: filename.clone(),
@@ -2088,7 +2088,7 @@ impl MentionInput {
                 }
                 GifStickerEmojiEvent::Gif { url, width, height } => {
                     this.abandon_search_ime(window, cx);
-                    this.close_popup();
+                    this.close_popup(window, cx);
                     cx.emit(MentionInputEvent::SendGif {
                         url: url.clone(),
                         width: *width,
@@ -2098,7 +2098,7 @@ impl MentionInput {
                 }
                 GifStickerEmojiEvent::Sound { url, filename } => {
                     this.abandon_search_ime(window, cx);
-                    this.close_popup();
+                    this.close_popup(window, cx);
                     cx.emit(MentionInputEvent::SendSound {
                         url: url.clone(),
                         filename: filename.clone(),
@@ -2107,17 +2107,26 @@ impl MentionInput {
                 }
             },
         );
-        let dismiss_sub = cx.subscribe(&popup, |this, _popup, _: &DismissEvent, cx| {
-            this.close_popup();
-            cx.notify();
-        });
+        let dismiss_sub = cx.subscribe_in(
+            &popup,
+            window,
+            |this, _popup, _: &DismissEvent, window, cx| {
+                this.close_popup(window, cx);
+                cx.notify();
+            },
+        );
         self._popup_subs = vec![pick_sub, dismiss_sub];
         self.popup = Some(popup);
     }
 
-    fn close_popup(&mut self) {
+    fn close_popup(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        let was_open = self.popup.is_some();
         self.popup = None;
         self._popup_subs.clear();
+        if was_open {
+            let handle = self.input.read(cx).focus_handle(cx);
+            window.focus(&handle, cx);
+        }
     }
 
     pub fn show_panel(&mut self, tab: SubPanel, window: &mut Window, cx: &mut Context<Self>) {
@@ -2129,8 +2138,8 @@ impl MentionInput {
         cx.notify();
     }
 
-    pub fn hide_panel(&mut self, cx: &mut Context<Self>) {
-        self.close_popup();
+    pub fn hide_panel(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        self.close_popup(window, cx);
         cx.notify();
     }
 
@@ -2207,9 +2216,10 @@ impl MentionInput {
         filename: String,
         width: i32,
         height: i32,
+        window: &mut Window,
         cx: &mut Context<Self>,
     ) -> anyhow::Result<()> {
-        self.close_popup();
+        self.close_popup(window, cx);
         match kind {
             "sticker" => cx.emit(MentionInputEvent::SendSticker { url, filename }),
             "gif" => cx.emit(MentionInputEvent::SendGif {
@@ -2309,7 +2319,7 @@ impl MentionInput {
                 input.set_value("", window, cx);
             });
         } else if self.popup.is_some() {
-            self.close_popup();
+            self.close_popup(window, cx);
             cx.notify();
         } else {
             cx.emit(MentionInputEvent::Cancel);
@@ -2512,11 +2522,11 @@ impl MentionInput {
                     .mb(px(10.))
                     .occlude()
                     .on_mouse_down_out(cx.listener(
-                        move |this, event: &gpui::MouseDownEvent, _, cx| {
+                        move |this, event: &gpui::MouseDownEvent, window, cx| {
                             if toggle_bounds.get().contains(&event.position) {
                                 return;
                             }
-                            this.close_popup();
+                            this.close_popup(window, cx);
                             cx.notify();
                         },
                     ))
--- a/crates/mezon-ui/src/chat/mention_input/text_field.rs
+++ b/crates/mezon-ui/src/chat/mention_input/text_field.rs
@@ -826,7 +826,7 @@ impl MentionInputState {
     }
 
     fn on_key_down(&mut self, event: &KeyDownEvent, _: &mut Window, _: &mut Context<Self>) {
-        if event.keystroke.key == "enter" {
+        if event.keystroke.key == "enter" && !event.keystroke.modifiers.modified() {
             return;
         }
         self.discard_ime_commit = None;
--- a/crates/mezon-ui/src/chat/mention_input/text_field.rs
+++ b/crates/mezon-ui/src/chat/mention_input/text_field.rs
@@ -32,7 +32,8 @@ use crate::util::text_edit::{
     EditKind, HistoryEntry, MAX_UNDO_HISTORY, SelectGranularity, extend_range_for_granularity,
     granularity_for_click, home_target, ime_replace_range, line_end, line_start,
     marked_caret_range, marked_range_after_delete, next_word_boundary, previous_word_boundary,
-    range_for_granularity, should_coalesce, surrounding_delete_range, swallow_discarded_ime_commit,
+    range_for_granularity, should_coalesce, splice_out_byte_range, surrounding_delete_range,
+    swallow_discarded_ime_commit,
 };
 
 const MASK: char = '\u{2022}';
@@ -338,16 +339,12 @@ impl MentionInputState {
         let Some(marked) = self.marked_range.take() else {
             return;
         };
-        let marked = self.clamp_range(marked);
-        if marked.start >= marked.end {
+        let Some((next, discarded, caret)) = splice_out_byte_range(&self.content, marked) else {
             return;
-        }
-        self.discard_ime_commit = self.content.get(marked.clone()).map(str::to_string);
-        let mut next = String::with_capacity(self.content.len() - (marked.end - marked.start));
-        next.push_str(&self.content[..marked.start]);
-        next.push_str(&self.content[marked.end..]);
+        };
+        self.discard_ime_commit = Some(discarded);
         self.set_content(next);
-        self.selected_range = marked.start..marked.start;
+        self.selected_range = caret..caret;
         self.selection_reversed = false;
         cx.notify();
         cx.emit(MentionFieldEvent::Change);
--- a/crates/mezon-ui/src/chat/mention_input/text_field.rs
+++ b/crates/mezon-ui/src/chat/mention_input/text_field.rs
@@ -334,6 +334,25 @@ impl MentionInputState {
         }
     }
 
+    pub(crate) fn drop_uncommitted_preedit(&mut self, cx: &mut Context<Self>) {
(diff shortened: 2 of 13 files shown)


crates/vendor/gpui/src/window.rs
```diff
@@ -5419,8 +5419,10 @@ impl Window {
         self.on_next_frame(|window, cx| {
             if let Some(mut input_handler) = window.platform_window.take_input_handler() {
                 let bounds = input_handler.selected_bounds(window, cx);
+                let surrounding = input_handler.surrounding_text_in(window, cx);
                 window.platform_window.set_input_handler(input_handler);
                 if let Some(bounds) = bounds {
+                    window.platform_window.set_ime_surrounding_hint(surrounding);
```
huylexuan44, line 5437: Hint is never retracted. Clear it right after update_ime_position, or a focus change mid-composition leaks a stale snapshot to the post-commit sync at x11/client.rs:1530.

crates/vendor/gpui/src/window.rs
```diff
@@ -5419,8 +5419,10 @@ impl Window {
         self.on_next_frame(|window, cx| {
             if let Some(mut input_handler) = window.platform_window.take_input_handler() {
                 let bounds = input_handler.selected_bounds(window, cx);
+                let surrounding = input_handler.surrounding_text_in(window, cx);
```
huylexuan44, line 5434: Result is dropped when bounds is None, leaving the previous hint live. draw() clears the input handler on blur, so a hint can outlive the field it came from.

crates/vendor/gpui/src/window.rs
```diff
@@ -5419,8 +5419,10 @@ impl Window {
...
                 window.platform_window.set_input_handler(input_handler);
                 if let Some(bounds) = bounds {
+                    window.platform_window.set_ime_surrounding_hint(surrounding);
                     window.platform_window.update_ime_position(bounds);
```
huylexuan44, line 5438: update_ime_position also calls drain_dbus_im (x11/client.rs:447) inside this App borrow, so handle_ime_commit still re-enters and logs. Verify the noise is actually gone.

crates/vendor/gpui_linux/src/linux/x11/window.rs
```diff
@@ -1275,6 +1277,9 @@ impl X11WindowStatePtr {
 
     pub fn get_ime_surrounding(&self) -> Option<gpui::ImeSurroundingText> {
         let mut state = self.state.borrow_mut();
+        if let Some(hint) = state.ime_surrounding_hint.take() {
```
huylexuan44, line 1285: A stale hint here corrupts ImEvent::DeleteSurrounding (x11/client.rs:1493) — its byte offsets are computed against this exact string.

crates/vendor/gpui_linux/src/linux/wayland/window.rs
```diff
@@ -1142,6 +1144,9 @@ impl WaylandWindowStatePtr {
 
     pub fn get_ime_surrounding(&self) -> Option<ImeSurroundingText> {
         let mut state = self.state.borrow_mut();
+        if let Some(hint) = state.ime_surrounding_hint.take() {
```
huylexuan44, line 1152: update_ime_position skips sync_text_input while composing (wayland/client.rs:534), so the hint survives to client.rs:1982 and reports pre-commit text.

crates/vendor/gpui/src/platform.rs
```diff
@@ -731,6 +731,8 @@ pub trait PlatformWindow: HasWindowHandle + HasDisplayHandle {
 
     fn update_ime_position(&self, _bounds: Bounds<Pixels>);
 
+    fn set_ime_surrounding_hint(&self, _surrounding: Option<ImeSurroundingText>) {}
```
huylexuan44, line 736: Add a "mezon vendor edit:" marker. 47 exist across vendor/gpui*, and unmarked deltas get lost when the 0.2.2 snapshot is bumped.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -2055,8 +2055,15 @@ impl MentionInput {
     }
 
     fn open_popup(&mut self, tab: SubPanel, window: &mut Window, cx: &mut Context<Self>) {
+        self.input.update(cx, |input, cx| {
+            input.drop_uncommitted_preedit(cx);
+        });
+        window.reset_ime();
```
huylexuan44, line 2061: reset_ime() runs inside a Context update, so on X11 it re-enters GPUI while App is leased. Defer it with cx.defer_in.

crates/mezon-widgets/src/input.rs
```diff
@@ -238,6 +238,26 @@ impl InputState {
...
+            return;
+        };
+        let start = marked.start.min(self.content.len());
+        let end = marked.end.min(self.content.len()).max(start);
+        if start >= end {
+            return;
+        }
+        self.discard_ime_commit = self.content.get(start..end).map(str::to_string);
+        let mut next = String::with_capacity(self.content.len() - (end - start));
+        next.push_str(&self.content[..start]);
```
huylexuan44: marked offsets are only min(len)-clamped, not snapped to a char boundary; this slice panics on Vietnamese/CJK preedit.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -1775,6 +1783,15 @@ impl X11Client {
...
                 let window = self.get_window(event.event)?;
                 window.set_active(false);
+                let fcitx_shim = self
+                    .0
+                    .borrow()
+                    .im
+                    .as_ref()
+                    .is_some_and(|im| im.ibus_fcitx_shim());
+                if !fcitx_shim {
+                    self.reset_ime();
```
huylexuan44: reset_ime zeroes composing first, so the later drain_dbus_im_with(true) keep-preedit branch is now dead on blur.

crates/vendor/gpui_linux/src/linux/wayland/client.rs
```diff
@@ -515,6 +515,19 @@ impl WaylandClientStatePtr {
...
+        let client = self.get_client();
+        let composing = {
+            let state = client.borrow();
+            state.composing || state.pre_edit_text.is_some()
+        };
+        if !composing {
+            return;
+        }
+        self.disable_ime();
+        self.enable_ime();
```
huylexuan44: Neither disable_ime nor enable_ime clears pre_edit_text/compose_state, so composing stays stuck true after reset.

crates/mezon-ui/src/chat/mention_input/text_field.rs
```diff
@@ -809,7 +828,10 @@ impl MentionInputState {
         self.replace_text_in_range(None, text, window, cx);
     }
 
-    fn on_key_down(&mut self, _: &KeyDownEvent, _: &mut Window, _: &mut Context<Self>) {
+    fn on_key_down(&mut self, event: &KeyDownEvent, _: &mut Window, _: &mut Context<Self>) {
+        if event.keystroke.key == "enter" {
```
huylexuan44: Shift+Enter also hits this, leaving a stale discard_ime_commit armed; gate on a pending composition, not key == enter.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -2055,8 +2055,15 @@ impl MentionInput {
...
         let locale = self.locale(cx);
         let popup = cx.new(|cx| GifStickerEmojiPopup::new(tab, locale, window, cx));
+        popup.update(cx, |popup, cx| {
+            popup.focus_search(window, cx);
```
huylexuan44, line 2065: close_popup never refocuses the composer, so Escape or click-outside leaves the window with no focused element.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -2214,12 +2224,21 @@ impl MentionInput {
...
+        }
+    }
+
     fn insert_emoji(
         &mut self,
         emoji: EmojiSuggestRaw,
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
+        self.abandon_search_ime(window, cx);
```
huylexuan44, line 2251: insert_emoji focuses the composer but never closes the popup, so the picker stays open with an unfocused search box.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -1106,6 +1112,8 @@ impl X11Client {
         state.composing = false;
         if let Some(im) = state.im.as_mut() {
             im.reset();
+            drop(state);
+            self.drain_dbus_im();
```
huylexuan44: Wayland reset_ime early-returns when not composing; X11 does not, so every picker click pays a reset plus a full drain.

crates/mezon-ui/src/chat/mention_input/text_field.rs
```diff
@@ -334,6 +334,25 @@ impl MentionInputState {
         }
     }
 
+    pub(crate) fn drop_uncommitted_preedit(&mut self, cx: &mut Context<Self>) {
```
huylexuan44, line 338: No record_history here, so Ctrl+Z after opening the picker mid-composition skips past this deletion.

crates/mezon-widgets/src/input.rs
```diff
@@ -238,6 +238,26 @@ impl InputState {
         self.marked_range.is_some()
     }
 
+    pub fn drop_uncommitted_preedit(&mut self, cx: &mut Context<Self>) {
```
huylexuan44, line 242: Duplicate of MentionInputState::drop_uncommitted_preedit; move it to text_edit.rs next to the other IME helpers.

crates/vendor/gpui/src/window.rs
```diff
@@ -5409,19 +5409,33 @@ impl Window {
...
+    ///
+    /// mezon vendor edit: read surrounding while App is already borrowed (avoids
+    /// AsyncApp reentrancy on Linux IME sync), pass a one-shot hint into
+    /// `update_ime_position`, then always retract so blur / mid-composition /
+    /// DeleteSurrounding never see a stale snapshot.
     pub fn invalidate_character_coordinates(&self) {
         self.on_next_frame(|window, cx| {
             if let Some(mut input_handler) = window.platform_window.take_input_handler() {
                 let bounds = input_handler.selected_bounds(window, cx);
+                let surrounding = input_handler.surrounding_text_in(window, cx);
```
huylexuan44, line 5434: surrounding_text_in runs on every platform, but set_ime_surrounding_hint is a no-op off Linux; cfg-gate the read.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -2214,12 +2224,21 @@ impl MentionInput {
         Ok(())
     }
 
+    fn abandon_search_ime(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        if let Some(popup) = &self.popup {
+            popup.update(cx, |popup, cx| popup.drop_search_preedit(window, cx));
+        } else {
```
huylexuan44, line 2240: This else is unreachable: every abandon_search_ime caller runs from pick_sub, which only exists while popup is Some.

crates/vendor/gpui_linux/src/linux/wayland/window.rs
```diff
@@ -1141,6 +1144,19 @@ impl WaylandWindowStatePtr {
     }
 
     pub fn get_ime_surrounding(&self) -> Option<ImeSurroundingText> {
+        self.ime_surrounding_from_handler()
```
huylexuan44, line 1147: get_ime_surrounding has no callers left after the switch to take_ime_surrounding_for_position_sync; delete it.

crates/vendor/gpui/src/window.rs
```diff
@@ -5409,19 +5409,33 @@ impl Window {
         self.platform_window.hide();
     }
 
+    pub fn reset_ime(&self) {
```
huylexuan44, line 5415: reset_ime lacks the mezon vendor edit marker and a doc comment, unlike set_ime_surrounding_hint right below it.

crates/mezon-widgets/src/input.rs
```diff
@@ -238,6 +238,26 @@ impl InputState {
...
+        self.discard_ime_commit = self.content.get(start..end).map(str::to_string);
+        let mut next = String::with_capacity(self.content.len() - (end - start));
+        next.push_str(&self.content[..start]);
+        next.push_str(&self.content[end..]);
+        self.content = next.into();
+        self.selected_range = start..start;
+        self.refresh_filter_token_chips(cx);
+        cx.notify();
+        cx.emit(InputEvent::Change);
+    }
```
huylexuan44, line 255: No test for drop_uncommitted_preedit; a multi-byte marked-range case would have caught the char-boundary slice above.

giachuvan:
Applied
#2 / #15 — marked offsets / missing test
Partially agreed. Offsets were only min(len)-clamped, and content[..start] would panic if a marked range were off a UTF-8 boundary. The stronger claim that Vietnamese/CJK preedit always panics is overstated: range_from_utf16 already walks chars, so the normal IME path stays on boundaries. Still a real hole. I extracted splice_out_byte_range (snaps with floor_char_boundary) and added tests for "xin chào" and "你好".

#3 — reset_ime zeroes composing before the keep-preedit drain on blur
Agreed. Calling reset_ime() in X11 FocusOut set composing = false and drained with keep_preedit=false, so the later drain_dbus_im_with(true) path was dead. A picker click is an in-window focus change, not a window FocusOut. I removed reset_ime() from FocusOut so alt-tab still keeps/unmarks the preedit.

#4 — Wayland disable_ime/enable_ime leave compose state stuck
Agreed. disable_ime only clears composing; pre_edit_text, ime_pre_edit, and compose_state were left behind, so a later reset_ime still treated the client as composing. reset_ime now clears those fields and resets compose state.

#6 — close_popup never refocuses the composer
Agreed; this was a regression from focusing search on open. Escape / click-outside dropped the search FocusHandle and left the window with no focused element. close_popup now refocuses the composer.

#8 — X11 reset_ime has no idle early-return
Agreed (minor). Wayland already returns when not composing; X11 paid im.reset() + a full drain on every picker click. X11 now early-returns when idle, and also resets xkb compose when it does run.

Not applied
#1 — defer reset_ime() with cx.defer_in
Disagree. reset_ime() already runs after input.update and before focus_search. Deferring it to the end of the effect cycle would reset the search field’s IME, not the composer’s. If X11 drain re-enters while App is leased, AsyncWindowContext::update fails with .ok(); that is fine here because the preedit was already dropped.

#5 — gate on_key_down on pending composition, not key == "enter"
Disagree. key == "enter" is intentional and covers Shift+Enter: both race an IME commit. The swallow is token-specific, so a stale discard_ime_commit does not eat unrelated commits. Gating on marked_range would clear the swallow if on_key_down runs after send.

#7 — insert_emoji should close the popup
Disagree. React insertEmojiDirectly keeps the emoji picker open so you can insert more than one. Stickers/gifs/sounds already close. Matching React.

#9 — record_history in drop_uncommitted_preedit
Disagree. Composition is one undo unit (snapshot taken on the first mark). Dropping the preedit returns to that snapshot. Recording history here would make Ctrl+Z restore abandoned preedit as committed text, which is odd after the user opened the picker to leave composition.

#10 — move drop_uncommitted_preedit into text_edit.rs
Disagree. These are methods on two different structs (InputState vs MentionInputState). text_edit.rs is for pure helpers; the UTF-8 splice now lives there. The GPUI notify/emit logic cannot.

#11 — cfg-gate surrounding_text_in
Disagree as a required change. set_ime_surrounding_hint is a no-op off Linux, so the extra read is unused, not incorrect.

#12 — abandon_search_ime else is unreachable
Disagree as a required change. True for current callers (pick_sub while popup is Some). The else is defensive.

#13 — delete get_ime_surrounding
Disagree. The note points at Wayland window.rs. X11 client.rs still uses get_ime_surrounding for DeleteSurrounding.

#14 — missing vendor-edit marker / doc on reset_ime
Disagree as a functional issue. Style/convention only.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -1103,9 +1109,19 @@ impl X11Client {
 
     pub fn reset_ime(&self) {
         let mut state = self.0.borrow_mut();
+        let composing = state.composing || state.pre_edit_text.is_some();
+        if !composing {
```
huylexuan44: Regression: callers at 1193 and 1211 zero composing just before this, so the guard skips im.reset() and fcitx keeps the abandoned buffer. Reset before clearing.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -1103,9 +1109,19 @@ impl X11Client {
...
+        }
         state.composing = false;
+        if let Some(compose_state) = state.compose_state.as_mut() {
+            compose_state.reset();
+        }
+        state.pre_edit_text.take();
         if let Some(im) = state.im.as_mut() {
             im.reset();
+            drop(state);
+            self.drain_dbus_im();
```
huylexuan44: drain_dbus_im calls handle_ime_commit, which re-enters the App borrow. reset_ime runs from a view, so the commit is logged and dropped.

crates/vendor/gpui_linux/src/linux/wayland/client.rs
```diff
@@ -515,6 +515,26 @@ impl WaylandClientStatePtr {
...
+            }
+            state.pre_edit_text.take();
+            state.ime_pre_edit.take();
+            state.pending_preedit.take();
+            if let Some(compose) = state.compose_state.as_mut() {
+                compose.reset();
+            }
+        }
+        self.disable_ime();
+        self.enable_ime();
```
huylexuan44: enable_ime reads surrounding text and candidate bounds. Called from a view the App borrow fails, so the IME is re-enabled with neither.

crates/mezon-ui/src/chat/mention_input/mod.rs
huylexuan44, line 2061: cx.defer_in will not fix this: deferred effects still run inside flush_effects, under the same App borrow. Make reset_ime stop touching the input handler instead.

crates/vendor/gpui/src/window.rs
huylexuan44, line 5415: Missing doc comment fails CI Lint: gpui denies missing_docs under -D warnings. This is why the Lint job is red.

crates/mezon-widgets/src/text_edit.rs
```diff
@@ -223,6 +223,22 @@ pub fn ime_replace_range(selected: &Range<usize>, marked: Option<&Range<usize>>)
     marked.cloned().unwrap_or_else(|| selected.clone())
 }
 
+pub fn splice_out_byte_range(
+    content: &str,
+    range: Range<usize>,
+) -> Option<(String, String, usize)> {
+    let start = floor_char_boundary(content, range.start);
+    let end = floor_char_boundary(content, range.end).max(start);
```
huylexuan44: Flooring the end leaves the last preedit char behind. Check emptiness first, then use ceil_char_boundary for the end.

crates/mezon-ui/src/chat/mention_input/text_field.rs
```diff
@@ -809,7 +825,10 @@ impl MentionInputState {
         self.replace_text_in_range(None, text, window, cx);
     }
 
-    fn on_key_down(&mut self, _: &KeyDownEvent, _: &mut Window, _: &mut Context<Self>) {
+    fn on_key_down(&mut self, event: &KeyDownEvent, _: &mut Window, _: &mut Context<Self>) {
+        if event.keystroke.key == "enter" {
```
huylexuan44: Also matches Shift+Enter and Ctrl+Enter, which never send. Gate on bare Enter: key == "enter" && !modifiers.shift.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -2214,12 +2234,21 @@ impl MentionInput {
...
+        }
+    }
+
     fn insert_emoji(
         &mut self,
         emoji: EmojiSuggestRaw,
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
+        self.abandon_search_ime(window, cx);
```
huylexuan44, line 2251: open_popup now focuses the picker search box, but insert_emoji still refocuses the composer below, so the first pick kills search focus.

crates/vendor/gpui/src/window.rs
```diff
@@ -5409,19 +5409,33 @@ impl Window {
...
             if let Some(mut input_handler) = window.platform_window.take_input_handler() {
                 let bounds = input_handler.selected_bounds(window, cx);
+                let surrounding = input_handler.surrounding_text_in(window, cx);
                 window.platform_window.set_input_handler(input_handler);
                 if let Some(bounds) = bounds {
+                    window.platform_window.set_ime_surrounding_hint(surrounding);
                     window.platform_window.update_ime_position(bounds);
+                    window.platform_window.set_ime_surrounding_hint(None);
+                } else {
+                    window.platform_window.set_ime_surrounding_hint(None);
```
huylexuan44: Dead branch: take_input_handler at 5429 already cleared the hint. Drop the else.

crates/vendor/gpui_linux/src/linux/x11/client.rs
```diff
@@ -1775,6 +1791,12 @@ impl X11Client {
                 }
                 let window = self.get_window(event.event)?;
                 window.set_active(false);
+                let fcitx_shim = self
```
huylexuan44: Hoisting fcitx_shim out of the borrow below changes nothing. Drop it to keep the vendor diff minimal.

huylexuan44:
Review summary:

- Lint is red: gpui denies missing_docs under -D warnings, so Window::reset_ime needs a doc comment before the rest of the build is even reached.
- Two regressions in x11/client.rs: the new reset_ime guard turns the callers at 1193 and 1211 into no-ops (both zero composing first), and reset_ime now drains the IM queue while App is borrowed, so handle_ime_commit is dropped.
- CI never compiles the Linux code this PR is about: Lint and Test run on macos-latest, and the Linux job in build.yml only triggers on push to develop. Verified locally instead - cargo check -p mezon-app -p gpui_linux passes on Linux.
- Branch is 51 commits behind develop. Merges clean, but worth rebasing; two clippy errors in overview_tab.rs come from that staleness, not from this PR.

giachuvan:
Applied

#1: Real regression. Callers at 1193/1211 clear composing then call reset_ime(), so the idle guard skipped im.reset(). Removed the guard; always reset the engine.
#2/#3/#4: Agreed that defer_in does not help (flush_effects still holds App). reset_ime no longer touches the input handler: X11 discards DBus events after im.reset() instead of apply_im_events; Wayland re-enables the protocol without sync_text_input / get_ime_area.
#5: Added a doc comment on Window::reset_ime for missing_docs under -D warnings.
#6: Empty-range check first, ceil_char_boundary on the end; added a mid-character end test.
#7: Gated on unmodified Enter (key == "enter" && !modifiers.modified()), which also covers Ctrl+Enter, not only Shift.
#9/#10: Dropped the dead else and moved fcitx_shim back inside the FocusOut borrow.
Not applied

#8: insert_emoji focusing the composer matches React (inputEl.focus()). The picker stays open for further clicks; keyboard returns to the composer. Keeping search focused would leave the insertion target wrong.

PR #441, opened 2026-08-31
Reimplement ctrl K search call new API

giachuvan:


crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -242,6 +252,26 @@ impl CommandPaletteModal {
...
+                if text.is_empty() {
+                    CtrlKSearchStore::global(cx).update(cx, |store, cx| store.clear(cx));
+                    this.recompute_filtered(cx);
+                    this.scroll
+                        .scroll_to_item(this.selected_visible, ScrollStrategy::Top);
+                    cx.notify();
+                    return;
+                }
+                CtrlKSearchStore::global(cx).update(cx, |store, cx| {
+                    store.search(text, search_type, cx);
```
huylexuan44, line 273: Regression: server ILIKE has no unaccent, so "nguyen" no longer matches "Nguyen" with diacritics; -/_ folding is gone too.

crates/mezon-client/src/transport.rs
```diff
@@ -6837,6 +6838,31 @@ impl MezonTransport {
         Ok(api::SearchMessageResponse::decode(response.as_slice())?)
     }
 
+    pub async fn search_ctrl_k(
```
huylexuan44, line 6850: Escape % and _ here: backend builds "%"+text+"%", so a lone % returns 50 arbitrary users and private channels.

crates/mezon-app/src/main.rs
```diff
@@ -953,6 +953,7 @@ fn open_main_window(
     mezon_store::MessagesStore::init(api.clone(), cx);
     mezon_store::ThreadsStore::init(api.clone(), cx);
     mezon_store::MessageSearchStore::init(api.clone(), cx);
+    mezon_store::CtrlKSearchStore::init(api.clone(), cx);
```
huylexuan44, line 957: Also register CtrlKSearchStore in reset_all_user_stores: results survive logout into the next account.

crates/mezon-store/src/ctrlk_search.rs
```diff
@@ -0,0 +1,258 @@
...
+                    return;
+                }
+                this.state.is_searching = false;
+                match result {
+                    Ok(response) => {
+                        this.state.users = map_users(response.users);
+                        this.state.channels = map_channels(response.channels);
+                    }
+                    Err(err) => {
+                        tracing::warn!("SearchCtrlK failed: {err}");
```
huylexuan44, line 139: Err arm clears results and renders "No results". Add an error state or fall back to local items; offline Ctrl+K is dead.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -454,7 +527,14 @@ impl Render for CommandPaletteModal {
 
         let count = self.display_rows.len();
         let search_query = self.debounced_query.clone();
-        let list = if count == 0 {
+        let ctrlk_searching = self.ctrlk_searching(cx);
+        let list = if count == 0 && ctrlk_searching {
```
huylexuan44, line 526: This branch has no child, so it renders an empty 120px box. Add a spinner or text; is_searching also has no timeout.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -252,6 +282,33 @@ impl CommandPaletteModal {
...
         let previous_selection = self.selected_item_id();
+        let (api_text, _) = parse_ctrlk_query(&self.debounced_query);
+        if self.debounced_query.trim().is_empty() || api_text.is_empty() {
+            self.recompute_local_filtered(cx, previous_selection);
+            return;
+        }
+
+        let ctrlk = CtrlKSearchStore::global(cx).read(cx);
+        let state = ctrlk.state();
+        self.items = Rc::new(build_palette_items_from_ctrlk(state, cx));
```
huylexuan44, line 295: recompute_local_filtered filters stale server items. Rebuild self.items via build_palette_items on the local path.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -655,19 +655,49 @@ impl ChannelSidebar {
         }
 
         let skeleton_changed = self.advance_skeleton(cold_loading, new_clan_id, cx);
-        items_changed || name_changed || clan_changed || skeleton_changed
+        let changed = items_changed || name_changed || clan_changed || skeleton_changed;
+        self.try_scroll_ctrlk_focus(cx);
```
huylexuan44: Runs right after reset/splice while rows are Unmeasured (height 0), so this scrolls to the top instead of the target.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -655,19 +655,49 @@ impl ChannelSidebar {
...
+            false
+        }
+    }
+
+    fn try_scroll_ctrlk_focus(&mut self, cx: &mut Context<Self>) {
+        let focus = self.channel_list.read(cx).ctrlk_focus_channel().copied();
+        let Some(focus) = focus else {
+            return;
+        };
+        if self.scroll_to_channel_row(focus.channel_id, true, cx) {
```
huylexuan44: Clear ctrlk_focus on miss (collapsed category, thread, favorite): it latches, rescans every rebuild, survives logout.

crates/mezon-ui/src/command_palette/filter.rs
```diff
@@ -48,6 +67,12 @@ pub fn filter_and_sort_indices(items: &[PaletteItem], raw_query: &str) -> Vec<us
     indices
 }
 
+pub fn sort_palette_indices(items: &[PaletteItem], raw_query: &str) -> Vec<usize> {
```
huylexuan44, line 70: sort_palette_indices drops the "#" retain(kind == Channel), so a #query lists DM and Group rows returned by the server.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -326,6 +327,224 @@ pub fn build_palette_items(cx: &App) -> Vec<PaletteItem> {
...
+            continue;
+        }
+        if let Some(item) =
+            ctrlk_channel_as_clan_channel(channel, channels, clans, members_store.as_ref(), cx)
+        {
+            items.push(item);
+        }
+    }
+
+    for user in &state.users {
```
huylexuan44, line 358: Missing the dm_user_ids/dm_channel_ids dedup: one contact shows as both Direct and Member with different Enter actions.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -326,6 +327,224 @@ pub fn build_palette_items(cx: &App) -> Vec<PaletteItem> {
...
+        .unwrap_or_else(|| {
+            clans
+                .clan(channel.clan_id)
+                .map(|clan| clan.name.clone())
+                .unwrap_or_default()
+                .to_uppercase()
+        });
+    let (unread_count, last_sent_timestamp, last_seen_timestamp) = stored
+        .map(|ch| channel_list.palette_channel_unread(ch))
+        .unwrap_or((0, 0, 0));
```
huylexuan44, line 452: Unread forced to (0,0,0) when the clan is not cached, so unread channels get no badge and sort last.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -326,6 +327,224 @@ pub fn build_palette_items(cx: &App) -> Vec<PaletteItem> {
...
+        DM_PEER_CHANNEL_TYPE => DirectKind::Dm,
+        _ if channel.clan_id.is_zero() => DirectKind::Dm,
+        _ => return None,
+    };
+    let existing = dms.iter().find(|dm| dm.id == channel.channel_id);
+    let (unread_count, last_sent_timestamp, last_seen_timestamp) = channel_list
+        .user_channel(channel.channel_id)
+        .map(|ch| channel_list.palette_channel_unread(ch))
+        .unwrap_or((0, 0, 0));
+    let label = channel.label.clone();
```
huylexuan44: Use the local DirectChannel label, not raw channel_label: the store overrides server garbage (direct.rs:421).

crates/mezon-store/src/voice.rs
```diff
@@ -2119,16 +2119,15 @@ impl VoiceStore {
         }
         let channel_id = channel_id.to_string();
         let clan_id = clan_id.to_string();
-        let room_name = self.room_name.clone();
         let api = self.api.clone();
         cx.spawn(async move |this, cx| {
             let result = match action {
                 ModerationAction::Mute => {
-                    api.mute_participant_mezon_meet(&channel_id, &clan_id, &room_name, &identity)
+                    api.mute_participant_mezon_meet(&channel_id, &clan_id, &identity)
```
huylexuan44: LiveKit identity is not always numeric (agent participants), and transport_runtime now parses it as i64, so kick/mute fails.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -347,8 +415,13 @@ impl CommandPaletteModal {
                 let Some(channel_id) = item.channel_id else {
                     return;
                 };
+                let parent_id = ChannelList::global(cx)
```
huylexuan44, line 413: Use CtrlKChannel.parent_id instead of re-deriving from ChannelList, which misses for server-only threads.

huylexuan44:
Two notes that have no line in this diff to anchor on:

- messages.rs:3181 `add_to_inbox` still builds `Message2InboxRequest` with `..Default::default()`, so the `topic_id = 10` field added by this protocol bump is never sent. The optimistic row one line above sets it, the server writes 0, and the inbox entry loses its topic on the next refetch.
- login.rs `reset_all_user_stores` needs `CtrlKSearchStore` (commented inline on main.rs where the global is registered). `try_global` at ctrlk_search.rs:96 has zero call sites, which is the hook for it.

Smaller cleanups: `CtrlKSearchEvent`/`EventEmitter` has no subscribers; `from_raw`, `_members_store`, `_cx` and `CtrlKChannel::{topic,age_restricted,e2ee,category_id}` are unused; `result_generation`'s guard is unreachable; the empty-query predicate is duplicated at 4-5 sites and the observer's copy already disagrees with the others; `parse_ctrlk_query` allocates a String inside `Render`; `dms.iter().find` is O(N) where `DirectMessageStore::find` is O(1). No tests cover `CtrlKSearchStore::search`/`clear` or the `ctrlk_focus` transitions.

giachuvan:
1. Diacritics / separator folding regression (mod.rs 270–273)
Partially valid — accepted trade-off, no change.

Typed queries go to SearchCtrlK (Postgres ILIKE, no unaccent). Local-only paths (@ / # with no text, empty browse) still use normalize_search_string (diacritics + -/_ folding).

This matches React web, which also delegates typed search to the same API without client-side unaccent. Regression vs old all-local Ctrl+K only; not vs web.

2. Escape % / _ in search_ctrl_k (transport.rs)
Valid issue — needs API-side fix; client-only escape is insufficient.

Backend builds pattern := "%" + text + "%" with plain ILIKE. A lone % can match broadly.

Backslash escaping in the desktop client does not help unless the API uses ESCAPE '\' (it does not today). Fix belongs in mezon-api (sanitize/escape pattern) or reject wildcard-only input. Not fixed in desktop-only in this pass.

3. CtrlKSearchStore survives logout (main.rs / login.rs)
Valid — fixed.

Added CtrlKSearchStore::try_global → clear() in LoginStore::reset_all_user_stores.

4. API error → empty “No results” (ctrlk_search.rs 143–146)
Valid observation — intentional by design.

Hybrid spec: API failure → empty results + warn log, no local fallback (avoids mixing stale cache with server intent). Offline Ctrl+K shows no results for typed queries; @/#-only and empty browse still work locally.

We can add an explicit error state later if product wants it.

5. Empty 120px box while searching (mod.rs 530–531)
Valid — fixed.

Searching state now shows loading text (channelTopbar.loading) instead of an empty box.

Timeout: not added. is_searching clears when the RPC completes/fails; hung socket is a broader transport issue.

6. Stale server items on local path (mod.rs 291–293)
Valid — fixed.

When api_text is empty (@/#-only or empty query), recompute_filtered now rebuilds via build_palette_items() before recompute_local_filtered, so switching back from API results does not filter stale CtrlK rows.

7. Scroll before rows measured (channel_sidebar/mod.rs 658–659)
Partially valid — mitigated.

Scroll now runs in cx.defer after list rebuild (post-layout), reducing “scroll to top on height-0 rows” after reset/splice. Kept retry-on-miss for collapsed categories until the active channel row appears.

8. ctrlk_focus latches on miss / survives logout (channel_sidebar 688–691)
Partially valid — partially fixed.

Logout: fixed via CtrlKSearchStore::clear() + ctrlk_focus = None in ChannelList::reset.
Miss (collapsed category): intentional retry until row is visible; clearing immediately would break that.
parent_id on miss: not used for scroll today (scroll is by channel_id only).
9. # query lists DM/Group via API sort (filter.rs 70–71)
Valid — fixed.

sort_palette_indices now retains PaletteItemKind::Channel when query starts with #, same as filter_and_sort_indices. Test added.

10. Missing DM dedup (items.rs 359–362)
Disagree — intentional.

Local browse dedupes members with active DMs (dm_user_ids). API path deliberately does not, aligned with observed web behavior (DM row + optional user row). Product choice, not an oversight.

11. Unread (0,0,0) when clan not cached (items.rs 450–453)
Valid limitation — not fixed.

Unread comes from ChannelList cache; API does not return badge timestamps. Uncached channels sort without unread until structure loads. Acceptable for v1; enrich when API exposes unread or we prefetch channel detail.

12. DM label from server vs local store (items.rs 386–389)
Valid — fixed.

ctrlk_channel_as_direct now prefers DirectChannel.label from DirectMessageStore when present, matching direct.rs create-DM labeling.

13. LiveKit identity vs i64 user_id (voice.rs 2126–2127)
Out of Ctrl+K scope / pre-existing proto fallout.

MeetParticipantRequest now takes user_id: i64; transport_runtime parses identity as i64. Human participants use numeric user id as LiveKit identity (works). Non-numeric identities (agents) fail parse — same class of issue as give_flower. Separate voice/proto fix, not introduced by Ctrl+K UI.

14. parent_id from ChannelList vs CtrlKChannel (mod.rs 418–419)
Minor — not fixed.

parent_id is stored in CtrlKFocusChannel but scroll uses channel_id only. Re-deriving from cache misses server-only threads for metadata; no current UX impact. Can thread CtrlKChannel.parent_id through PaletteItem in a follow-up.

15. summary :
- Message2InboxRequest / topic_id: Current proto submodule has no topic_id on Message2InboxRequest. N/A until protocol bump lands; then wire msg.topic_id into the request.
- reset_all_user_stores: Fixed (see #3).
- try_global unused: 	
Used now via logout reset.
- Smaller cleanups (unused fields, no tests, duplicated empty-query checks, etc.): Acknowledged; non-blocking.n Can trim in a cleanup pass.
- CtrlKSearchEvent no subscribers: Observe via entity cx.observe on store; event emitter is optional.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -326,6 +327,227 @@ pub fn build_palette_items(cx: &App) -> Vec<PaletteItem> {
...
+    };
+    let existing = dms.iter().find(|dm| dm.id == channel.channel_id);
+    let (unread_count, last_sent_timestamp, last_seen_timestamp) = channel_list
+        .user_channel(channel.channel_id)
+        .map(|ch| channel_list.palette_channel_unread(ch))
+        .unwrap_or((0, 0, 0));
+    let label = existing
+        .map(|dm| dm.label.clone())
+        .filter(|name| !name.is_empty())
+        .unwrap_or_else(|| channel.label.clone());
```
huylexuan44, line 390: For a 1:1 DM the server channel_label is a raw <uid>,<uid> pair (core_channel.go:244), so an unresolved DM renders as ids. Skip it.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -252,6 +282,34 @@ impl CommandPaletteModal {
...
     fn recompute_filtered(&mut self, cx: &App) {
         let previous_selection = self.selected_item_id();
+        let (api_text, _) = parse_ctrlk_query(&self.debounced_query);
+        if self.debounced_query.trim().is_empty() || api_text.is_empty() {
+            self.items = Rc::new(build_palette_items(cx));
+            self.recompute_local_filtered(cx, previous_selection);
+            return;
+        }
+
+        let ctrlk = CtrlKSearchStore::global(cx).read(cx);
```
huylexuan44, line 293: Any non-empty query drops local items, and the server matches DMs on that id-pair label, so a 1:1 DM is unfindable by peer name.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -297,7 +363,10 @@ impl CommandPaletteModal {
         ChannelList::global(cx).update(cx, |store, cx| {
             items::ensure_palette_clans_loaded(store, cx);
         });
-        self.items = Rc::new(build_palette_items(cx));
+        let (api_text, _) = parse_ctrlk_query(&self.debounced_query);
+        if self.debounced_query.trim().is_empty() || api_text.is_empty() {
+            self.items = Rc::new(build_palette_items(cx));
```
huylexuan44: Dead work: recompute_filtered runs build_palette_items again under the same condition at 287. Delete 367-369.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -326,6 +327,227 @@ pub fn build_palette_items(cx: &App) -> Vec<PaletteItem> {
...
+        }
+    }
+
+    for user in &state.users {
+        if let Some(item) = ctrlk_user_as_member(user, active_clan_id, members_store.as_ref(), cx) {
+            items.push(item);
+        }
+    }
+
+    items.sort_by(cmp_items);
```
huylexuan44: This sort is thrown away: sort_palette_indices re-sorts the indices immediately after. Drop it.

crates/mezon-store/src/ctrlk_search.rs
```diff
@@ -0,0 +1,258 @@
...
+        if trimmed.is_empty() {
+            self.clear(cx);
+            return;
+        }
+
+        self.cancel_pending();
+        self.state.users.clear();
+        self.state.channels.clear();
+        self.state.is_searching = true;
+        self.result_generation = self.result_generation.wrapping_add(1);
```
huylexuan44: result_generation duplicates search_generation: both bump only here and both are checked at 133/136. Keep one.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -326,6 +327,227 @@ pub fn build_palette_items(cx: &App) -> Vec<PaletteItem> {
...
+        filter_blob: normalize_search_string(&label),
+        voice_busy: false,
+    })
+}
+
+fn ctrlk_channel_as_clan_channel(
+    channel: &CtrlKChannel,
+    channel_list: &ChannelList,
+    clans: &ClanList,
+    _members_store: Option<&Entity<ClanMembersStore>>,
```
huylexuan44: _members_store and _cx are unused, and this fn never returns None. Simplify the signature and return PaletteItem.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -252,6 +282,34 @@ impl CommandPaletteModal {
...
+            self.recompute_local_filtered(cx, previous_selection);
+            return;
+        }
+
+        let ctrlk = CtrlKSearchStore::global(cx).read(cx);
+        let state = ctrlk.state();
+        self.items = Rc::new(build_palette_items_from_ctrlk(state, cx));
+        self.filtered = Rc::new(sort_palette_indices(
+            self.items.as_ref(),
+            &self.debounced_query,
```
huylexuan44: parse_ctrlk_query trims but this passes the raw query, so a leading space (" @alice") loses both ranking and highlighting.

crates/mezon-ui/src/command_palette/mod.rs
```diff
@@ -252,6 +282,34 @@ impl CommandPaletteModal {
...
+        ));
+        self.display_rows = Rc::new(
```
huylexuan44: This duplicates the non-empty-query branch of build_display_rows (groups.rs:45). Call it instead.

giachuvan:
1 — Skip unresolved 1:1 DM from API
Accepted. Server channel_label for type-3 DM is creatorId,userId (core_channel.go:244). We now omit API 1:1 DM hits unless DirectMessageStore already has that channel (label/avatar from local store).

2 — Local items dropped on typed search
Partially accepted. By design the hybrid path replaces local browse with SearchCtrlK for non-empty text (React parity). Id-pair DM channels were not searchable by peer name on the server; fix #1 removes those rows. Peer discovery goes through the users result set. Optional follow-up: merge local Direct rows for name/label search — not in this PR unless product asks.

3 — Duplicate build_palette_items in refresh_items_if_needed
Accepted. Removed; recompute_filtered owns the local rebuild.

4 — Redundant items.sort_by before sort_palette_indices
Accepted. Removed.

5 — Duplicate generation counters
Accepted. Removed result_generation; search_generation alone cancels stale/cancelled RPCs.

6 — ctrlk_channel_as_clan_channel signature
Accepted. Simplified to return PaletteItem.

7 — Trim vs raw query for sort/highlight
Accepted. All palette query handling uses trimmed text consistently.

8 — Duplicate display-row construction
Accepted. API path uses build_display_rows like the local path.

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -655,19 +655,55 @@ impl ChannelSidebar {
         }
 
         let skeleton_changed = self.advance_skeleton(cold_loading, new_clan_id, cx);
-        items_changed || name_changed || clan_changed || skeleton_changed
+        let changed = items_changed || name_changed || clan_changed || skeleton_changed;
+        if self.channel_list.read(cx).ctrlk_focus_channel().is_some() {
+            cx.defer(|cx| {
```
huylexuan44: cx.defer still runs before the list lays out, so after list_state.reset every row is Unmeasured (height 0).

crates/mezon-ui/src/sidebar/channel_sidebar/mod.rs
```diff
@@ -655,19 +655,55 @@ impl ChannelSidebar {
...
+            false
+        }
+    }
+
+    fn try_scroll_ctrlk_focus(&mut self, cx: &mut Context<Self>) {
+        let focus = self.channel_list.read(cx).ctrlk_focus_channel().copied();
+        let Some(focus) = focus else {
+            return;
+        };
+        if self.scroll_to_channel_row(focus.channel_id, true, cx) {
```
huylexuan44: scroll_to_channel_row only checks self.items, so focus is cleared even when scroll_to_center_item landed at the top - no retry.

crates/mezon-client/src/transport.rs
```diff
@@ -4416,6 +4416,7 @@ impl MezonTransport {
             "DeletePinMessage" => 207,
             "MarkAsRead" => 208,
             "UploadBatchAttachmentFile" => 209,
+            "SearchCtrlK" => 210,
```
huylexuan44, line 4423: Index 210 exists only on mezon-api develop/hotfix2, not master. Confirm SearchCtrlK is live on the backend this release targets.

crates/mezon-ui/src/command_palette/filter.rs
```diff
@@ -48,6 +67,15 @@ pub fn filter_and_sort_indices(items: &[PaletteItem], raw_query: &str) -> Vec<us
...
+    let mut indices: Vec<usize> = (0..items.len()).collect();
+    if raw_query.starts_with('#') {
+        indices.retain(|&index| items[index].kind == PaletteItemKind::Channel);
```
huylexuan44, line 75: No Member retain for @ here, unlike filter_and_sort_indices - correct today only because the server honors type=1.

mezon-protocol
```diff
@@ -1 +1 @@
-Subproject commit ef52020b22bc7e5302c5033c1fa7e92ef135c3f2
+Subproject commit ee5c8a62c4da4eba43ad1226c0825ebe0b398dd6
```
huylexuan44, line 1: ee5c8a6 exists only on the mezon-protocol nosfu branch, not main. Merge it to main so the gitlink cannot dangle.

huylexuan44:
Re-reviewed after the SFU revert commit.

- Revert looks correct: ee5c8a6 restores both MeetParticipantRequest (api.proto) and MeetParticipantEvent (realtime.proto) to string username/room_name, identical to what develop pins. Net proto delta vs develop is additive only (SearchCtrlK messages, topic_id on Message2InboxRequest and DirectFcmProto, plus the Go-only proto/build.go removal). The param reorder to (username, room_name) is consistent across transport, transport_runtime, app_api and voice, so behaviour matches develop.
- Two things to confirm before merge, commented inline: SearchCtrlK is not on mezon-api master, and the submodule is pinned to a commit that only exists on the nosfu branch.
- Comment 7 from the first round is not actually fixed - cx.defer does not outlive list_state.reset, so the sidebar still scrolls to the top and the focus is cleared without a retry. Details inline.
- messages.rs:3181 add_to_inbox still builds Message2InboxRequest with ..Default::default(). The bump that adds topic_id = 10 is in this PR now, so it is a one-line wire-up.

PR #465, opened 2026-09-06
Fix send channel messages via writeChatMessage pipeline and fallback …

giachuvan:
…http

### DIFF
--- a/crates/mezon-client/src/transport.rs
+++ b/crates/mezon-client/src/transport.rs
@@ -5353,14 +5353,6 @@ impl MezonTransport {
     ) -> Result<ApiMessage> {
         let parsed_clan_id: i64 = clan_id;
         let parsed_channel_id: i64 = channel_id;
-        tracing::debug!(
-            "send_channel_message: clan_id={} channel_id={} is_public={} content_len={} attachments={}",
-            parsed_clan_id,
-            parsed_channel_id,
-            is_public,
-            content.len(),
-            attachments.len()
-        );
         let mention_everyone = mentions.iter().any(OutgoingMention::is_here);
         let mentions = if flags.anonymous_message {
             Vec::new()
@@ -5416,15 +5408,6 @@ impl MezonTransport {
         };
 
         let ack = self.write_or_http_channel_message(message).await?;
-        tracing::debug!(
-            "send_channel_message ack: message_id={} channel_id={} code={}",
-            ack.message_id,
-            ack.channel_id,
-            ack.code
-        );
-        if ack.message_id == 0 {
-            return Err(anyhow::anyhow!("send returned no message id"));
-        }
         let mut content_tokens = if content_is_json {
             serde_json::from_str(&content_json).unwrap_or_default()
         } else {
@@ -6355,10 +6338,7 @@ impl MezonTransport {
             mention_everyone,
             ..Default::default()
         };
-        let ack = self.write_or_http_channel_message(message).await?;
-        if ack.message_id == 0 {
-            return Err(anyhow::anyhow!("forward returned no message id"));
-        }
+        let _ack = self.write_or_http_channel_message(message).await?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


crates/mezon-client/src/transport.rs
```diff
@@ -8051,6 +8034,79 @@ impl MezonTransport {
...
+
+    async fn write_or_http_channel_message(
+        &self,
+        message: realtime::ChannelMessageSend,
+    ) -> Result<realtime::ChannelMessageAck> {
+        let has_fallback = self.http_fallback.read().is_some();
+        if self.is_open().await {
+            match self.write_chat_message(message.clone()).await {
+                Ok(ack) => return Ok(ack),
+                Err(err) if has_fallback => {
```
huylexuan44: Retrying after a socket error duplicates the message: a timeout or a closed channel means the frame already left the client. Decide up front, like send_api_request_with_http_fallback at line 3933.

crates/mezon-client/src/transport.rs
```diff
@@ -5417,18 +5413,9 @@ impl MezonTransport {
...
-            .send_api_request_with_http_fallback(cid, api_name, body)
-            .await?;
-
-        if code != 0 {
-            return Err(anyhow::anyhow!("API error: code={}", code));
-        }
+        };
 
-        let ack = realtime::ChannelMessageAck::decode(response.as_slice())?;
+        let ack = self.write_or_http_channel_message(message).await?;
```
huylexuan44, line 5410: Regression: proto-server requires a per-session channel join (403 Must join channel), where mezon-api only checked DB membership. MessagesStore::resync never re-joins the open private channel, thread or DM after a reconnect.

crates/mezon-client/src/transport.rs
```diff
@@ -8051,6 +8034,79 @@ impl MezonTransport {
...
+    }
+
+    async fn write_or_http_channel_message(
+        &self,
+        message: realtime::ChannelMessageSend,
+    ) -> Result<realtime::ChannelMessageAck> {
+        let has_fallback = self.http_fallback.read().is_some();
+        if self.is_open().await {
+            match self.write_chat_message(message.clone()).await {
+                Ok(ack) => return Ok(ack),
```
huylexuan44: ChannelMessageAck with message_id 0 is how proto-server reports a failed send (pipeline_channel.c:237), but it returns Ok here and skips the fallback.

crates/mezon-client/src/transport.rs
```diff
@@ -5351,14 +5351,10 @@ impl MezonTransport {
         ogp: Option<OutgoingOgp>,
         flags: OutgoingMessageFlags,
     ) -> Result<ApiMessage> {
-        let cid = self.generate_cid();
-
-        let api_name = "SendChannelMessage";
         let parsed_clan_id: i64 = clan_id;
         let parsed_channel_id: i64 = channel_id;
         tracing::debug!(
```
huylexuan44: Dropping cid here, plus bypassing send_api_request, removes the api_send/api_ok logs; nothing correlates a send to its frame any more.

crates/mezon-client/src/transport.rs
```diff
@@ -8051,6 +8034,79 @@ impl MezonTransport {
...
+        &self,
+        message: realtime::ChannelMessageSend,
+    ) -> Result<realtime::ChannelMessageAck> {
+        let cid = self.generate_cid();
+        let envelope = realtime::Envelope {
+            cid: i32::from(cid),
+            message: Some(realtime::envelope::Message::ChannelMessageSend(message)),
+        };
+        let (code, response) = self.send(cid, encode_envelope_cid_last(envelope)).await?;
+        if code != 0 {
```
huylexuan44: Dead check: realtime frames are always dispatched with code 0 (abridged_tcp_adapter.rs:576); only Frame::Raw carries a code.

crates/mezon-client/src/transport.rs
```diff
@@ -8051,6 +8034,79 @@ impl MezonTransport {
...
+            .context("decode ChannelMessageAck from HTTP")
+    }
+
+    async fn write_or_http_channel_message(
+        &self,
+        message: realtime::ChannelMessageSend,
+    ) -> Result<realtime::ChannelMessageAck> {
+        let has_fallback = self.http_fallback.read().is_some();
+        if self.is_open().await {
+            match self.write_chat_message(message.clone()).await {
```
huylexuan44: message.clone() on every send exists only to serve the retry; it goes away if the fallback decides up front.

crates/mezon-client/src/transport.rs
```diff
@@ -8051,6 +8034,79 @@ impl MezonTransport {
...
+    ) -> Result<realtime::ChannelMessageAck> {
+        let response = self
+            .send_api_request_over_http("SendChannelMessage", message.encode_to_vec())
+            .await
+            .context("SendChannelMessage HTTP")?;
+        realtime::ChannelMessageAck::decode(response.as_slice())
+            .context("decode ChannelMessageAck from HTTP")
+    }
+
+    async fn write_or_http_channel_message(
```
huylexuan44, line 8106: No test covers this; a_send_falls_back_to_http_while_the_socket_is_down still exercises send_api_request_with_http_fallback, which sends no longer use.

PR #466, opened 2026-09-07
Fix user status consistent in different screens

giachuvan:


crates/mezon-ui/src/components/compositions/user_info_bar.rs
```diff
@@ -217,15 +224,14 @@ impl Render for UserInfoBar {
...
-                                        .bottom_0()
-                                        .right_0()
-                                        .size_2()
-                                        .rounded_full()
-                                        .bg(status_dot_color),
-                                ),
+                                div()
+                                    .relative()
+                                    .child(avatar)
+                                    .children(presence_badge_element(
```
huylexuan44, line 225: Uses hardcoded #22c55e/#ef4444; FooterProfilePopup above it uses theme.status_* (#23a55a/#E1024F) - two colors at once.

crates/mezon-ui/src/components/compositions/user_info_bar.rs
```diff
@@ -217,15 +224,14 @@ impl Render for UserInfoBar {
...
+                                        presence_badge,
+                                        theme.bg_secondary,
```
huylexuan44: Bar renders on theme.surfaces.surface, not bg_secondary - the 2px ring shows a halo. Pass theme.tokens.bg_surface.

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -464,11 +481,29 @@ impl ProfilePage {
...
+                    .filter(|url| !url.is_empty())
+                    .map(SharedString::from)
+            });
+        let avatar_display = avatar_raw.as_ref().map(|url| {
+            SharedString::from(crate::util::imgproxy::proxied(
+                cx,
+                url.as_ref(),
+                300,
+                300,
+                "fit",
```
huylexuan44: Was profile_url (300x300 fill); fit letterboxes non-square avatars under ObjectFit::Cover. Revert to profile_url.

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -464,11 +481,29 @@ impl ProfilePage {
...
+                AccountStore::global(cx)
+                    .read(cx)
+                    .account
+                    .as_ref()
+                    .and_then(|account| account.avatar_url.clone())
+                    .filter(|url| !url.is_empty())
+                    .map(SharedString::from)
+            });
+        let avatar_display = avatar_raw.as_ref().map(|url| {
+            SharedString::from(crate::util::imgproxy::proxied(
```
huylexuan44: refresh_banner_color still uses profile_url, so the same avatar is now fetched through two different proxy URLs.

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -115,6 +115,22 @@ impl ProfilePage {
                     this.profile = Some(ProfileState::from_account(account));
                     this.refresh_banner_color(cx);
                     cx.notify();
+                } else if !this.is_dirty() {
```
huylexuan44, line 118: Duplicates the AccountLoaded handler but skips its input reset, so display_name/about_me inputs keep stale text.

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -169,6 +185,7 @@ impl ProfilePage {
                     if let Some(state) = &mut this.profile {
                         state.avatar_url = Some(url.clone().into());
                     }
+                    this.avatar_local_preview = None;
```
huylexuan44: Clearing the local preview here swaps to an imgproxy URL the proxy has not cached - avatar blinks to initials.

crates/mezon-ui/src/sidebar/direct_sidebar.rs
```diff
@@ -136,12 +137,12 @@ fn dm_in_voice(ch: &DirectChannel, channels: &ChannelList) -> bool {
...
 }
 
-fn dm_presence_badge(ch: &DirectChannel, presence: &PresenceStore) -> DmAvatarPresence {
+fn dm_presence_badge(ch: &DirectChannel, presence: &PresenceStore, cx: &App) -> DmAvatarPresence {
     if ch.kind != DirectKind::Dm {
         return DmAvatarPresence::None;
     }
     ch.peer_user_id
-        .map(|user_id| presence.dm_avatar_presence(user_id, ch.online))
+        .map(|user_id| presence.dm_avatar_presence(user_id, ch.online, current_user_presence(cx)))
```
huylexuan44: current_user_presence(cx) runs per DM row; hoist it once like member_list.rs:622 and friends_page.rs:334.

crates/mezon-ui/src/components/compositions/user_info_bar.rs
```diff
@@ -132,7 +134,12 @@ impl UserInfoBar {
 impl Render for UserInfoBar {
     fn render(&mut self, _window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
         let theme = cx.theme();
-        let status_dot_color = status_color(self.status, theme);
+        let presence_badge = match self.status {
```
huylexuan44: presence_badge() in mezon-store already does this mapping - make it pub or add a From impl instead.

crates/mezon-store/src/presence.rs
```diff
@@ -272,7 +272,15 @@ impl PresenceStore {
...
     }
 
-    pub fn dm_avatar_presence(&self, user_id: UserId, api_online: bool) -> DmAvatarPresence {
+    pub fn dm_avatar_presence(
+        &self,
+        user_id: UserId,
+        api_online: bool,
+        own: Option<(UserId, UserPresence)>,
+    ) -> DmAvatarPresence {
+        if let Some((_, presence)) = own.filter(|(me, _)| *me == user_id) {
```
huylexuan44, line 292: Only fires when peer_user_id == me; for a DM that means dm_peer_index picked self. Which screen did this fix?

crates/mezon-ui/src/components/compositions/footer_profile_popup.rs
```diff
@@ -279,23 +308,51 @@ impl Render for FooterProfilePopup {
...
+                    }
+                    avatar
+                })
+                .children(match self.status {
+                    UserPresence::Idle => Some(
+                        div()
+                            .absolute()
+                            .bottom(px(4.))
+                            .right(px(6.))
+                            .size(px(20.))
```
huylexuan44, line 331: Idle badge is 20px, Online/Dnd 18px - the badge resizes when switching status.

crates/mezon-ui/src/util/user_status.rs
```diff
@@ -76,6 +76,24 @@ pub fn status_icon(presence: UserPresence) -> IconName {
     }
 }
 
+pub fn status_glyph(
```
huylexuan44, line 83: status_icon_and_color has no callers left after this PR - delete it.

giachuvan:
1. Hardcoded 
#22c55e / 
#ef4444 vs theme.status_*
Skipped. Those hardcodes match React StatusUser (bg-green-500 / Tailwind green-500). Theme status_online / status_dnd (#23a55a / #E1024F) are for status-menu icons. Switching avatar dots to theme tokens would diverge from React across every DM/member badge, so we kept the existing presence-dot colors.

2. Ring uses bg_secondary → halo
Fixed. The footer bar sits on theme.surfaces.surface (tokens.bg_surface). theme.bg_secondary is remapped to the DM theme color, so the 2px ring showed a halo. Ring color is now theme.tokens.bg_surface.

3. Preview proxied with fit 300×300
Fixed. Reverted avatar display to profile_url (300×300 fill) so non-square avatars are not letterboxed under ObjectFit::Cover.

4. Banner still used profile_url → two proxy URLs
Fixed by (3). Preview and banner now share the same profile_url proxy path.

5. observe !is_dirty duplicates AccountLoaded but skips input reset
Fixed. When syncing from account while clean, we now clear display_name_input / about_me_input (and their subscriptions) the same way as AccountLoaded, so inputs don’t keep stale text.

6. Clear avatar_local_preview on UserAvatarUploaded → blink
Fixed. Stopped clearing the local preview on upload success so the UI doesn’t briefly fall back to an uncached imgproxy URL / initials. Preview is still cleared on upload failure / discard / AccountLoaded as before.

7. current_user_presence(cx) per DM row
Fixed. Hoisted once in dm_items_fingerprint and build_dm_items, same pattern as member list / friends page.

8. Duplicate UserPresence → DmAvatarPresence mapping
Fixed. Added impl From<UserPresence> for DmAvatarPresence in mezon-store; UserInfoBar uses DmAvatarPresence::from(self.status) instead of an inline match.

9. own-presence branch only when peer == me — which screen?
Intentional for self-DM. When the DM peer is the signed-in user, React’s useMemberStatus prefers account status over socket presence. Without this, a self-DM row (and matching channel header) could show Online while the account is Idle/Invisible. The branch does not change badges for other peers.

10. Idle badge 20px vs Online/Dnd 18px
Fixed. Online/Dnd avatar badges in the footer popup are now 20px so size doesn’t jump when switching status.

11. status_icon_and_color unused
Fixed. Removed; no remaining callers after this PR.

crates/mezon-ui/src/settings/profile_page.rs
```diff
@@ -115,6 +115,25 @@ impl ProfilePage {
...
+                            || prev.display_name != next.display_name
+                            || prev.about_me != next.about_me
+                            || prev.logo_url != next.logo_url
+                            || prev.username != next.username
+                            || prev.status != next.status
+                            || prev.custom_status != next.custom_status
+                    });
+                    if changed {
+                        this.profile = Some(next);
+                        this.display_name_input = None;
```
huylexuan44: changed includes status, so a status push drops both inputs and steals focus. Gate the reset on the text fields only.

crates/mezon-ui/src/components/compositions/user_info_bar.rs
huylexuan44: Still open after a9912c93 - presence_badge_color is unchanged, so the bar and the popup still disagree on Online/Dnd.

PR #470, opened 2026-09-08
Fix logic reactive archived thread

giachuvan:


crates/mezon-store/src/channel.rs
```diff
@@ -4028,6 +4026,46 @@ impl ChannelList {
...
+        channel_id: ChannelId,
+        clan_id: ClanId,
+        mode: i32,
+        cx: &App,
+    ) -> bool {
+        let state = self.reactivate_state_for_send(channel_id, clan_id, mode, cx);
+        state.should_reactivate && state.is_public
+    }
+
+    pub fn maybe_reactivate_after_send(
```
huylexuan44: Regression: thread_needs_reactivate is true whenever last_sent_timestamp is older than 7 days, even when active == JOINED, and apply_thread_reactivated never clears it. Stale threads now fire ActiveArchivedThread twice per message instead of once. Bump the cached last_sent_timestamp when your own send is confirmed.

crates/mezon-store/src/channel.rs
```diff
@@ -4772,17 +4811,38 @@ impl ChannelList {
...
+        let channel = self.channel(clan_id, channel_id);
+
+        let (is_thread, is_public, channel_archived, sync_cached_archived) = match channel {
+            Some(ch) => (
+                ch.is_thread(),
+                !ch.private,
+                thread_needs_reactivate(ch),
+                threads_archived && !ch.is_archived(),
+            ),
+            None => (true, false, false, false),
```
huylexuan44: apply_local_archive drops the channel from the cache, so an archived public thread always lands here with is_public false and the pre-send AddChannelUsers never runs. Read channel_private from ThreadsStore instead.

crates/mezon-store/src/channel.rs
```diff
@@ -4772,17 +4811,38 @@ impl ChannelList {
...
-            should_reactivate: should_reactivate_thread_after_send(mode, &channel, archived),
-            sync_cached_archived: threads_archived && !channel.is_archived(),
-        })
+        let locally_archived = self.is_locally_archived(channel_id);
+        let channel = self.channel(clan_id, channel_id);
+
+        let (is_thread, is_public, channel_archived, sync_cached_archived) = match channel {
+            Some(ch) => (
+                ch.is_thread(),
+                !ch.private,
```
huylexuan44: thread_channel_from_context hardcodes private: false, and that is how ChatLayout caches a thread opened from the popover, so private threads read as public here. Channel.private is not a reliable source.

crates/mezon-store/src/messages.rs
```diff
@@ -5269,6 +5271,7 @@ impl MessagesStore {
                             }
                             this.reconcile_temp(channel_id, temp_id, confirmed, cx);
                         });
+                        maybe_reactivate_archived_thread(&this, channel_id, clan_id, mode, cx);
```
huylexuan44: maybe_reactivate_archived_thread only runs in the Ok arm, so a send rejected because the user is not in the thread never triggers reactivation and the thread stays stuck.

crates/mezon-store/src/messages.rs
```diff
@@ -6720,6 +6725,9 @@ async fn ensure_archived_thread_reactivated(
     let needs = this
         .update(cx, |_this, cx| {
             ChannelList::global(cx).update(cx, |list, cx| {
+                if !list.needs_public_pre_send_reactivate(channel_id, clan_id, mode, cx) {
```
huylexuan44: needs_public_pre_send_reactivate and begin_reactivate_for_send each recompute reactivate_state_for_send: two ThreadsStore global reads and two cache lookups per send. Compute the state once.

crates/mezon-store/src/channel.rs
```diff
@@ -4028,6 +4026,46 @@ impl ChannelList {
...
+        clan_id: ClanId,
+        mode: i32,
+        cx: &mut Context<Self>,
+    ) {
+        if !self.begin_reactivate_for_send(channel_id, clan_id, mode, cx) {
+            return;
+        }
+
+        let api = self.api.clone();
+        cx.spawn(async move |this, cx| {
```
huylexuan44: This duplicates the RPC plus apply/finish body of ensure_public_archived_thread_reactivated in another module. Collapse to one helper.

crates/mezon-store/src/channel.rs
```diff
@@ -10180,25 +10240,25 @@ mod tests {
         thread.active = CHANNEL_ACTIVE_ARCHIVED;
         assert!(should_reactivate_thread_after_send(
             6,
-            &thread,
+            true,
```
huylexuan44: Passing a literal true drops the only coverage of Channel::is_thread here; the assert now reduces to 6 == 6 && true && archived. No test covers needs_public_pre_send_reactivate, maybe_reactivate_after_send, the locally_archived term or the None fallback.

giachuvan:
Comment 1 (thread_needs_reactivate / double ActiveArchivedThread)
Addressed (option A).
apply_thread_reactivated now bumps last_sent_timestamp to now, and successful sends call note_own_send_confirmed before post-send reactivate. That clears the stale predicate so we don’t fire ActiveArchivedThread twice on the same / following message. Covered by note_own_send_clears_stale_so_second_reactivate_skips and apply_thread_reactivated_bumps_stale_last_sent.

Comment 2 (None after apply_local_archive → is_public: false)
Addressed.
is_public no longer comes from the (missing) Channel row. We read channel_private from ThreadsStore, so a locally archived public thread still gets pre-send reactivate. Covered by public_pre_send_uses_threads_store_privacy_after_local_archive.

Comment 3 (thread_channel_from_context hardcodes private: false)
Addressed.
Same change: privacy for reactivate gating is taken from ThreadsStore::thread_channel_private, not Channel.private. Private threads no longer look public and skip public pre-send. Covered by private_thread_skips_public_pre_send_even_when_cache_says_public.

Comment 4 (reactivate on send Err)
Rejected (by design).
We only reactivate after a successful send (aligned with product plan and React’s success-path activation). Membership / public pre-send handles “not in thread” before send; calling ActiveArchivedThread on send failure would change that contract.

Comment 5 (double reactivate_state_for_send)
Addressed.
Public pre-send uses begin_public_pre_send_reactivate, which computes state once and then begin_reactivate_with_state.

Comment 6 (duplicate RPC + apply/finish)
Addressed.
Shared invoke_active_archived_thread + ChannelList::complete_reactivate for both pre-send and post-send paths.

Comment 7 (weak tests / literal true)
Addressed.
Dropped the tautology assert; tests now use Channel::is_thread() / STREAM_MODE_THREAD. Added coverage for public pre-send privacy, private skip, locally-archived None cache, stale clear via own send, and last_sent bump on reactivate.

crates/mezon-store/src/channel.rs
```diff
@@ -4028,6 +4050,43 @@ impl ChannelList {
...
+        message_id: MessageId,
+        cx: &mut Context<Self>,
+    ) {
+        let Some(ch) = self.channel_mut(clan_id, channel_id) else {
+            return;
+        };
+        if ts <= ch.last_sent_timestamp {
+            return;
+        }
+        ch.last_sent_timestamp = ts;
```
huylexuan44: Regression: last_seen_timestamp is not advanced with it, so Channel::is_unread() flips true for every channel you send into and the sidebar row plus the clan dot light up on your own message. Bump last_seen_timestamp and last_seen_message_id here too.

crates/mezon-store/src/channel.rs
```diff
@@ -4772,17 +4840,40 @@ impl ChannelList {
...
+            };
+        }
+
+        let channel_key = channel_id.to_string();
+        let threads = crate::threads::ThreadsStore::global(cx).read(cx);
+        let threads_archived = threads.thread_active(&channel_key) == Some(CHANNEL_ACTIVE_ARCHIVED);
+        let is_public = threads
+            .thread_channel_private(&channel_key)
+            .map(|private| private == 0)
+            .unwrap_or(false);
```
huylexuan44, line 4924: ThreadsStore::on_active_channel_changed clears self.threads, and refresh() only runs on ThreadCreated, so after navigating into the thread thread_channel_private returns None and is_public is false again. Verified: privacy is Some(0) after the popover loads, None once the thread becomes active, and begin_public_pre_send_reactivate returns false. Also, dropping the !ch.private fallback now misreads a correctly cached public channel. Suggest or_else(|| channel.map(|ch| !ch.private)).

crates/mezon-store/src/channel.rs
```diff
@@ -10722,6 +10816,157 @@ mod tests {
...
+        cx: &mut gpui::TestAppContext,
+    ) {
+        cx.update(|cx| {
+            let channels = init_channel_list_with_threads(cx);
+            crate::threads::ThreadsStore::global(cx).update(cx, |store, cx| {
+                store.seed_threads_for_test("1", vec![seed_archived_thread_summary(0)], cx);
+            });
+            channels.update(cx, |channels, cx| {
+                channels.apply_clan_structure(ClanId(1), structure_with_a_thread(), None, cx);
+                channels.apply_local_archive(ClanId(1), ChannelId(9), ChannelId(1), cx);
```
huylexuan44, line 10916: This passes only because it asserts in the same update block as the seed. Add the active channel change that ChatLayout::navigate_to_thread triggers before asserting and it fails.

giachuvan:
Comment 1 (last_seen not advanced)
Addressed. note_own_send_confirmed now advances last_seen_timestamp / last_seen_message_id with the send (and keeps last_seen >= last_sent). apply_thread_reactivated does the same when bumping last_sent, so own sends don’t mark the row/clan unread. Covered in note_own_send_clears_stale_so_second_reactivate_skips.

Comment 2 (thread_channel_private None after navigate)
Addressed. is_public is now ThreadsStore privacy, then channel.map(|ch| !ch.private). ChatLayout::navigate_to_thread reads privacy while the popover list still has it and passes it into ensure_thread_*, so the cached Channel.private stays correct after on_active_channel_changed clears ThreadsStore.

Comment 3 (test only passed in same update as seed)
Addressed. public_pre_send_survives_threads_store_clear_after_navigate seeds from ThreadsStore, ensures the channel with that privacy, simulates the active-channel clear, asserts privacy is None, then asserts begin_public_pre_send_reactivate still succeeds via the cached channel fallback.

PR #477, opened 2026-09-09
Improve UX for sending long text message

giachuvan:


crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1051,21 +1068,17 @@ impl MentionInput {
...
-            if json_string_utf16_len(&combined) > CONVERT_TO_FILE_THRESHOLD {
-                self.committed.clear();
-                self.reset_popup();
-                self.input.update(cx, |input, cx| {
-                    input.set_mention_spans(Vec::new(), cx);
-                    input.set_value("", window, cx);
-                });
-                self.convert_text_to_file(combined, window, cx);
-                cx.notify();
+            if should_convert_paste_to_file(&current, &text) {
```
huylexuan44, line 1076: Removes the current+pasted cap: repeated sub-threshold pastes now grow the composer unbounded. Keep a cap.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,27 @@ impl MentionInput {
...
-        if self.overflow_to_file
+        let (text, content, ogp) = if self.overflow_to_file
             && !text.is_empty()
             && content_payload_utf8_len(&text) > CONVERT_TO_FILE_THRESHOLD
         {
-            self.convert_text_to_file(text, window, cx);
-            self.committed.clear();
-            self.reset_popup();
-            self.close_popup(window, cx);
+            let pending = write_text_as_pending_attachment(&text)?;
```
huylexuan44: Sync fs::write plus build_pending metadata on the foreground thread; the old path used background_spawn.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -520,6 +520,20 @@ fn content_payload_utf8_len(text: &str) -> usize {
...
+fn should_convert_paste_to_file(current_input: &str, pasted: &str) -> bool {
+    current_input.trim().is_empty() && json_string_utf16_len(pasted) > CONVERT_TO_FILE_THRESHOLD
+}
+
+fn write_text_as_pending_attachment(text: &str) -> Option<PendingAttachment> {
+    let filename = format!("{}.txt", chrono::Utc::now().timestamp_millis());
+    let path = std::env::temp_dir().join(&filename);
+    if let Err(err) = std::fs::write(&path, text.as_bytes()) {
+        tracing::warn!("failed to write converted text attachment: {err}");
+        return None;
```
huylexuan44, line 532: Write failure only warns; take_payload ? makes Enter a silent no-op. Surface an error toast instead.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,27 @@ impl MentionInput {
...
+        let (text, content, ogp) = if self.overflow_to_file
             && !text.is_empty()
             && content_payload_utf8_len(&text) > CONVERT_TO_FILE_THRESHOLD
         {
-            self.convert_text_to_file(text, window, cx);
-            self.committed.clear();
-            self.reset_popup();
-            self.close_popup(window, cx);
+            let pending = write_text_as_pending_attachment(&text)?;
+            if let Err(limit) = validate_batch(
```
huylexuan44: Temp .txt is written before validate_batch, so a rejected batch leaks the file. Check the count first.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,27 @@ impl MentionInput {
...
+                self.pending_attachments.len(),
+                std::slice::from_ref(&pending),
+            ) {
+                Self::show_upload_limit(limit, window, cx);
+                return None;
+            }
+            self.pending_attachments.push(pending);
```
huylexuan44: Duplicates add_pending (validate_batch + push + show_upload_limit). Reuse it, or share one async convert path.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -520,6 +520,20 @@ fn content_payload_utf8_len(text: &str) -> usize {
         .unwrap_or(text.len() + CONVERT_PREFIX_LEN)
 }
 
+fn should_convert_paste_to_file(current_input: &str, pasted: &str) -> bool {
+    current_input.trim().is_empty() && json_string_utf16_len(pasted) > CONVERT_TO_FILE_THRESHOLD
```
huylexuan44, line 524: Uses UTF-16 JSON length while take_payload uses UTF-8 bytes; Vietnamese text converts at send with no warning.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -520,6 +520,20 @@ fn content_payload_utf8_len(text: &str) -> usize {
...
+}
+
+fn write_text_as_pending_attachment(text: &str) -> Option<PendingAttachment> {
+    let filename = format!("{}.txt", chrono::Utc::now().timestamp_millis());
```
huylexuan44, line 528: Bare millisecond name in the shared temp dir can collide; prefix it like write_clipboard_image does.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1051,21 +1068,17 @@ impl MentionInput {
...
-                self.committed.clear();
-                self.reset_popup();
-                self.input.update(cx, |input, cx| {
-                    input.set_mention_spans(Vec::new(), cx);
-                    input.set_value("", window, cx);
-                });
-                self.convert_text_to_file(combined, window, cx);
-                cx.notify();
+            if should_convert_paste_to_file(&current, &text) {
+                if !current.is_empty() {
```
huylexuan44, line 1077: Unreachable except for whitespace-only input, where only set_value("") does anything. Drop the block.

giachuvan:
1. Paste cap / unbounded composer

Intentional. Per product UX we only auto-convert when the composer is empty and the pasted chunk alone exceeds the threshold. While the user is already typing, additional pastes append and the red remaining-character countdown updates (including negative). Conversion + send happens on Enter. Restoring the old current+pasted convert-on-paste would undo that UX.

2. Sync fs::write on the foreground thread

Intentional. Enter must convert and send in a single action. The previous background_spawn path cleared the draft and returned None, forcing a second Enter after the pending attachment appeared. For plain .txt the write + build_pending cost is small (no media probe), so sync write is an acceptable tradeoff for one-shot send.

3. Write failure → silent no-op

Fixed. On write failure we now show an error toast and keep the draft instead of returning None silently via ?.

4. Temp file written before validate_batch

Fixed. We pre-check attachment count and payload size against the upload limits before writing the temp file, then build the pending attachment and add it via add_pending.

5. Duplicate add_pending logic

Fixed. The send path now reuses add_pending after a successful write (with the pre-checks above so we don’t leak a temp file on a full batch).

6. UTF-16 paste vs UTF-8 send metrics

Intentional / keep as-is. This matches the React client: paste uses JSON.stringify UTF-16 length; send uses UTF-8 bytes of {"t": text}. Unifying metrics would diverge from web parity. The countdown uses the same UTF-16 content measure as before.

7. Bare millisecond temp filename

Won’t change for now. Same naming pattern as the previous convert-to-file path; collision risk in the same millisecond is accepted. We can revisit if we see real collisions.

8. Whitespace-only if !current.is_empty() block

Keeping. The block is reachable when the composer is whitespace-only (trim() empty but not is_empty()). Clearing that whitespace before converting an oversized paste avoids leftover spaces in the input.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1051,21 +1079,17 @@ impl MentionInput {
...
-            if json_string_utf16_len(&combined) > CONVERT_TO_FILE_THRESHOLD {
-                self.committed.clear();
-                self.reset_popup();
-                self.input.update(cx, |input, cx| {
-                    input.set_mention_spans(Vec::new(), cx);
-                    input.set_value("", window, cx);
-                });
-                self.convert_text_to_file(combined, window, cx);
-                cx.notify();
+            if should_convert_paste_to_file(&current, &text) {
```
huylexuan44, line 1076: Regression: with a draft, big pastes now enter the editor and re-shape all text per keystroke. Restore a paste-time cap.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +840,37 @@ impl MentionInput {
...
-            self.close_popup(window, cx);
+            if self.pending_attachments.len() + 1 > MAX_FILE_ATTACHMENTS {
+                Self::show_upload_limit(AttachmentLimit::Count, window, cx);
+                return None;
+            }
+            if text.len() as u64 > MAX_FILE_SIZE {
+                Self::show_upload_limit(AttachmentLimit::Size(MAX_FILE_SIZE), window, cx);
+                return None;
+            }
+            let Some(pending) = write_text_as_pending_attachment(&text) else {
```
huylexuan44, line 846: Sync fs::write plus metadata on the foreground thread; keep it in background_spawn per performance-core rules.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +840,37 @@ impl MentionInput {
...
         {
-            self.convert_text_to_file(text, window, cx);
-            self.committed.clear();
-            self.reset_popup();
-            self.close_popup(window, cx);
+            if self.pending_attachments.len() + 1 > MAX_FILE_ATTACHMENTS {
+                Self::show_upload_limit(AttachmentLimit::Count, window, cx);
+                return None;
+            }
+            if text.len() as u64 > MAX_FILE_SIZE {
```
huylexuan44: Dead check: MAX_FILE_SIZE is 1GiB. Both pre-checks duplicate validate_batch in add_pending - drop them.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +840,37 @@ impl MentionInput {
...
+            }
+            let Some(pending) = write_text_as_pending_attachment(&text) else {
+                Shell::global(cx).update(cx, |shell, cx| {
+                    shell.toast(ToastKind::Error, "Failed to convert message to file", cx)
+                });
+                return None;
+            };
+            let before = self.pending_attachments.len();
+            self.add_pending(vec![pending], window, cx);
+            if self.pending_attachments.len() == before {
```
huylexuan44, line 855: Temp .txt is left behind in temp_dir on this reject path - remove_file before returning.

giachuvan:
1. Paste-time cap with a draft

Intentional — not a regression vs the agreed UX. When the composer already has a draft, additional pastes append and the red remaining-character countdown updates (including negative values). We only auto-convert on paste when the composer is empty and the pasted chunk alone exceeds the threshold. Oversized in-composer text is converted and sent on Enter. Restoring the old current+pasted convert-on-paste would undo that product behavior.

2. Sync fs::write on the foreground thread

Intentional tradeoff for one-shot Enter. Convert + send must happen in a single keypress; the previous background_spawn path cleared the draft and returned None, forcing a second Enter after the pending attachment appeared. For plain .txt, write + build_pending is small (no media probe). Moving this back to a background task would require a deferred send pipeline after the file is ready, which we are not doing in this change.

3. Dead / duplicate pre-checks before write

Fixed. Dropped both pre-checks. The size check against MAX_FILE_SIZE (1GiB) was effectively dead for chat text, and both checks duplicated validate_batch inside add_pending. Limits are enforced solely via add_pending again.

4. Temp .txt leaked on reject

Fixed. If add_pending rejects after a successful write, we remove_file the temp path before returning None, so a full attachment batch no longer leaks a file in temp_dir.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,31 @@ impl MentionInput {
...
-        if self.overflow_to_file
+        let (text, content, ogp) = if self.overflow_to_file
             && !text.is_empty()
             && content_payload_utf8_len(&text) > CONVERT_TO_FILE_THRESHOLD
         {
-            self.convert_text_to_file(text, window, cx);
-            self.committed.clear();
-            self.reset_popup();
-            self.close_popup(window, cx);
+            let Some(pending) = write_text_as_pending_attachment(&text) else {
```
huylexuan44, line 846: Sync fs write + stat on the GPUI foreground thread inside the Enter action; base used background_spawn. Write in background, then send.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1051,21 +1072,17 @@ impl MentionInput {
...
-            if json_string_utf16_len(&combined) > CONVERT_TO_FILE_THRESHOLD {
-                self.committed.clear();
-                self.reset_popup();
-                self.input.update(cx, |input, cx| {
-                    input.set_mention_spans(Vec::new(), cx);
-                    input.set_value("", window, cx);
-                });
-                self.convert_text_to_file(combined, window, cx);
-                cx.notify();
+            if should_convert_paste_to_file(&current, &text) {
```
huylexuan44: Append here is the agreed UX. Still worth a large safety cap (e.g. ~100k chars): the field shapes the whole draft every frame, so a multi-MB paste freezes the UI.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,31 @@ impl MentionInput {
...
+            }
             self.clear_ogp_preview(cx);
-            self.input.update(cx, |input, cx| {
-                input.set_mention_spans(Vec::new(), cx);
-                input.clear_after_send(swallow.clone(), window, cx);
-            });
-            return None;
-        }
-        let content = outgoing_content_from_committed(&raw, &self.committed);
+            (String::new(), OutgoingContent::default(), None)
```
huylexuan44: Whole draft goes into the .txt, so @mention/#hashtag tokens lose their metadata and no mention notification fires. Confirm this is acceptable for the agreed flow.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,31 @@ impl MentionInput {
         } else {
             raw.trim_end().to_string()
         };
-        if self.overflow_to_file
+        let (text, content, ogp) = if self.overflow_to_file
             && !text.is_empty()
             && content_payload_utf8_len(&text) > CONVERT_TO_FILE_THRESHOLD
```
huylexuan44, line 844: Enter converts on UTF-8 bytes but counter and paste gate use UTF-16 units, so ~1300 Vietnamese chars get sent as .txt with no warning. Use one predicate.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,31 @@ impl MentionInput {
         } else {
```
huylexuan44, line 839: No anonymous-mode gate: the .txt URL embeds the real user id and the chip checkpoint is gone. Refuse conversion in anonymous mode like buzz/location.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1074,17 +1091,9 @@ impl MentionInput {
...
-                    let path = std::env::temp_dir().join(&filename);
-                    if let Err(err) = std::fs::write(&path, text.as_bytes()) {
-                        tracing::warn!("failed to write converted text attachment: {err}");
-                        return None;
-                    }
-                    build_pending(path)
-                })
+                .background_spawn(async move { write_text_as_pending_attachment(&text) })
                 .await;
             if let Some(pending) = pending {
```
huylexuan44, line 1098: Paste path still has no failure toast and leaks the temp .txt when add_pending rejects; mirror the Enter-path handling here.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +839,31 @@ impl MentionInput {
...
-            self.committed.clear();
-            self.reset_popup();
-            self.close_popup(window, cx);
+            let Some(pending) = write_text_as_pending_attachment(&text) else {
+                Shell::global(cx).update(cx, |shell, cx| {
+                    shell.toast(ToastKind::Error, "Failed to convert message to file", cx)
+                });
+                return None;
+            };
+            let path = pending.path.clone();
```
huylexuan44, line 852: Writes and stats the file before validating, then infers rejection from pending_attachments.len(). Validate the count first or return Result from add_pending.

crates/mezon-ui/src/chat/mention_input/mod.rs
huylexuan44, line 528: Bare {millis}.txt in the shared temp dir collides for two conversions in one ms; remove_file below can then delete the other file. Prefix with pid + counter.

giachuvan:
1. Sync write on Enter

Intentional. Enter must convert and send in one keypress. Moving the write to background_spawn would defer send until the file is ready and regresses to a two-step flow. Paste conversion still uses background_spawn.

2. Large paste safety cap

Done. Appending into a draft is unchanged below the cap. If current + pasted would exceed 100k UTF-16 chars, we trim the paste chunk only and keep the existing draft. Empty-composer pastes over 3700 still convert to .txt (not trimmed).

3. Mentions / hashtags lost in .txt

Confirmed acceptable. The converted file is plain text, same as the web convert-to-file flow. Mention/hashtag metadata and mention notifications are not preserved.

4. UTF-16 vs UTF-8 predicates

Keeping React parity: paste/countdown use UTF-16 (JSON.stringify length); send uses UTF-8 bytes of {"t": text}. Unifying would diverge from web.

5. Anonymous mode

Done. Convert-to-file and send are refused in anonymous mode (toast), because the attachment can carry sensitive content and identity is not controlled. Oversized paste in anonymous mode stays in the composer (trimmed to the 100k cap) instead of becoming a file. Web will be updated separately.

6. Paste-path toast / leak

Fixed. Write failure shows the same error toast as Enter. If add_pending rejects, we remove_file the temp .txt.

7. Validate before inferring rejection

Fixed. Attachment count is checked before writing on Enter. add_pending now returns bool; both Enter and paste use that instead of inferring from pending_attachments.len().

8. Temp filename collision

Fixed. Converted files are named converted-text-{pid}-{counter}.txt so a reject-path remove_file cannot delete another conversion from the same millisecond.

assets/i18n/en.json
```diff
@@ -1743,6 +1743,7 @@
   "common.cannotFetchClans": "Cannot fetch clans, please try again later",
   "common.cannotJoinClan": "Cannot join clan",
   "common.cannotSendBuzzWithAnonymous": "Cannot send buzz using Anonymous.",
+  "common.cannotSendConvertedFileWithAnonymous": "Cannot convert and send a file using Anonymous.",
```
huylexuan44, line 1746: Key added only to en/vi; the other 14 locales miss it, so mezon-i18n every_locale_carries_every_english_key fails in CI (ru first). Add it to all 16 bundles.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -825,24 +867,35 @@ impl MentionInput {
...
-            self.close_popup(window, cx);
+            if MessagesStore::global(cx).read(cx).is_anonymous_mode() {
+                self.show_anonymous_convert_blocked(cx);
+                return None;
+            }
+            if self.pending_attachments.len() + 1 > MAX_FILE_ATTACHMENTS {
+                Self::show_upload_limit(AttachmentLimit::Count, window, cx);
+                return None;
+            }
+            let Some(pending) = write_text_as_pending_attachment(&text) else {
```
huylexuan44: One-key send does not need a sync write: clear the draft, run convert_text_to_file, and cx.emit(Submit) after add_pending; take_payload then sends the pending .txt with empty text.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -520,6 +522,46 @@ fn content_payload_utf8_len(text: &str) -> usize {
         .unwrap_or(text.len() + CONVERT_PREFIX_LEN)
 }
 
+fn should_convert_paste_to_file(current_input: &str, pasted: &str) -> bool {
+    current_input.trim().is_empty() && json_string_utf16_len(pasted) > CONVERT_TO_FILE_THRESHOLD
```
huylexuan44: Backend limit is bytes (proto-server MAX_BUF_SIZE 4096), so the UTF-8 send gate is right and the UTF-16 counter/paste gates are what stay silent: 2804 Vietnamese chars convert on Enter with no countdown. Pre-existing, fine as a follow-up: counter should use content_payload_utf8_len.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -1050,48 +1103,46 @@ impl MentionInput {
...
-                self.input.update(cx, |input, cx| {
-                    input.set_mention_spans(Vec::new(), cx);
-                    input.set_value("", window, cx);
-                });
-                self.convert_text_to_file(combined, window, cx);
-                cx.notify();
-                return;
-            }
+        }
+        let pasted = trim_paste_to_utf16_cap(&current, &text, PASTE_SAFETY_CAP_UTF16);
```
huylexuan44, line 1118: Trim drops the tail silently. Toast when the pasted chunk was cut at the cap.

crates/mezon-ui/src/chat/mention_input/mod.rs
```diff
@@ -520,6 +522,46 @@ fn content_payload_utf8_len(text: &str) -> usize {
...
+        end = i + ch.len_utf8();
+    }
+    &pasted[..end]
+}
+
+fn next_converted_text_filename() -> String {
+    static COUNTER: AtomicU64 = AtomicU64::new(0);
+    let pid = std::process::id();
+    let n = COUNTER.fetch_add(1, Ordering::Relaxed);
+    format!("converted-text-{pid}-{n}.txt")
```
huylexuan44: This name is what recipients see in chat (was {millis}.txt). Consider message-{millis}-{n}.txt, keeping the counter for uniqueness.

huylexuan44:
Round 4 on a4cbd255:
- Test CI is red because of this PR: the new i18n key exists only in en/vi (develop 54c75772 Test is green). Format/Lint/Deny/Windows PDF are red on develop already (mezon-store/src/lib.rs:251 fmt); clippy on this PR is clean.
- Merged into develop locally: mezon-ui mention_input 37/37 pass, mezon-i18n fails only on the missing key.
- Fixes verified: paste-path toast + temp cleanup, count check before write, add_pending -> bool, pid+counter filename, anonymous gate, 100k paste cap.
- Still open: sync fs::write on the Enter path (see inline: background write + re-emit Submit keeps the one-key flow). UTF-16/UTF-8 mismatch is pre-existing and can be a follow-up. Mentions lost in .txt and the whitespace-only block: resolved.

huylexuan44:
Pushed de7572cc on top of a4cbd255 with the round-4 items applied:
- i18n key added to the 14 missing locales (mezon-i18n green again)
- Enter on an oversized draft now writes the .txt in background_spawn and re-emits Submit, so it is still one keypress; second Enter mid-write is a no-op, channel switch mid-write discards the file
- counter / paste gate / send gate all measure the {"t":text} payload in UTF-8 bytes; json_string_utf16_len removed
- toast when a paste is cut at the 100k cap; converted files are named message-{millis}-{n}.txt
Verified locally (merged into develop): clippy + fmt clean, mezon-i18n 11/11, mention_input 38/38; live: 4000 ASCII and 2900 Vietnamese chars each send one .txt on a single Enter with the countdown visible, double Enter sends one message, anonymous mode toasts and keeps the draft, empty-composer paste stages a .txt, draft paste appends, 100k paste truncates with a toast.

PR #488, opened 2026-09-11
fix: remove mention here and fix sort mention timestamp in inbox ment…

duocphungvanNCC:
> `Description:`

**env development:** window
- remove  mention @here in inbox mention
- fix sort mention timestamp.
image

### DIFF
--- a/crates/mezon-client/src/inbox.rs
+++ b/crates/mezon-client/src/inbox.rs
@@ -814,6 +814,15 @@ impl InboxNotification {
             .filter(|ts| *ts > 0)
             .unwrap_or(self.create_time_seconds)
     }
+
+    pub fn contains_here_mention(&self) -> bool {
+        self.message.as_ref().is_some_and(|message| {
+            message
+                .mention_spans
+                .iter()
+                .any(|span| crate::transport::is_here_user_id(&span.user_id))
+        })
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -1089,6 +1098,34 @@ mod tests {
             inbox_category_from_notification(&n),
             Some(InboxCategory::Mentions)
         );
+    }
+
+    #[test]
+    fn inbox_notification_detects_here_mention_span() {
+        let mut message = InboxMessagePreview::empty_content("@here".into());
+        message.mention_spans.push(InboxMentionSpan {
+            start: 0,
+            end: 5,
+            user_id: crate::transport::MENTION_HERE_USER_ID.into(),
+            role_id: String::new(),
+            is_role: false,
+        });
+        let notification = InboxNotification {
+            id: "1".into(),
+            category: InboxCategory::Mentions,
+            subject: String::new(),
+            sender_id: String::new(),
+            clan_id: "1".into(),
+            channel_id: "1".into(),
+            topic_id: None,
+            channel_type: 1,
+            avatar_url: String::new(),
+            create_time_seconds: 1,
+            code: INBOX_USER_MENTIONED_CODE,
+            message: Some(message),
+        };
+
+        assert!(notification.contains_here_mention());
     }
 
     #[test]
--- a/crates/mezon-store/src/badge.rs
+++ b/crates/mezon-store/src/badge.rs
@@ -115,6 +115,19 @@ fn is_clan_message_seen(cx: &App, m: &ChannelMessage, from_me: bool) -> bool {
     active == badge_id || active == parent_id
 }
 
+fn mention_check_policy(
+    is_new_message: bool,
+    seen: bool,
+    already_seen: bool,
+    removed_by_other: bool,
+) -> (bool, bool) {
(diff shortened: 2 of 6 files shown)


crates/mezon-client/src/inbox.rs
```diff
@@ -814,6 +814,15 @@ impl InboxNotification {
             .filter(|ts| *ts > 0)
             .unwrap_or(self.create_time_seconds)
     }
+
+    pub fn contains_here_mention(&self) -> bool {
```
huylexuan44, line 818: Also drops direct mentions/replies that merely include @here (e.g. "@here @alice"); realtime shows them, refetch removes them. Make this the inverse of is_inbox_mention_or_reply: no span/ref targets me or my roles.

crates/mezon-store/src/inbox.rs
```diff
@@ -433,17 +453,48 @@ impl InboxStore {
...
+            .filtered_here_badges
+            .entry(clan_id.to_string())
+            .or_default()
+            .insert(message_id.to_string());
+        if inserted {
+            cx.notify();
+        }
+    }
+
+    pub fn has_visible_inbox_badge(&self, clan_id: &str, total_badges: u32) -> bool {
```
huylexuan44: filtered_here_badges only shrinks on the MarkAsRead push; LastSeenUpdated, deleted @here, and muted clans decrement/skip badge_count but not this set, so total_badges > filtered goes false and a real mention badge stays hidden.

crates/mezon-store/src/inbox.rs
```diff
@@ -117,6 +120,11 @@ impl InboxStore {
             dispatch.on(RealtimeKind::Notifications, &entity, |this, event, cx| {
                 this.handle_event(event, cx);
             });
+            dispatch.on(RealtimeKind::MarkAsRead, &entity, |this, event, cx| {
+                if let RealtimeEvent::MarkAsRead(event) = event {
+                    this.clear_filtered_here_badges(&event.clan_id.to_string(), cx);
```
huylexuan44: Channel/category-scoped MarkAsRead wipes the whole clan set: @here in #a and #b, mark #a read -> badge 1, filtered 0 -> dot shows for an @here-only badge. Scope the clear to event.channel_id/category_id.

crates/mezon-store/src/inbox.rs
```diff
@@ -433,17 +453,48 @@ impl InboxStore {
...
         items.sort_by(|a, b| {
-            match (
-                is_pending_inbox_notification_id(&a.id),
-                is_pending_inbox_notification_id(&b.id),
-            ) {
-                (true, false) => std::cmp::Ordering::Less,
-                (false, true) => std::cmp::Ordering::Greater,
-                _ => b.create_time_seconds.cmp(&a.create_time_seconds),
-            }
+            b.message_timestamp()
```
huylexuan44, line 558: Sorting by message_timestamp breaks page_cursor: server pages by id < cursor but the last sorted item is no longer the min id (your own test: cursor "10" while "20" is loaded), so load-more re-fetches duplicates and stalls. Cursor should be min(id) over non-pending items.

crates/mezon-store/src/inbox.rs
```diff
@@ -461,7 +512,11 @@ impl InboxStore {
                 bucket.loading = false;
                 match result {
                     Ok(mut items) => {
-                        bucket.has_more = items.len() >= INBOX_PAGE_LIMIT as usize;
+                        let page_has_more = items.len() >= INBOX_PAGE_LIMIT as usize;
+                        if category == InboxCategory::Mentions {
+                            items.retain(|item| !item.contains_here_mention());
```
huylexuan44: If page 1 is all @here, items is empty -> page_cursor None -> fetch_more bails despite has_more; older real mentions become unreachable. Compute the cursor from the unfiltered page or keep fetching until a non-empty page.

huylexuan44: Fetched @here rows are dropped without feeding filtered_here_badges, but badge_count is seeded from the server, so after restart the inbox dot shows for an empty Mentions list.

crates/mezon-store/src/badge.rs
```diff
@@ -115,6 +115,19 @@ fn is_clan_message_seen(cx: &App, m: &ChannelMessage, from_me: bool) -> bool {
...
 
+fn mention_check_policy(
+    is_new_message: bool,
+    seen: bool,
+    already_seen: bool,
+    removed_by_other: bool,
+) -> (bool, bool) {
+    let needs_badge_mention_check = (is_new_message && !seen && !already_seen) || removed_by_other;
+
+    let needs_inbox_mention_check = is_new_message;
```
huylexuan44, line 126: is_inbox_mention now runs a full content JSON parse for every clan message incl. own and currently-viewed ones, on top of is_mention. Gate on !from_me and share one parse between the two predicates.

PR #495, opened 2026-09-14
fix: set permisstion manage channel for interactive button in voice

duocphungvanNCC:
> `Description:`

**env development:** window
- check permisstion manage channel for interactive button
- add param synce board of blackboard interactive app.


crates/mezon-store/src/voice.rs
```diff
@@ -1041,6 +1045,17 @@ impl VoiceStore {
...
+    fn open_interactive_app_with_params(
+        &mut self,
+        app: VoiceInteractiveApp,
+        sender_id: i64,
+        clan_id: i64,
+        params: Option<String>,
+        cx: &mut Context<Self>,
     ) {
         let mut hasher = DefaultHasher::new();
         (sender_id, clan_id).hash(&mut hasher);
```
huylexuan44, line 1055: Dedupe key ignores params: the reaction-token path (open_interactive_app at ~1160, sent at ~3151) can open Blackboard first without userId, then this event with userId is dropped for 10s. Carry userId in app_reaction_token or stop opening apps from the reaction path.

crates/mezon-store/src/voice.rs
```diff
@@ -1085,14 +1100,22 @@ impl VoiceStore {
...
                     clan_id: &clan_id_string,
                     clan_name: clan_name.as_deref(),
                 },
             );
+            if let Some(params) = params.filter(|params| !params.is_empty())
+                && let Ok(mut parsed) = url::Url::parse(&url)
+            {
+                parsed
+                    .query_pairs_mut()
+                    .extend_pairs(url::form_urlencoded::parse(params.as_bytes()));
```
huylexuan44: event.params is peer-controlled and merged with no key whitelist; a peer sending data=<their web_app_data> yields a duplicate data pair and the app may auth as them. Copy only the expected key (userId) instead of extend_pairs over everything.

PR #513, opened 2026-09-16
Fix/confirmation open interactive app in voice

duocphungvanNCC:
> `Description:`

**env development:** window
- fix: remove permisstion check open interactive app
- fix not auto open app when receive event, need join app then open.
- check open 1 app only for each interactive app. 

image

- also: raised-hands overlay sizes to its content (min 100px / max 360px) instead of a fixed 320px.


crates/mezon-store/src/voice.rs
```diff
@@ -1120,7 +1159,32 @@ impl VoiceStore {
...
+                tracing::warn!("managed voice app launch skipped: no opener is registered");
+                let _ = this.update(cx, |store, cx| {
+                    store.interactive_launches.remove(&launch_key);
+                    store.opened_interactive_apps.remove(&app);
+                    cx.notify();
+                });
+                return;
+            };
+            let window_key = format!("voice-{}", app.app_id());
+            let result = cx
```
huylexuan44: Regression on macOS: Chrome keeps running after the app window closes, so status() never returns and opened_interactive_apps holds the app forever (Launch/Join blocked, temp profile never removed). Verified live with these exact switches.

crates/mezon-native/src/browser.rs
```diff
@@ -108,6 +108,38 @@ pub fn open_url_app_window(url: &str) -> anyhow::Result<()> {
...
+            } else {
+                '_'
+            }
+        })
+        .collect();
+    let profile = std::env::temp_dir()
+        .join("mezon-managed-apps")
+        .join(format!("{}-{safe_key}", std::process::id()));
+    std::fs::create_dir_all(&profile)?;
+    let result = match default_chromium_browser() {
```
huylexuan44: Regression: Safari/Firefox default and the MAS build used to fall back to a browser tab; now Launch silently does nothing. Fall back to open_url_app_window or surface the error.

crates/mezon-store/src/voice.rs
```diff
@@ -1120,7 +1159,32 @@ impl VoiceStore {
...
+                .background_executor()
+                .spawn(async move { opener(&url, &window_key) })
```
huylexuan44: Perf: this blocks a background executor thread for the whole window lifetime; the Linux pool is fixed at available_parallelism threads. Wait on a std::thread (or spawn + try_wait) instead.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -3768,10 +3769,12 @@ impl InteractiveAppPopoverPanel {
         let focus_handle = cx.focus_handle();
         cx.on_blur(&focus_handle, window, |_, _, cx| cx.emit(DismissEvent))
             .detach();
+        let voice_subscription = cx.subscribe(&voice, |_, _, _, cx| cx.notify());
```
huylexuan44: cx.subscribe only fires on emit(VoiceStoreEvent); these state changes use cx.notify(), so use cx.observe or the open popover never refreshes.

crates/mezon-store/src/voice.rs
```diff
@@ -3913,6 +4012,7 @@ impl VoiceStore {
         self.device_submenu = None;
         self.participant_menu = None;
         self.interactive_launches.clear();
+        self.active_interactive_apps.clear();
```
huylexuan44, line 4039: opened_interactive_apps survives reset(): in the next voice channel the button stays green and handle_interactive_app_signal drops that app signals, so Join never appears.

crates/mezon-native/src/browser.rs
```diff
@@ -108,6 +108,38 @@ pub fn open_url_app_window(url: &str) -> anyhow::Result<()> {
...
+        Some(browser) => launch_managed_app_window(&browser, url, &profile),
+        None => Err(anyhow::anyhow!(
+            "a Chromium-based default browser is required"
+        )),
+    };
+    if let Err(error) = std::fs::remove_dir_all(&profile) {
```
huylexuan44: On Windows Chrome still holds the profile briefly after exit, so this remove_dir_all fails and the pid-keyed dir accumulates in temp; use a stable per-app dir or sweep stale dirs at startup.

crates/mezon-native/src/browser.rs
```diff
@@ -108,6 +108,38 @@ pub fn open_url_app_window(url: &str) -> anyhow::Result<()> {
     }
 }
 
+/// Opens a Chromium app in an isolated process and blocks until that app window closes.
```
huylexuan44: The doc on open_url_app_window (lines 82-87) still says a dedicated profile was rejected; update it now that this path relies on one.

crates/mezon-ui/src/chat/voice.rs
```diff
@@ -1307,7 +1307,8 @@ fn raised_hands_overlay(
             .absolute()
             .top(px(68.))
             .right(px(8.))
-            .w(px(320.))
+            .min_w(px(100.))
```
huylexuan44, line 1310: Unrelated change (raised hands overlay width) — split it out or note it in the PR description.

crates/mezon-store/src/voice.rs
huylexuan44: Fixed in dcb8d312: on macOS the launcher now watches the process's app windows (CGWindowList, layer 0) and sends SIGTERM once the window is gone or never appears within 30s, so the wait returns and the app is released.

crates/mezon-native/src/browser.rs
huylexuan44: Fixed in dcb8d312: no Chromium default / sandbox / spawn failure now fall back to open_url and return immediately, so Launch works everywhere as before.

crates/mezon-store/src/voice.rs
huylexuan44: Fixed in dcb8d312: the window wait runs on a dedicated std::thread and the task awaits a oneshot, no executor thread is held.

crates/mezon-ui/src/chat/voice.rs
huylexuan44: Fixed in dcb8d312: switched to cx.observe.

crates/mezon-store/src/voice.rs
huylexuan44: Fixed in dcb8d312: reset() clears opened_interactive_apps; each launch carries a window token so an earlier window's exit cannot release a newer one.

crates/mezon-native/src/browser.rs
huylexuan44: Fixed in dcb8d312: profile dir is now stable per app key (no pid), so a failed cleanup is reused on the next launch instead of accumulating.

crates/mezon-native/src/browser.rs
huylexuan44: Fixed in dcb8d312: doc updated to point at the managed variant.

crates/mezon-ui/src/chat/voice.rs
huylexuan44: Noted in the PR description.

PR #515, opened 2026-09-16
fix: fix event date selection

duocphungvanNCC:
> `Description:`
The separate End Date picker could be confused with the end of a recurring series when combined with repeat frequency options.
**env development:** window
- remove end date feild
- Use a single date for both the event start and end timestamps.
- Rename the start date label to date 
image


crates/mezon-ui/src/chat/create_event_modal.rs
```diff
@@ -568,9 +548,6 @@ impl CreateEventModal {
             .timestamp_opt(event.end_time_seconds as i64, 0)
             .single()
         {
-            modal.end_date.update(cx, |picker, cx| {
-                picker.set_selected_silent(Some(end.date_naive()), cx)
-            });
             modal.end_time.update(cx, |select, cx| {
```
huylexuan44, line 654: Regression: only the end time-of-day is restored, and timestamps() re-anchors end on start_date, so saving any edit (even title-only) of a multi-day event created on web silently rewrites end_time to the start date; a cross-midnight event cannot be saved at all (end < start).

crates/mezon-ui/src/chat/create_event_modal.rs
```diff
@@ -691,7 +667,7 @@ impl CreateEventModal {
         };
         Some((
             local_timestamp(start_date, start_seconds)?,
-            local_timestamp(end_date, end_seconds)?,
+            local_timestamp(start_date, end_seconds)?,
```
huylexuan44: Regression: end is anchored on start_date, so cross-midnight events can no longer be created (web still allows them, backend only requires end >= start); a 23:45 start has no valid end slot. Keep the end date picker (the clamp above already fixes the default-end rollover) or link the product decision.

crates/mezon-ui/src/chat/create_event_modal.rs
```diff
@@ -86,6 +86,10 @@ fn event_repeat_labels(locale: &str, date: chrono::NaiveDate) -> Vec<SharedStrin
     labels.into_iter().map(Into::into).collect()
 }
 
+fn event_default_end_minutes(start_minutes: u32) -> u32 {
```
huylexuan44: Hardcoded 23*60+45 duplicates the 96-slot grid — derive the cap from time_values.last().

assets/i18n/en.json
```diff
@@ -2702,7 +2702,7 @@
   "eventCreator.fields.eventName.placeholder": "What's your event?",
   "eventCreator.fields.eventName.title": "Event Topic",
   "eventCreator.fields.startDate.errorMessage": "Please choose a date on or after",
-  "eventCreator.fields.startDate.title": "Start Date",
+  "eventCreator.fields.startDate.title": "Date",
```
huylexuan44: Do not repurpose startDate.title (web keeps it as Start Date) — add a new key such as eventCreator.fields.date.title.

huylexuan44: eventCreator.fields.endDate.title (line 2679) is now unreferenced in all 16 locales — remove it or restore the end date field.

crates/mezon-ui/src/chat/create_event_modal.rs
duocphungvanNCC: Removing the end-date picker is intentional to avoid confusion between an occurrence’s end date and the repeat frequency. The change has been apply into web in dev enviroment

PR #517, opened 2026-09-16
Fix ctrl K search layout issues

giachuvan:


mezon-protocol
```diff
@@ -1 +1 @@
-Subproject commit a4710196187cc0936fb393475efbfc32a223cc2a
+Subproject commit 0c8f186c965f59d665754c2b9014d18ac9831617
```
huylexuan44: Unrelated submodule bump (a4710196 -> 0c8f186c, v2.0.63 -> v2.0.67) in a layout PR - drop it here or split into its own reviewed protocol bump.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -693,10 +693,19 @@ pub fn render_palette_row(
             .flex()
```
huylexuan44: Both match arms (Channel and Direct|Member) are now identical - collapse row_content into one expression, item.kind selects nothing here.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -780,13 +789,12 @@ fn render_palette_label(
         .items_center()
         .gap_1()
         .min_w_0()
-        .overflow_hidden()
-        .child(label)
+        .child(div().min_w_0().overflow_hidden().child(label))
```
huylexuan44: Wrapper is redundant: label already has .truncate() and an overflow!=visible flex item gets min-size 0 in taffy - add label directly and keep overflow_hidden on the row.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -780,13 +789,12 @@ fn render_palette_label(
...
         .child(
             div()
                 .flex_shrink_0()
                 .text_size(px(15.))
-                .italic()
+                .font_weight(FontWeight::NORMAL)
```
huylexuan44: font_weight(NORMAL) is a no-op (default weight, nothing to override) - if the intent is just not italic, deleting .italic() is the whole change.

crates/mezon-ui/src/command_palette/items.rs
```diff
@@ -780,13 +789,12 @@ fn render_palette_label(
...
                 .text_color(theme.danger_text)
                 .child("(busy)"),
```
huylexuan44: The same (busy) tag in mention_input/mod.rs:2537-2545 still has .italic() - share one busy-tag helper so both lists render alike; also a candidate for mezon_i18n (clan.busy exists).

PR #520, opened 2026-09-16
fix: wrong file attachment in inbox message after add to inbox and ki…

duocphungvanNCC:
> `Description:`

**env development:** window
- bug desscription: add to inbox a file then the file show in inbox message but after kill app and turn back the file turn into a log file. 
- fixed show wrong log file attach in inbox message after add to inbox file type
image
image


crates/mezon-client/src/inbox.rs
```diff
@@ -735,11 +735,11 @@ fn parse_message_preview_json(bytes: &[u8]) -> Option<InboxMessagePreview> {
     if preview.content.is_empty() && !preview.raw_content.is_empty() {
         preview.content = display_text_from_message_content(&preview.raw_content);
     }
-    apply_preview_attachments(&mut preview);
     apply_first_attachment(
         &mut preview,
         &serde_json::json!({ "attachments": raw.attachments }),
     );
+    apply_preview_attachments(&mut preview);
```
huylexuan44, line 742: Regression: sibling attachments now win over content JSON, and first_attachment_from_json synthesizes filename_from_url when filename is missing, so a sibling entry without filename locks in the URL basename and the real filename in content is ignored. Revert this swap or only run it when the sibling entry has a filename.

crates/mezon-client/src/inbox.rs
```diff
@@ -1177,6 +1177,29 @@ mod tests {
         assert_eq!(preview.attachment_type, "application/pdf");
     }
 
+    #[test]
+    fn saved_inbox_fcm_restores_original_attachment_metadata() {
```
huylexuan44, line 1181: This test never hits changed code: the bytes are DirectFcmProto (tag 0x1A) so it takes the untouched preview_from_fcm path and hand-builds the JSON this PR should produce; it passes on develop. Cover the add_to_inbox embedding and the reordered precedence instead.

crates/mezon-store/src/messages.rs
```diff
@@ -3363,6 +3363,33 @@ impl MessagesStore {
                 size: i32::try_from(att.size).unwrap_or(0),
             })
             .collect();
+
+        let content_json = if attachments.is_empty() {
```
huylexuan44, line 3367: This roughly doubles the per-attachment bytes of a request sent as one socket frame; proto-server drops the connection above MAX_BUF_SIZE 4096 and there is no outbound size gate. Albums of ~12+ files will fail. Preview only renders attachments[0], so embed only the first attachment metadata.

crates/mezon-store/src/messages.rs
```diff
@@ -3363,6 +3363,33 @@ impl MessagesStore {
...
+            content_json
+        } else {
+            let mut content = serde_json::from_str::<serde_json::Value>(&content_json)
+                .ok()
+                .filter(serde_json::Value::is_object)
+                .unwrap_or_else(|| serde_json::json!({ "t": msg.content }));
```
huylexuan44, line 3373: Double-encoded content (a JSON string wrapping the object, accepted by parse_message_content_tokens) is replaced by {"t": msg.content} here, dropping mk/lk/ej tokens from the inbox row. Unwrap a Value::String that parses to an object before falling back.

crates/mezon-store/src/messages.rs
```diff
@@ -3363,6 +3363,33 @@ impl MessagesStore {
...
+            content["attachments"] = serde_json::Value::Array(
```
huylexuan44, line 3374: Third copy of this attachment shape: pinned.rs:420-430 has the identical block and ApiAttachment already derives Serialize with these eight fields. Reuse it, and extract this whole block into a pure fn so it can be unit-tested.

crates/mezon-store/src/messages.rs
huylexuan44, line 3367: Root cause is server-side: CreateMessage2Inbox (mezon-api api_pin_message.go:63-77) and the mention path in mezon-fcm drop filename/size/thumbnail. This only fixes desktop-saved rows; mention/reply rows and rows saved from web/mobile still show basename + 0 B. Persisting in.Attachments in the handler fixes every client and avoids the frame-size issue.

PR #529, opened 2026-09-17
fix: pin image get attachment from message first then image attachment of pin

duocphungvanNCC:
- bug description: upload and then pin an image sometime not show image and wrong image in pinned message 

image


crates/mezon-ui/src/chat/pinned_popover.rs
```diff
@@ -217,6 +227,10 @@ impl PinnedPopoverPanel {
                 this.pin_cards = this.compute_pin_cards(cx);
                 cx.notify();
             }),
+            cx.observe(&MessagesStore::global(cx), |this, _, cx| {
```
huylexuan44: Broad observe on MessagesStore rebuilds every pin card and re-renders the popover on each store notify (every message/reaction/scroll) — subscribe to MessagesEvent::Updated { message_id } and refresh only the matching card, like refresh_name_rows.

crates/mezon-ui/src/chat/pinned_popover.rs
```diff
@@ -91,6 +91,16 @@ impl PinCardVm {
         let (poll, poll_my_vote) = resolve_pin_poll(msg, channel_id, cx);
         let text_spans = prepare_pin_text_spans(msg);
         let selectable_text = pin_canonical_text(msg, &text_spans);
+        let mut pin = msg.clone();
```
huylexuan44, line 110: Only masks the bug while the message is in the local cache; the pin record written by PinnedMessagesStore::pin (pinned.rs write_last_pin_message) still carries the wrong attachment, so other users and post-restart views still show it — fix the write path.

PR #533, opened 2026-09-18
Add self DM

giachuvan:


crates/mezon-store/src/direct.rs
```diff
@@ -666,12 +667,8 @@ impl DirectMessageStore {
...
             })?;
 
-            if let Err(e) = api
-                .join_chat(0, channel_id.get(), channel_type, false)
-                .await
-            {
-                tracing::warn!("join_chat after create DM failed: {e}");
-            }
+            api.join_chat(0, channel_id.get(), channel_type, false)
+                .await?;
```
huylexuan44, line 671: Regression: join_chat has no REST fallback, so ? now aborts before the send that does fall back to HTTP; token card is silently lost while the socket reconnects (callers detach). Keep warn-and-continue, or only propagate server rejections.

crates/mezon-ui/src/components/compositions/footer_profile_popup.rs
```diff
@@ -204,6 +227,63 @@ impl FooterProfilePopup {
...
+                        cx,
+                        Route::DirectMessage {
+                            direct_id: channel_id,
+                            message_type: channel_type.to_string(),
+                        },
+                    );
+                });
+            }
+            Err(err) => {
+                tracing::warn!("footer profile message send failed: {err}");
```
huylexuan44, line 277: Send failure is only logged; the input just reappears with the text. Report via shell.error like friends_page open_dm_with_user does.

crates/mezon-ui/src/components/compositions/footer_profile_popup.rs
```diff
@@ -734,7 +819,37 @@ impl Render for FooterProfilePopup {
...
+                                    div()
+                                        .mt_2()
+                                        .p_2()
+                                        .rounded(px(5.))
+                                        .text_center()
+                                        .text_sm()
+                                        .italic()
+                                        .bg(bg_box)
+                                        .text_color(text_primary)
+                                        .child(tk("userProfile.labels.userNotFound")),
```
huylexuan44, line 850: userNotFound renders "User is not in a Clan" under your own profile when the account has not loaded; username/display_name are set once in new() and never refreshed by _account_sub.

PR #535, opened 2026-09-18
fix: fix bot qr message not show qr image

duocphungvanNCC:
> `Description:`
send *qr bot response without qr image.
image

**env development:** window
-  decode inline image data uri before creating HTTP requests allowing bot QR images to render


crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -22,6 +23,38 @@ const STATS_LOG_INTERVAL: u64 = 256;
...
+    let mime = parameters.next().unwrap_or_default();
+    anyhow::ensure!(mime.starts_with("image/"), "data URI must contain an image");
+    anyhow::ensure!(
+        payload.len() <= MAX_ENTRY_BYTES * 3,
+        "inline image exceeds size limit"
+    );
+    let encoded = percent_encoding::percent_decode_str(payload).collect::<Vec<u8>>();
+    let bytes = if parameters.any(|parameter| parameter.eq_ignore_ascii_case("base64")) {
+        base64::engine::general_purpose::STANDARD
+            .decode(encoded)
```
huylexuan44: STANDARD (RequireCanonical) rejects unpadded (Go RawStdEncoding) and line-wrapped base64 that browsers accept, so those bot images still fail. Strip ASCII whitespace and use DecodePaddingMode::Indifferent.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -212,6 +245,31 @@ async fn buffered_response(
 }
 
 impl HttpClient for DiskImageCacheClient {
+    fn get(
+        &self,
+        uri: &str,
+        body: AsyncBody,
+        follow_redirects: bool,
+    ) -> BoxFuture<'static, anyhow::Result<Response<AsyncBody>>> {
+        if uri.starts_with("data:") {
```
huylexuan44: A malformed data URI errors as ImageCacheError::Other, which retryable_failure() treats as transient: it is re-decoded every 15s while on screen and the warn is suppressed. Return a non-retryable class for data: failures.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -22,6 +23,38 @@ const STATS_LOG_INTERVAL: u64 = 256;
...
+fn inline_image_response(uri: &str) -> anyhow::Result<Response<AsyncBody>> {
+    let data = uri
+        .strip_prefix("data:")
+        .ok_or_else(|| anyhow::anyhow!("invalid image data URI"))?;
+    let (metadata, payload) = data
+        .split_once(',')
+        .ok_or_else(|| anyhow::anyhow!("missing image data payload"))?;
+    let mut parameters = metadata.split(';');
+    let mime = parameters.next().unwrap_or_default();
+    anyhow::ensure!(mime.starts_with("image/"), "data URI must contain an image");
```
huylexuan44: starts_with("image/") admits image/svg+xml; guess_format fails and the bytes fall through to usvg, which every other message-image path refuses (is_image_type, canvas decode_data_image). Restrict to raster mimes.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -22,6 +23,38 @@ const STATS_LOG_INTERVAL: u64 = 256;
...
+    anyhow::ensure!(
+        payload.len() <= MAX_ENTRY_BYTES * 3,
```
huylexuan44: Bound is 48 MB for both encodings and percent_decode always copies: a 40 MB base64 payload allocates 40+30 MB before the 16 MB check. Cap base64 at MAX_ENTRY_BYTES*4/3+4 and use Cow<[u8]> to skip the copy.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -22,6 +23,38 @@ const STATS_LOG_INTERVAL: u64 = 256;
...
+        "inline image exceeds size limit"
+    );
+    let encoded = percent_encoding::percent_decode_str(payload).collect::<Vec<u8>>();
+    let bytes = if parameters.any(|parameter| parameter.eq_ignore_ascii_case("base64")) {
```
huylexuan44: Exact match misses 'image/png; base64' (space after ';', accepted by Chrome/Firefox) and a trailing #fragment; payload then goes down the raw path and fails. Trim each parameter and strip the fragment.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -22,6 +23,38 @@ const STATS_LOG_INTERVAL: u64 = 256;
 
 type ImageHybridCache = HybridCache<u64, Vec<u8>>;
 
+fn inline_image_response(uri: &str) -> anyhow::Result<Response<AsyncBody>> {
+    let data = uri
+        .strip_prefix("data:")
```
huylexuan44: Scheme and mime checks are byte-case-sensitive; DATA:image/... or data:IMAGE/PNG parses as a URL so gpui routes it here, then Request::builder fails with InvalidUri and it retries forever. Use eq_ignore_ascii_case.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -237,6 +295,9 @@ impl HttpClient for DiskImageCacheClient {
             }
         }
         let uri = req.uri().to_string();
+        if *req.method() == Method::GET && uri.starts_with("data:") {
```
huylexuan44: Dead code: http::Uri cannot represent data:image/... (parses as authority, InvalidFormat), so no Request<AsyncBody> ever reaches this branch. get() is the only entry point; delete this and its to_string scan.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -22,6 +23,38 @@ const STATS_LOG_INTERVAL: u64 = 256;
...
+    } else {
+        encoded
+    };
+    anyhow::ensure!(
+        !bytes.is_empty() && bytes.len() <= MAX_ENTRY_BYTES,
+        "invalid inline image size"
+    );
+    Ok(Response::builder()
+        .status(200)
+        .header(http::header::CONTENT_TYPE, mime)
```
huylexuan44: Nothing reads Content-Type (loaders sniff with guess_format), but a mime with a non-visible-ASCII byte makes builder() fail and rejects a valid PNG. Drop the header or set it only if HeaderValue::from_str(mime) succeeds.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -22,6 +23,38 @@ const STATS_LOG_INTERVAL: u64 = 256;
 
 type ImageHybridCache = HybridCache<u64, Vec<u8>>;
 
+fn inline_image_response(uri: &str) -> anyhow::Result<Response<AsyncBody>> {
```
huylexuan44, line 25: Second data-URI decoder diverging from mezon_canvas::image::decode_data_image (image.rs:387) on percent-decoding, raw payloads, SVG and size caps. Extract one shared helper and call it from both.

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -529,4 +590,54 @@ mod tests {
...
+        assert_eq!(response.status(), 200);
+        assert_eq!(response.headers()[http::header::CONTENT_TYPE], "image/png");
+        let mut body = response.into_body();
+        let mut bytes = Vec::new();
+        futures::executor::block_on(body.read_to_end(&mut bytes)).unwrap();
+        assert_eq!(bytes, b"\x89PNG\r\n\x1a\n");
+    }
+
+    #[test]
+    fn image_client_get_resolves_inline_png() {
```
huylexuan44, line 588: Builds a real ReqwestClient (spins up the transport runtime) for a branch that never touches inner, and only asserts 200. Use a fake client and assert a cacheable URL passed to get() still reaches send().

crates/mezon-client/src/image_disk_cache.rs
```diff
@@ -212,6 +245,31 @@ async fn buffered_response(
...
 
 impl HttpClient for DiskImageCacheClient {
+    fn get(
+        &self,
+        uri: &str,
+        body: AsyncBody,
+        follow_redirects: bool,
+    ) -> BoxFuture<'static, anyhow::Result<Response<AsyncBody>>> {
+        if uri.starts_with("data:") {
+            let result = inline_image_response(uri);
```
huylexuan44: Once data URIs reach the decoders, a corrupt-but-well-formed payload makes mezon-ui/src/image_cache.rs:1354 warn with resource_source(), i.e. the entire base64 blob, on every reload. Truncate data: sources there to scheme+mime+length.

crates/mezon-client/src/data_image.rs
```diff
@@ -0,0 +1,88 @@
+use std::borrow::Cow;
+
+use base64::Engine as _;
+
+const MAX_IMAGE_BYTES: usize = 16 * 1024 * 1024;
+const MAX_BASE64_ENCODED_BYTES: usize = MAX_IMAGE_BYTES * 4 / 3 + 4;
+const MAX_BASE64_WHITESPACE_BYTES: usize = MAX_BASE64_ENCODED_BYTES / 32;
+const RASTER_MIMES: &[&str] = &[
```
huylexuan44, line 8: Regression: image/jpg, image/x-icon, image/apng, image/tiff are rejected with a permanent 400 although every loader sniffs bytes with guess_format, and canvas used to accept any data:image/*. Reject only image/svg+xml and let image/* through.

crates/mezon-ui/src/image_cache.rs
```diff
@@ -618,7 +618,21 @@ fn retryable_failure(error: &ImageCacheError) -> bool {
...
+            let uri = uri.as_ref();
+            if uri
+                .get(..5)
+                .is_some_and(|prefix| prefix.eq_ignore_ascii_case("data:"))
+            {
+                let metadata = uri[5..]
+                    .split_once(',')
+                    .map_or("unknown", |(metadata, _)| metadata);
+                let metadata = metadata.get(..metadata.len().min(80)).unwrap_or("unknown");
+                format!("data:{metadata},<redacted> (length={})", uri.len())
```
huylexuan44, line 631: This redaction is bypassed on the failure path: a bad data URI yields BadStatus { uri: full data URI } whose Display is 'unexpected http status for {uri}' and is printed right after {source} in the warn at 1369. Redact uri when building BadStatus for data: sources too.

huylexuan44:
Round 2: 10 of the 11 previous comments are addressed, thanks. Two items to fix before merge (inline): the raster mime whitelist regresses image/jpg and canvas, and the log redaction is bypassed via BadStatus Display.
Follow-ups, fine after merge: embed click-through on data URIs (viewer filename, Save, Copy image all assume http), is_data_image_uri now matches any data: scheme, orphan base64 dep in mezon-canvas/Cargo.toml, oversized test asserts against MAX_ENTRY_BYTES instead of MAX_IMAGE_BYTES, the two get() tests still spin a real ReqwestClient.
Perf checked: decode runs in the background loader, nothing added to render; CI Format/Lint failures are the pre-existing develop fmt drift.

